PR #61, opened 2022-04-01
fix: Stop blocking the async runtime when connecting to quilc or a QPU.

### LINKED ISSUE #31: Stop blocking the Tokio runtime
If connecting to `quilc` or the eventual QPU over RPCQ fails, the SDK currently just spins indefinitely. A timeout should be set (and preferably configurable) so that the SDK gives up on executing and specifies where it failed (was it quilc or QPU connection).

dbanty:
Closes #31

This basically introduces threads to run all the RPCQ connections inside of. As a result, there's some `Arc`ing and `Box`ing. This PR will result in a bit slower performance overall but I think it's well worth it to not block the runtime. Importantly, this also enables callers to use timeouts if they want to give up eventually rather than waiting forever.

### DIFF
--- a/qcs/src/executable.rs
+++ b/qcs/src/executable.rs
@@ -2,6 +2,7 @@
 //! users will interact with QCS, quilc, and QVM.
 
 use std::collections::HashMap;
+use std::sync::Arc;
 use std::time::Duration;
 
 use eyre::{eyre, Report, Result, WrapErr};
@@ -53,17 +54,17 @@ use crate::{qpu, qvm, ExecutionResult};
 /// the methods most likely needs to live as long as this struct. Check individual methods for
 /// specifics. If only using `'static` strings then everything should just work.
 pub struct Executable<'executable, 'execution> {
-    quil: &'executable str,
+    quil: Arc<str>,
     shots: u16,
     readout_memory_region_names: Option<Vec<&'executable str>>,
-    params: Parameters<'executable>,
+    params: Parameters,
     compile_with_quilc: bool,
-    config: Option<Configuration>,
+    config: Option<Arc<Configuration>>,
     qpu: Option<qpu::Execution<'execution>>,
     qvm: Option<qvm::Execution>,
 }
 
-pub(crate) type Parameters<'a> = HashMap<&'a str, Vec<f64>>;
+pub(crate) type Parameters = HashMap<Box<str>, Vec<f64>>;
 
 impl<'executable> Executable<'executable, '_> {
     /// Create an [`Executable`] from a string containing a  [quil](https://github.com/quil-lang/quil)
@@ -81,9 +82,9 @@ impl<'executable> Executable<'executable, '_> {
     /// 1. `quil` is a string slice representing the original program to be run. The returned
     ///     [`Executable`] will only live as long as this reference.
     #[must_use]
-    pub fn from_quil(quil: &'executable str) -> Self {
+    pub fn from_quil<Quil: Into<Arc<str>>>(quil: Quil) -> Self {
         Self {
-            quil,
+            quil: quil.into(),
             shots: 1,
             readout_memory_region_names: None,
             params: Parameters::new(),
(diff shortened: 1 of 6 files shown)


qcs/src/executable.rs
```diff
@@ -81,9 +82,9 @@ impl<'executable> Executable<'executable, '_> {
     /// 1. `quil` is a string slice representing the original program to be run. The returned
     ///     [`Executable`] will only live as long as this reference.
     #[must_use]
-    pub fn from_quil(quil: &'executable str) -> Self {
+    pub fn from_quil<Quil: Into<Arc<str>>>(quil: Quil) -> Self {
```
nilslice, line 85: Maybe personal preference, but I tend to shy away from long generic names vs. single letter (`Q` over `Quil`), only because it can be confusing for folks less familiar with the concept & mistake it for a variable or something.

qcs/src/executable.rs
```diff
@@ -185,28 +186,24 @@ impl<'executable> Executable<'executable, '_> {
     /// ```
     ///
     /// [parametric compilation]: https://pyquil-docs.rigetti.com/en/stable/basics.html?highlight=parametric#parametric-compilation
-    pub fn with_parameter(
+    pub fn with_parameter<Param: Into<Box<str>>>(
```
nilslice, line 189: Similarly, here `P` over `Param`

qcs/src/qpu/execution.rs
```diff
@@ -90,11 +92,16 @@ impl<'a> Execution<'a> {
...
             trace!("Converting to Native Quil");
-            quilc::compile_program(quil, &isa, config)
-                .wrap_err("When attempting to compile your program to Native Quil")?
+            let thread_config = config.clone();
+            spawn_blocking(move || {
+                quilc::compile_program(&quil, &isa, &thread_config)
+                    .wrap_err("When attempting to compile your program to Native Quil")
+            })
+            .await
+            .map_err(|_| eyre!("Error in quilc thread."))??
```
nilslice, line 101: nit: `s/thread/task`, just to be consistent with the underlying runtime.

qcs/src/qpu/execution.rs
```diff
@@ -243,7 +260,12 @@ impl<'a> Execution<'a> {
                     })
             })
             .collect::<Result<Vec<f64>>>()?;
-        patch_values.insert(SUBSTITUTION_NAME, values);
+        // Convert back to the format that this library expects
+        let mut patch_values: Parameters = params
+            .into_iter()
+            .map(|(key, value)| (key.into(), value))
+            .collect();
+        patch_values.insert(SUBSTITUTION_NAME.into(), values);
```
nilslice, line 268: I don't think there is a test covering this, but I might have missed it. Might be good to add one, since based on the comments "this library expects" and "that quil-rs expects", we would benefit from those invariants being checked for regression over time.

nilslice (changes requested):
See note about testing the substitution / format things - maybe super minor and in that case no worries.. the renaming of the generic params is certainly personal preference, but might be worth considering since we also use exact or similar names as variables or types throughout (Parameters, quil, etc).

qcs/src/executable.rs
dbanty: I have a pretty strong preference for more descriptive generic names simply because I've found it frustrating to trace types for single letters, particularly when there are multiple generics it becomes pretty unreadable (e.g., in `tokio` their generic type for `response` tends to be `S` because `R` is taken up by `request`... or maybe it's the other way around, I can never remember).

For a short signature like this, it doesn't make much of a difference, but since I want consistency in the naming convention, I try to always stick to naming the generic after the thing it's typing.

qcs/src/qpu/execution.rs
dbanty:
```suggestion
            .map_err(|_| eyre!("Task running quilc did not complete."))??
```

qcs/src/qpu/execution.rs
dbanty: It's really a type conversion—this would not compile if it wasn't converting correctly. We have a `HashMap<Box<str>, Vec<f64>>` and `quil-rs` wants a `HashMap<&str, Vec<f64>>`. The comment is just to clarify the fact that this iteration looks kinda useless.

Is there a better comment I could leave to reduce confusion?

qcs/src/executable.rs
nilslice: Ok that's fair, yea not a major thing!

qcs/src/qpu/execution.rs
nilslice: I see -- that's good enough for me :) Thanks for clarifying.

nilslice (approved):

PR #63, opened 2022-04-04
feat: Replace `eyre` with `thiserror` for all errors.

### LINKED ISSUE #62: Stop exporting `eyre::Report` through pubic interfaces
This practice, while very easy to set up initially, is generally discouraged. Instead, we should define well-organized errors via `thiserror` which will more accurately tell users at runtime what went wrong. It's especially important to differentiate between an error resulting from the input (e.g., compilation problem) and a networking issue (e.g., quilc is not running).

dbanty:
Closes #62 

BREAKING CHANGE: Changed the error types returned from all public methods.

### DIFF
--- a/qcs/src/configuration/mod.rs
+++ b/qcs/src/configuration/mod.rs
@@ -9,7 +9,8 @@
 //! can set the [`PROFILE_NAME_VAR`] to select a different profile. You can also use
 //! [`SECRETS_PATH_VAR`] and [`SETTINGS_PATH_VAR`] to change which files are loaded.
 
-use eyre::{eyre, Result, WrapErr};
+use std::path::PathBuf;
+
 use futures::future::try_join;
 use serde::{Deserialize, Serialize};
 
@@ -18,6 +19,8 @@ use secrets::Secrets;
 pub use secrets::SECRETS_PATH_VAR;
 pub use settings::SETTINGS_PATH_VAR;
 use settings::{AuthServer, Pyquil, Settings};
+
+use crate::configuration::LoadError::AuthServerNotFound;
 
 mod path;
 mod secrets;
@@ -36,14 +39,42 @@ pub(crate) struct Configuration {
 /// Setting this environment variable will change which profile is used from the loaded config files
 pub const PROFILE_NAME_VAR: &str = "QCS_PROFILE_NAME";
 
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum RefreshError {
+    #[error("No refresh token is in secrets")]
+    NoRefreshToken,
+    #[error("Error fetching new token")]
+    FetchError(#[from] reqwest::Error),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum LoadError {
+    #[error("Expected profile {0} in settings.profiles but it didn't exist")]
+    ProfileNotFound(String),
+    #[error("Expected auth server {0} in settings.auth_servers but it didn't exist")]
+    AuthServerNotFound(String),
+    #[error("Failed to determine home directory. You can use an explicit path by setting the {env} environment variable")]
+    HomeDirError { env: String },
+    #[error("Could not open file at {path}")]
+    FileOpenError {
+        path: PathBuf,
+        source: std::io::Error,
+    },
+    #[error("Could not parse file at {path}")]
+    FileParseError {
+        path: PathBuf,
+        source: toml::de::Error,
+    },
+}
+
 impl Configuration {
     /// Attempt to load config files from ~/.qcs and create a Configuration object
     /// for use with qcs-api.
     ///
     /// # Errors
-    /// Errors are human-readable (using eyre) since generally they aren't recoverable at runtime.
-    /// Usually this function will error if one of the config files is missing or malformed.
-    pub(crate) async fn load() -> Result<Self> {
+    ///
+    /// See [`LoadError`].
+    pub(crate) async fn load() -> Result<Self, LoadError> {
         let (settings, secrets) = try_join(settings::load(), secrets::load()).await?;
         Self::new(settings, secrets)
     }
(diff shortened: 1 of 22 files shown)


qcs/src/qpu/quilc/isa/mod.rs
```diff
@@ -99,7 +120,7 @@ mod describe_compiler_isa {
                         second_f64,
                         F64Margin {
                             ulps: 1,
-                            epsilon: 0.0000001
+                            epsilon: 0.000_000_1
```
dbanty, line 123: There are a bunch of these adding separators to large constants... they're just here to get rid of excess red squiggles from my IDE so it's easier to see and fix real errors.

Allegedly that warning comes from Clippy but running Clippy didn't report them so 🤷

qcs/src/configuration/settings.rs
```diff
@@ -1,21 +1,23 @@
 use std::collections::HashMap;
 
-use eyre::{Result, WrapErr};
 use serde::{Deserialize, Serialize};
 
 use super::path::path_from_env_or_home;
+use super::LoadError;
 
 /// Setting this environment variable will change which file is used for loading settings
 pub const SETTINGS_PATH_VAR: &str = "QCS_SETTINGS_FILE_PATH";
```
nilslice, line 9: nit: in the docstring, it might be helpful to repeat the value for this.. because of the const, both are in caps & I had to think about it for more than 5 seconds: "Is it `$SETTINGS_PATH_VAR`, or `$QCS_SETTINGS_FILE_PATH`..."
```diff
- /// Setting this environment variable will change which file is used for loading settings
+ /// Setting the `QCS_SETTINGS_FILE_PATH` environment variable will change which file is used for loading settings
```

(Same for the other env vars defined by consts)

qcs/src/executable.rs
```diff
@@ -362,64 +347,141 @@ impl<'execution> Executable<'_, 'execution> {
...
-                    None => Err(source.into()),
-                }
-            }
-            Err(quil_err @ qpu::Error::Quil { .. }) => Err(Error::from(
-                eyre!(quil_err).wrap_err("When compiling for the QPU"),
-            )),
-        }
+        self.qpu = if let Err(ExecutionError::QpuUnavailable(_)) = response {
+            // Could mean maintenance which requires recompile
+            None
```
nilslice, line 363: Is this the only error that should cause `self.qpu = None` ? I assume the others would cause the request to fail earlier?

qcs/src/qpu/execution.rs
```diff
@@ -36,13 +36,105 @@ pub(crate) struct Execution<'a> {
...
+            runner::Error::Connection(_) => Self::QcsCommunication,
+            runner::Error::Bug(err) => Self::Bug(Bug::Qcs(format!("{:?}", err))),
+            runner::Error::Qpu(err) => Self::Qcs(err),
+        }
+    }
+}
+
+/// Errors that are likely due to bugs in this library or QCS.
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum Bug {
```
nilslice, line 97: nit: maybe classify these as `Internal` or something. Although Bug does make sense, I feel like an end-user might ask "if you knew this was a bug, why not fix it?", when really these are cases where further action needs to be taken on Rigetti's side due to unforeseen causes. 

Maybe I'm misinterpreting the intent here, but I take this as "you probably can't code your way out of this problem, and we need to change something internally".

qcs/src/executable.rs
dbanty: Other failures do not necessarily need to void the compilation cache (which is what setting this to `None` does). The `QpuUnavailable` error is the only one which signifies planned maintenance—which could mean the QPU settings changed.

qcs/src/configuration/mod.rs
```diff
@@ -36,14 +39,42 @@ pub(crate) struct Configuration {
...
+    #[error("Error fetching new token")]
+    FetchError(#[from] reqwest::Error),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum LoadError {
+    #[error("Expected profile {0} in settings.profiles but it didn't exist")]
+    ProfileNotFound(String),
+    #[error("Expected auth server {0} in settings.auth_servers but it didn't exist")]
+    AuthServerNotFound(String),
```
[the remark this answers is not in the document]
dbanty: I don't think the full format of the settings TOML file is documented anywhere publicly—this implementation was copied from pyquil. I think the settings files for users tend to be auto-generated in JupyterHub today.

This error will only ever be surfaced to users within a [`SettingsError`](https://github.com/rigetti/qcs-sdk-rust/blob/2236439acc57d89c1e29c7781aa8e0de2517c35d/qcs/src/executable.rs#L383) which has a bit more info for debugging.

qcs/src/configuration/settings.rs
dbanty: Yeah... I originally resisted repeating so that it wouldn't fall out of sync if we change the value but it's _very_ unlikely we'll change this (it's kept in line with pyquil today).

qcs/src/configuration/settings.rs
```diff
@@ -1,21 +1,23 @@
 use std::collections::HashMap;
 
-use eyre::{Result, WrapErr};
 use serde::{Deserialize, Serialize};
 
 use super::path::path_from_env_or_home;
+use super::LoadError;
 
 /// Setting this environment variable will change which file is used for loading settings
```
dbanty, line 8:
```suggestion
/// Setting the `QCS_SETTINGS_FILE_PATH` environment variable will change which file is used for loading settings
```

qcs/src/configuration/secrets.rs
```diff
@@ -1,21 +1,24 @@
 use std::collections::HashMap;
 
-use eyre::{Result, WrapErr};
 use serde::{Deserialize, Serialize};
 
+use crate::configuration::LoadError;
+
 use super::path::path_from_env_or_home;
 
 /// Setting this environment variable will change which file is used for loading secrets
```
dbanty, line 9:
```suggestion
/// Setting the `QCS_SECRETS_FILE_PATH` environment variable will change which file is used for loading secrets
```

nilslice (commented):
Generally looks great - just a couple thoughts/questions

qcs/src/qpu/execution.rs
nilslice: Yea, I totally agree with the approach - just not the term. But, it's just a name after all 👍

dbanty: I could call this `Unexpected` as well to match the top one? Maybe that better conveys the intent of the error—we don't expect this to ever be returned, not "we return this on purpose when there's a known bug".

dbanty: > I could call this Unexpected as well to match the top one? Maybe that better conveys the intent of the error—we don't expect this to ever be returned, not "we return this on purpose when there's a known bug".

I did that

nilslice: Thanks, @dbanty!

PR #122, opened 2022-08-09
chore: Switch from workspace API client to standalone

dbanty:


Shadow53 (approved):
Changes seem fine. API files were deleted as expected, and the QCS crate files were largely moved without changes. Since CI checks are passing, the minimal changes that were made don't seem to have broken anything. Approved.

PR #144, opened 2022-09-08
Setup distribution for python bindings

MarquessV:
# Description

The goal of this PR is to implement a GitHub workflow that can easily build wheels that are compatible with various permutations of platforms, architectures, and python versions. 

## TODO

- [X] ~Create build wheels step for MacOS~ [Done](https://github.com/rigetti/qcs-sdk-rust/actions/runs/3017737018)
- [X] ~Create build wheels step for Linux~ [Done](https://github.com/rigetti/qcs-sdk-rust/actions/runs/3018358655) (without manylinux, see note below)
- [X] ~Create build wheels step for Linux on ARM archs~ [Done](https://github.com/rigetti/qcs-sdk-rust/actions/runs/3024379974)
- [ ] ~Create build wheels step for Windows~ (Punting this, see below)
- [X] ~Create build sdist step~ (part of macos step)
- [X] ~Create step to publish all wheels to TestPyPI~
- [X] ~Update release step to use `maturin upload`~ [Done](https://github.com/rigetti/qcs-sdk-rust/actions/runs/3019024434)
- [X] ~Update Python package level README.md~
- [X] ~Get metadata and README uploading to PyPI~ [Done](https://test.pypi.org/project/qcs-sdk-python/) (fixed the typos that are there, they'll be gone when we release to real PyPi)
- [X] ~Automate prereleases on merge to main and use tags/workflow-dispatch to trigger full releases~
- [X] ~Update release step to point to real PyPI~
- [X] Add documentation

### manylinux and openssl

**Note**: leaving this here for posterity, but using the `rustls-tls` feature of `reqwest` allows us to build on manylinux images without any issue.

Our Rust code depends on `openssl-sys`, which in turn is a dependency of `reqwest`. The `openssl-sys` crate depends on development tools for OpenSSL to be installed, but the `manylinux` images don't have them. I am able to build on their image to install `openssl-devel` that enables these builds but:

1) This isn't possible with `messense/maturin-action@v1`, so we'd have to fork it or otherwise build our own action that did the same thing, but allowed the installation of additional packages on the manylinux image.

2) The implications of building wheels with openssl might compromise security of users. See [this comment](https://github.com/pypa/manylinux/issues/74#issuecomment-229795706) on why the manylinux maintainers don't simply add openssl to the image.

~For now, I have the wheels building directly on the linux host since it's easy enough to use apt to install `libssl-dev` before the build.~

Using the `rustls-tls` feature of `reqwest` solves the issue of (1) by eliminating the need for the openssl dependency. With that, I'm not sure how big of a concern (2) is anymore. Seems to me that it poses a similar risk to using any other third party package. 

### Windows
Builds on windows seem to be [failing](https://github.com/rigetti/qcs-sdk-rust/runs/8259691282?check_suite_focus=true) due to the `libzmq` package. In the interest of time, I'm going to create a followup issue to build Windows wheels. See #146


Shadow53:
As another note, you should be able to install the SSL development tools (and other packages) in the manylinux container. I think it's based on CentOS, so `yum install openssl-devel` should do it?

MarquessV:
> Depending on where reqwest surfaces as a dependency, we may be able to do something like this.
>
> reqwest = { version = "whatever", default-features = false, features = ["rustls-tls"] }
>
> This will disable the use of native-tls (OpenSSL) and use rustls instead, which will be statically compiled into the library. It does not fix the issue of if a vulnerability is found in our TLS provider, as something would need to be rebuilt either way, but it does mean (1) not installing extra stuff in CI and (2) being able to upgrade the TLS provider out of sync with a distribution's OpenSSL package.
>
> Another option is using the native-tls-vendored feature, which is the same as default but uses a vendored version of OpenSSL instead of the system one.

Didn't know about that feature and it's good to know that there is a statically linked option. I'll experiment with it.

> As another note, you should be able to install the SSL development tools (and other packages) in the manylinux container. I think it's based on CentOS, so yum install openssl-devel should do it?

Yeah, that is exactly what I've done locally with success. It's just that isn't possible with the maturin action since it doesn't give us the ability to do anything with the manylinux image they build on. It wouldn't be too hard to make a fork and add that, but I moreso wanted to align on (2) before moving forward with any particular option.

MarquessV:
Another option is to look towards the [cryptography package](https://github.com/pyca/cryptography/blob/main/.github/workflows/wheel-builder.yml#L34) for inspiration. They appear to use a custom image that includes openssl and being a crpytograhpy package, I'm sure they've put some thought into this issue. I'll have to look into the details of it when time allows.

Shadow53:
> Yeah, that is exactly what I've done locally with success. It's just that it isn't possible to add additional packages with the maturin action since it doesn't give us the ability to do anything inside the manylinux image they build on. It wouldn't be too hard to make a fork and add that, but I moreso wanted to align on (2) before moving forward with any particular option.

My mistake, I didn't realize you were directly using the action.

Looks like https://github.com/messense/manylinux-cross includes OpenSSL when not targeting `musl`, which the action uses for all linux targets _except_ x86(_64) :/. It also includes maturin, though, so shouldn't be too much more work to use the image instead of the action, maybe?

MarquessV:
Interestingly, we had a `rustls` feature enabled [on the reqwest dependency](https://github.com/rigetti/qcs-sdk-rust/pull/144/commits/4337d39c10ba468fa7c6bf85c773d6d3916548b9#diff-85c569622acbdbaa969d5d4f5257f224abde8df87639fff31e6bab15fbc8ff3eL25). No such feature exists, so I can only assume the intent was to use rustls-tls all along.

MarquessV:
> Interestingly, we had a `rustls` feature enabled [on the reqwest dependency](https://github.com/rigetti/qcs-sdk-rust/pull/144/commits/4337d39c10ba468fa7c6bf85c773d6d3916548b9#diff-85c569622acbdbaa969d5d4f5257f224abde8df87639fff31e6bab15fbc8ff3eL25). No such feature exists, so I can only assume the intent was to use rustls-tls all along.

`rustls-tls` works. Thanks for the suggestion @Shadow53. I prefer that solution overall, and I think it was what we were intending to use all along.

MarquessV:
Here is the [full workflow suceeding](https://github.com/rigetti/qcs-sdk-rust/actions/runs/3019190378) and the [resulting wheels on test.pypi](https://test.pypi.org/project/qcs-sdk-python/#files)

kalzoo:
Looks good! We'll want package metadata as well, a pyproject.toml should do the trick. That way the PyPI landing page will have more information: 

* https://python-poetry.org/docs/pyproject/
* https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/
* https://github.com/PyO3/maturin#python-metadata

The minimum seems like it'd be a description, readme, license, and link to the repo.

----

Beyond that, I'm tracking two things left for just prior to merge:

- Target real PyPI rather than test
- Change the publication trigger. In the comment you mention workflow-dispatch; any reason not to just run on push to `main`?

Also, CI seems to be jamming up on the main crate's tests; it doesn't seem to me like that's related to your changes here. If not, we can ignore that for now, although it'll need a fix later.

notmgsk:
@MarquessV Are we going to support both AMD64 _and_ ARM64 in this update? It may not be required immediately, but something we should eventually support.

notmgsk:
We might also want to include some Python unit tests?

MarquessV:
> Looks good! We'll want package metadata as well, a pyproject.toml should do the trick. That way the PyPI landing page will have more information:
> 
>     * https://python-poetry.org/docs/pyproject/
> 
>     * https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/
> 
>     * https://github.com/PyO3/maturin#python-metadata
> 
> 
> The minimum seems like it'd be a description, readme, license, and link to the repo.

I'll get on that. The README could definitely use some love, but there is already a `pyproject.toml`, so I'll have to figure out why that metadata isn't propagating. 

> Beyond that, I'm tracking two things left for just prior to merge:
> 
>     * Target real PyPI rather than test
> 
>     * Change the publication trigger. In the comment you mention workflow-dispatch; any reason not to just run on push to `main`?

The Rust SDK [is released via `workflow-dispatch`](https://github.com/rigetti/qcs-sdk-rust/blob/main/.github/workflows/release.yml#L4) so I was keeping in line with that pattern. Also, I believe in a conversation we had earlier this week we talked about favoring a button to trigger releases for our Python packages. That said, I don't feel super strongly about either approach, but I would want to make sure both the Rust and Python SDKs were released in the same way. Eventually, I was planning on tying those workflows together, but I thought it'd be better to keep them separate while we are still in RC. 

> Also, CI seems to be jamming up on the main crate's tests; it doesn't seem to me like that's related to your changes here. If not, we can ignore that for now, although it'll need a fix later.

Yeah, this seems to be happening on the [branch this is based off as well](https://github.com/rigetti/qcs-sdk-rust/runs/8257237538?check_suite_focus=true). I can reproduce locally too. I'll dig in and figure out why those tests are hanging. Though I'll track that separately from this PR.

MarquessV:
> @MarquessV Are we going to support both AMD64 _and_ ARM64 in this update? It may not be required immediately, but something we should eventually support.

I did setup ARM64 for the MacOS side. I'll look into it for the Linux side.

> We might also want to include some Python unit tests?

I thought we wanted to do this post RC? I agree that we definitely want them, but maybe after the API is a little more stable? I haven't put a ton of thought into it yet, but it seems non-trivial to setup stable unit tests for this project considering the Rust side will expect a real QPU. Seems like its worth tracking as an issue separate to this.

kalzoo:
> Also, I believe in a conversation we had earlier this week we talked about favoring a button to trigger releases for our Python packages.

Yep, we do want a button, but that's not a button for _all_ releases - it would be automatic for `rc` releases, and the button would say to CI, "cut a full release at this commit, not just an RC." If code is on `main`, it merits being published, even if only as an `rc` (unless we have some reason to add in a human decision - ideally that decision should have been made as part of PR approval though).

kalzoo:
> I did setup ARM64 for the MacOS side. I'll look into it for the Linux side.

I think it's wise to build for that, but that doesn't have to happen here. We should keep in mind that building wheels is a UX enhancement and is not required for the MVP - if a user's platform isn't supported, they can still use the package, they just need to build it from source and have Rust available. So supporting more architectures can wait for follow-on issues, IMO, now that you've gotten the overall pattern worked out.

MarquessV:
@kalzoo @notmgsk How are we feeling about the state of this now?

- Wheels for linux on arm architectures are now being built
- Metadata and a README is being published to PyPI.
- Created #146
- Created #147 

The last thread I'm tracking is what event to trigger these releases on. I'm leaning towards `workflow-dispatch`, at least in the short term, that way we have more granular control over how we release this while we are in heavy development, but I can be swayed.

MarquessV:
> > Also, I believe in a conversation we had earlier this week we talked about favoring a button to trigger releases for our Python packages.
> 
> Yep, we do want a button, but that's not a button for _all_ releases - publication would be automatic for `rc` releases, but the button would say to CI, "cut a full release at this commit, not just an RC." If code is on `main`, it merits being published, even if only as an `rc` (unless we have some reason to add in a human decision - ideally that decision should have been made as part of PR approval though).

Okay, I see, thanks for the clarification. I can start putting that together.

MarquessV:
@Shadow53 and I tried to get `knope` working, but it turns out it doesn't bump `project.version` in `pyproject.toml`, just `poetry.tools.version`. `Maturin` uses the former, so we need it to be bumped. We've created an [issue](https://github.com/knope-dev/knope/issues/271) to support `project.version` within the `knope` repo that I may take a stab at later. As a workaround, we did [try using `sed`](https://github.com/rigetti/qcs-sdk-rust/pull/144/commits/c959733142c85a5641f958a6c166d48d8938108e) to replace `project.version`, but `knope` doesn't support using the version as a variable if it is configured with multiple projects. There are probably other workarounds that we could employ, but it gets into hacky territory pretty quickly.

While we wait for that to be sorted, are we okay with triggering the workflow manually? I think the versioning and release automation should definitely be a requirement after RC, but until then, I think we can get by making commits manually and hitting a button until knope gets updated. cc: @kalzoo

kalzoo:
I think this is looking good, pending the CI pipeline and the knope fix. Sounded from a chat with @Shadow53 like this is something we'd be able to support tomorrow (Tuesday) with our own fork of `knope` (until it's upstreamed) - if that's the case, let's get that in first. 

If it looks like it would drag out past tomorrow, let's ship this without auto-versioning and add it later, like you suggest.

MarquessV:
Got the knope automation working. 

On a push to main, the `knope prerelease` workflow will push a commit [like this](https://github.com/rigetti/qcs-sdk-rust/pull/144/commits/0b0ae5ae5e2c136126d016cbf3673ee96c51f8e4) and create a [corresponding tag](https://github.com/rigetti/qcs-sdk-rust/releases/tag/python%2Fv0.2.0-rc.0), which will trigger the `Python Release` workflow. 

I've reverted the commits knope made against this branch, so once this and #145 have been merged into main, it should repeat a commit similar to the one seen above and we should see a prerelease of `qcs-sdk-rust@v0.2.0rc0` automatically kick off (this will require that I delete the `python/v0.2.0rc0` tag so `knope` doesn't fail by creating a duplicate tag). We may also want to manually cleanup the resulting changelogs since we haven't been following the conventional commit syntax leading up to the knope changes.

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,115 @@
...
+          python-version: ${{ matrix.python-version }}
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: aarch64-apple-darwin
+          profile: minimal
+          default: true
+      - name: Build
+        run: cargo build --release
```
Shadow53, line 27: I don't think this is necessary for building wheels. Is there a specific reason this is here?

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,115 @@
...
+      - name: Build wheels - x86_64
+        uses: messense/maturin-action@v1
+        with:
+          target: x86_64
+          args: -i python --release --manifest-path crates/python/Cargo.toml --out dist --sdist
+      - name: Install wheel - x86_64
+        run: |
+          pip install qcs-sdk-python --no-index --find-links dist --force-reinstall
+      - name: Build wheels - universal2
+        if: ${{ matrix.python-version >= '3.8' || matrix.python-version == '3.10' }}
```
Shadow53, line 37: Frankly, this could be `!= 3.7` instead. then above could then be `== 3.7`, since we don't need both x86_64 and universal2 wheels for a given Python version.

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,115 @@
...
+        uses: messense/maturin-action@v1
+        with:
+          args: -i python --release --universal2 --manifest-path crates/python/Cargo.toml --out dist
+      - name: Install wheel - universal2
+        run: |
+          pip install qcs-sdk-python --no-index --find-links dist --force-reinstall
```
Shadow53, line 43: I assume installing is just to make sure it doesn't break for some reason?

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,115 @@
...
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: aarch64-apple-darwin
+          profile: minimal
+          default: true
```
Shadow53, line 25: Either this should be when `matrix.python-version != '3.7'` or there should be a separate step for building an ARM64 wheel for 3.7.

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,115 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        profile: minimal
+        default: true
+    - name: Build
+      run: cargo build --release
```
Shadow53, line 63: Not sure why this is necessary, same as above

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,115 @@
...
+        target: x86_64
+        manylinux: 2014
+        args: -i ${{ matrix.python-version }} --release --manifest-path crates/python/Cargo.toml --out dist
+    - name: Upload wheels
+      uses: actions/upload-artifact@v2
+      with:
+        name: wheels
+        path: dist
+
+  linux-arm:
```
Shadow53, line 79: Why not merge this with the x86_64 matrix?

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,115 @@
...
+
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    needs: [ macos, linux, linux-arm ]
+    steps:
+      - uses: actions/download-artifact@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
```
Shadow53, line 107: Is this necessary when using `maturin` to upload?

crates/python/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "qcs-sdk-python"
 description = "Python bindings to qcs-sdk-rust"
-version = "0.1.2"
+version = "0.1.0"
```
Shadow53, line 4: Not sure if this is intended, but the version number went *down*

crates/python/pyproject.toml
```diff
@@ -1,17 +1,43 @@
...
+
+# PEP 621 specifies the [project] table as the source for project metadata. However, Poetry only supports [tool.poetry]
+# We can remove this table once this issue is resolved: https://github.com/python-poetry/poetry/issues/3332
 [tool.poetry]
 name = "qcs-sdk-python"
-version = "0.1.0-rc1"
+version = "0.1.0"
 description = "Python interface for the QCS Rust SDK"
-license = "Apache-2.0"
+readme = "README.md"
```
Shadow53, line 29: Did you intend to remove the `license` item?

Shadow53 (changes requested):
Left some notes on the wheel-building stuff.

crates/python/README.md
```diff
@@ -0,0 +1,8 @@
+# QCS SDK Python
+
+`qcs-sdk-python` is an interface to Rigetti's [Quantum Cloud Services](https://docs.rigetti.com/qcs/) (QCS). Allowing users
+to compile Quil programs using [quilc](https://github.com/rigetti/quilc) and run them on real quantum processors.
+
+While this package could be used directly, [pyQuil](https://pypi.org/project/pyquil/) offers more functionality and a 
+higher-level interface for building and executing Quil programs. This package is still in early development and breaking
+changes should be expected. Eventually, some fucntionality of `pyQuil` will be powered by this package.
```
kalzoo, line 8:
```suggestion
# QCS SDK Python

⚠️ In Development

`qcs-sdk-python` provides an interface to Rigetti [Quantum Cloud Services](https://docs.rigetti.com/qcs/) (QCS), allowing users
to compile and run Quil programs on Rigetti quantum processors. Internally, it is powered by the [QCS Rust SDK](https://github.com/rigetti/qcs-sdk-rust).

While this package can be used directly, [pyQuil](https://pypi.org/project/pyquil/) offers more functionality and a 
higher-level interface for building and executing Quil programs. This package is still in early development and breaking changes should be expected between minor versions.
```

crates/python/pyproject.toml
```diff
@@ -1,17 +1,43 @@
...
 authors = [
+	"Rigetti Computing <softapps@rigetti.com>",
```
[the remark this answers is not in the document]
MarquessV: I borrowed it from [pyquil](https://github.com/rigetti/pyquil/blob/master/pyproject.toml#L5)

.github/workflows/release-python.yml
MarquessV: At first I thought it might make debugging easier by having discrete steps for building the Rust code and the wheels but it wasn't particularly helpful. I'll remove it.

.github/workflows/release-python.yml
MarquessV: Yes, exactly.

.github/workflows/release-python.yml
MarquessV: I wasn't sure if the cross compilation would have it's own quirks or not, so I kept them separate. I can try consolidating them now.

.github/workflows/release-python.yml
MarquessV: Recommended for the `setup-python` action. If not set and no `.python-version` file exists, it will fallback on whatever is in the PATH for the runner, which can vary and change unexpectedly. 

https://github.com/actions/setup-python#basic-usage

crates/python/pyproject.toml
MarquessV: Correct, `maturin upload` uses the metadata in the `[project]` table. The metadata in `[tools.poetry]` is just what `poetry` needs to not complain. I removed everything else to minimize duplication.

.github/workflows/release-python.yml
MarquessV: I will update the condition, thanks for the suggestion. 

For the `universal2` wheels. `pip` started supporting them in `20.3` so I was building them for compatibilities sake. `20.3` is coming up on 2 years old now, so I can see an argument for them not being relevant anymore. That said, they were easy enough to build so I left it in.

crates/python/Cargo.toml
MarquessV: That was the result of me fussing around with the version numbers to test knope. I reverted back to 0.1.4, which was the latest version to be released on TestPyPI. 

`knope` is bumping to `0.2.0` whether it's going from `0.1.0` [or `0.1.4`](https://github.com/rigetti/qcs-sdk-rust/pull/144/commits/f2f422808dd98742b86c64439c7d3d3ff999f3db). Not sure what rule it's following that is triggering a minor bump here.

With the `knope` configuration we have, versions are tracked separately. For example, if all commits after the previous release are scoped to `(rust)`, then a new release would only bump the Rust crate.

We can choose to track them together. I think the intent was to track them separately since it is possible to make a breaking change in one package and not the other.

.github/workflows/release-python.yml
MarquessV: Removed here as well

.github/workflows/release-python.yml
MarquessV: 3.7 doesn't support apple silicon so I can try and make this conditional. Maybe you have a clever way to do it, but I can't find a way to add conditionals to action inputs.

.github/workflows/release-python.yml
Shadow53: I meant the entire `setup-python` action. I don't think `maturin` uses Python at all.

.github/workflows/release-python.yml
Shadow53: Wouldn't this work?

```suggestion
      - name: Install Rust
        uses: actions-rs/toolchain@v1
        if: matrix.python-version != '3.7'
        with:
          toolchain: stable
          target: aarch64-apple-darwin
          profile: minimal
          default: true
```

.github/workflows/release-python.yml
Shadow53: I'm confused what you are saying. I _think_ you mean that you see an argument for `x86_64`-only not being relevant anymore?

My primary reason for suggesting doing only `universal2` is to decrease CI time, since `macos` (and `windows`) cost more than `linux` (not sure if we're paying for CI or getting it free).

Shadow53 (commented):
Clarified some things

.github/workflows/release-python.yml
MarquessV: Let me clarify. There are two reasons to build for both `x86_64` and `universal2`:

1) `universal2` is basically just a `x86_64` and `arm64` binary in one, meaning it's twice the size.

2) Since `universal2` is relatively new, it's only supported on `pip >= 20.3`. If we don't build x86_64 wheels, then users on a previous version will have to update `pip`.

What I was trying to say is that you could argue against (2) since `pip` 20.3 is almost 2 years old now.

In short the tradeoff is either:

Support smaller binaries and older versions of pip for `x86_64` users OR decrease CI time.

I'll go ahead and skip building `x86_64` wheels for >= `3.8`. It's an experience improvement that no one may care about. If someone raises it, we can easily add it back in.

.github/workflows/release-python.yml
MarquessV: That seems to work, thank you.

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,90 @@
...
+    runs-on: macos-11
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.8', '3.9', '3.10']
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Rust
```
notmgsk, line 21: Could you leave a comment here, for future-us, that describes
* why we don't include 3.7 here;
* why only 3.7 is built for x86_64; and
* why we don't include 3.7 for universal2.

I think I know the answer, but am likely to forget :D

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,90 @@
...
+        uses: actions/upload-artifact@v2
+        with:
+          name: wheels
+          path: dist
+  linux:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.8', '3.9', '3.10']
+        target: [x86_64, aarch64, armv7]
```
notmgsk, line 53: Why armv7? Not saying we shouldn't include it, but (IIUC) it's not a particularly prevelant architecure.

crates/python/Cargo.toml
```diff
@@ -17,9 +17,12 @@ crate-type = ["cdylib"]
...
-pyo3 = { version = "0.16.0", features = ["extension-module", "abi3-py37"] }
+pyo3 = { version = "0.16", features = ["extension-module"] }
 pyo3-asyncio = { version = "0.16", features = ["tokio-runtime"] }
 pythonize = "0.16"
 tokio = "1.4"
 qcs-api = "0.2.1"
 serde_json = "1.0.85"
+
+[build-dependencies]
+pyo3-build-config = { version = '0.16' }
```
notmgsk, line 28: single quotes 🤢

notmgsk:
@MarquessV I guess this should be out of draft now?

notmgsk (approved):
Looks good to me. The CI is unhappy though.

.github/workflows/release-python.yml
MarquessV: `armv7` runs on `armv8` as well, and you see it architecture on devices like the Raspberry PI. I believe some servers use it as well. I think we can just remove it until someone asks for it.

crates/python/Cargo.toml
MarquessV: Dang, caught me.

.github/workflows/release-python.yml
MarquessV: Done

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,90 @@
...
+          args: -i python --release --manifest-path crates/python/Cargo.toml --out dist --sdist
+      - name: Build wheels - universal2
+        # universal2 supports both x86_64 and aarch64 so every Python > 3.7 only needs this wheel
+        if: matrix.python-version != '3.7'
+        uses: messense/maturin-action@v1
+        with:
+          args: -i python --release --universal2 --manifest-path crates/python/Cargo.toml --out dist
+      - name: Install wheel
+        run: |
+          pip install qcs-sdk-python --no-index --find-links dist --force-reinstall
```
Shadow53, line 42: Looking at this again, are we sure this is installing the new wheel? I don't see any file path being passed in.

I'm not terribly familiar with how pip works, so I suppose it could be finding it automatically, but I couldn't find a CI job to verify with.

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,90 @@
...
+    steps:
+      - uses: actions/download-artifact@v2
+      - name: Publish to PyPi
+        env:
+          MATURIN_USERNAME: ${{ secrets.PYPI_USERNAME }}
+          MATURIN_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        uses: messense/maturin-action@v1
+        with:
+          command: upload 
+          args: --skip-existing wheels/*
```
Shadow53, line 90: I don't see a place where we also upload an sdist. Is this something we plan to do?

Shadow53 (approved):
Looks good, only real request is to make sure we also are uploading a working sdist

.github/workflows/release-python.yml
MarquessV: Yes. The wheels are output to the `dist` directory and the `--no-index --find-links dist` flag is whats telling pip to look in that directory for the wheels and not the package index.

.github/workflows/release-python.yml
MarquessV: The `sdist` is built [here](https://github.com/rigetti/qcs-sdk-rust/pull/144/files#diff-340cae5fb7f7deb7c69851b32c3b128dbd935c9a347a37aa1f648559447a7792R33) and stored in the same directory as the wheels, so it gets uploaded all the same. You can see it in the [uploaded files on TestPyPI](https://test.pypi.org/project/qcs-sdk-python/#files)

PR #145, opened 2022-09-09
feat: python bindings

notmgsk:
This MR introduces Python bindings for translation, compilation, and execution using the Rust SDK. Namely, we expose the following functions:
- `compile`: compiles, using quilc, a Quil program into a "native" Quil program (i.e. a program that uses instructions supported by the target hardware);
- `translate`: creates a binary executable from a native Quil program;
- `rewrite_arithmetic`: munges parametric arithmetic so that it can be executed on hardware;
- `submit`: submits the executable for execution on hardware
- `retrieve_results:` collects results of execution.

Names are not set in stone. Make your suggestions now. In a follow-on change, we will support passing more Rust types across the language boundary (this MR passes around base types like strings and numbers).

PR #153, opened 2022-09-14
ci: use knope 0.6.1 in release dry run and secrets.PAT to auth with gh

MarquessV:
Fixes for the two workflows that failed after merging #145

[Release Dry Run](https://github.com/rigetti/qcs-sdk-rust/actions/runs/3055481949/jobs/4928586660) failed because it was still pointing to an old version of knope (was 0.4.1, now 0.6.1)

[Prepare Release](https://github.com/rigetti/qcs-sdk-rust/actions/runs/3055481949/jobs/4928586660) failed because `GITHUB_TOKEN` can't be used to push to a protected branch (and can't be configured to do so safely). I took inspiration from the release workflow and use the already configured `secrets.PAT` instead.

While looking at the `Release` workflow, I realized it should probably be following the same pattern for releases as the Python package, so I updated it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,9 @@
-name: Release
+name: Release Rust
 
 on:
-  workflow_dispatch
+  push:
+    tags:
+      - 'lib/v**'
 
 jobs:
   release:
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -1,7 +1,9 @@
-name: Release
+name: Release Rust
```
MarquessV, line 1: @Shadow53 do we need to do anything with this workflow so that it publishes the `lib` package correctly, since it's now a crate in part of a workspace? Should we by excluding the python from being published somehow?

notmgsk: I think you'll want to specify `path: ./crates/lib` per https://github.com/katyo/publish-crates#inputs

PR #154, opened 2022-09-15
Fix prepare release loop

MarquessV:
On push to main the `Prepare Release` workflow gets triggered and uses `knope` to create and push a release commit to main, which the triggers the `Prepare Release` workflow which uses `knope` to... 🙃 

The fix here is to update the `knope` configuration to add `[skip ci]` to its commit messages. That way [GitHub actions will skip any push triggered workflows](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs). 

**Without `[skip ci]`:**
image

**With `[skip ci]`:**
SCR-20220915-gtv


notmgsk:
Good catch!

notmgsk:
Still draft?

Shadow53:
A couple thoughts, but overall looks good.

1. Should we be checking for `[skip ci]` instead of `skip ci` to prevent false positives?
2. Is there a way to set the condition once, like into a variable, so we don't have to copy/paste any changes?

(2) isn't too important since I don't see much changing after this gets merged, but I think (1) is worth doing.

MarquessV:
@Shadow53 GitHub sorted (1) and (2) for us: https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

Shadow53 (approved):
LGTM, assuming this is still confirmed to work correctly.

PR #158, opened 2022-09-20
ci: trigger releases on release events instead of tags being pushed

MarquessV:
`[skip ci]` applies to _all_ push events, including tag creation. This means that the tags `knope` creates don't trigger the workflows, since it pushes a `[skip ci]` commit right before.  

Setting these workflows up to run on `release` events instead. The `published` event triggers on pre-release and release creation, as well as when pre-releases are promoted to full releases. I also added `workflow_dispatch` as an option just so we have the flexibility of kicking these off on our own terms.

I've manually verified these with [act](https://github.com/nektos/act). 

```sh
act release --dryrun -e payload.json
```

Where `payload.json` is mocking the [release webhook payload](https://docs.github.com/en/rest/releases/releases#get-a-release) with `tag_name`
```json
{
	"release": {
		"tag_name": "lib/v1.0.0"
	}
}
```

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   macos:
-    if: ${{ github.event_name == "workflow_dispatch" || startsWith(github.event.release.tag_name, "python/v") }}
+    if: ${{ github.event_name == 'workflow_dispatch' || startsWith(github.event.release.tag_name, 'python/v') }}
     runs-on: macos-11
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-python.yml
```diff
@@ -1,12 +1,15 @@
...
-      - 'python/v**'
+  release:
+    types: [published]
+  workflow_dispatch:
+    description: "Manually publish release"
+
 
 jobs:
   macos:
+    if: ${{ github.event_name == "workflow_dispatch" || startsWith(github.event.release.tag_name, "python/v") }}
```
MarquessV, line 12: This condition makes sure the workflow is only ran when a release for the respective crate has been created. Without it, a release created for the library crate would trigger this workflow and vice-versa.
 
Unfortunately we can't set this condition at the workflow level, so some duplication is required. Open to feedback and ideas on how to resolve that.

Shadow53:
Can't confirm this, but I think you can exit the workflow early with a job that succeeds only if the tag matches. Something like:

```
jobs:
    is-python:
        if: tag is for python
        steps:
            - echo "Run other jobs"
    macos:
        needs: is-python
    # ...
    linux:
        needs: is-python
```

I *think* a skipped job counts as not succeeded for the purposes of dependencies.

MarquessV:
@Shadow53 that works! Thanks for the suggestion.

notmgsk (approved):

PR #171, opened 2022-10-06
feat: support gRPC translation and execution

notmgsk:
This MR introduces support for Rigetti's gRPC APIs. Prior to this, translation was performed using a REST API (api.qcs.rigetti.com), and execution was performed using the bespoke RPCQ-based API. Both services now provide gRPC alternatives, which will be the standard APIs going forward.

The change should be largely transparent to any user of the library. The translation gRPC endpoint is read from the QCS Client configuration (`~/.qcs/settings.toml`) under each profile's `grpc_api_url` key. If the key is not present, it will take a default value (currently `https://legacy.grpc.qcs.api.rigetti`).

RPCQ remains to support communication with quilc.


crates/lib/src/execution_data.rs
```diff
@@ -18,3 +23,192 @@ pub struct ExecutionData {
...
+                    readout_values.resize(index + 1, None);
+                }
+                readout_values[index] = Some(values.clone());
+            }
+        }
+
+        Ok(match readout_values.len() {
+            0 => None,
+            _ => Some(readout_values),
+        })
```
Shadow53, line 62: Optional, just a couple alternative ways to write this.

```suggestion
        Ok((!readout_values.is_empty()).then_some(readout_values))
```

```suggestion
        Ok(if readout_values.is_empty() {
            None
        } else {
            Some(readout_values)
        })
```

crates/lib/src/execution_data.rs
```diff
@@ -18,3 +23,192 @@ pub struct ExecutionData {
...
+        let result = readout_values
+            .iter()
+            .flat_map(|(readout_name, values)| {
+                readout_mappings
+                    .iter()
+                    .filter(|(_, program_alias)| *program_alias == readout_name)
+                    .map(|(program_name, _)| program_name.clone())
+                    .map(parse_readout_register)
+                    .map(|reference| Ok((reference?, values.clone())))
+                    .collect::<Result<Vec<_>, MemoryReferenceParseError>>()
```
Shadow53, line 82: Shouldn't need to specify the error type.

```suggestion
                    .collect::<Result<Vec<_>, _>>()
```

crates/lib/src/execution_data.rs
```diff
@@ -18,3 +23,192 @@ pub struct ExecutionData {
...
+}
+
+impl From<HashMap<MemoryReference, ReadoutValues>> for ReadoutMap {
+    fn from(map: HashMap<MemoryReference, ReadoutValues>) -> Self {
+        Self(map)
+    }
+}
+
+/// The result of executing an [`Executable`][crate::Executable] via
+/// [`Executable::execute_on_qpu`][crate::Executable::execute_on_qpu].
```
Shadow53, line 99: If linking to an item by path, the path should be in parentheses: https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

```suggestion
/// The result of executing an [`Executable`](crate::Executable) via
/// [`Executable::execute_on_qpu`](crate::Executable::execute_on_qpu).
```

Shadow53: Also applies below. Might be worth doing a. project-wide find for `][` and fixing all the results.

crates/lib/src/lib.rs
```diff
@@ -1,20 +1,19 @@
-#![deny(clippy::all)]
-#![deny(clippy::pedantic)]
+//#![deny(clippy::all)]
+//#![deny(clippy::pedantic)]
 #![deny(clippy::cargo)]
 #![allow(clippy::multiple_crate_versions)] // This should be enforced by cargo-deny
-#![deny(missing_docs)]
+//#![deny(missing_docs)]
 #![forbid(unsafe_code)]
```
Shadow53, line 6: These should be re-enabled and any lints triggered, fixed.

crates/lib/src/qpu/client.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    pub(crate) async fn get_controller_client(
+        &self,
+        quantum_processor_id: &str,
+    ) -> Result<ControllerClient<RefreshService<Channel>>, GrpcEndpointError> {
+        self.get_controller_endpoint(quantum_processor_id)
+            .await
+            .map(get_channel)
+            .map(|channel| wrap_channel_with(channel, self.get_config()))
```
Shadow53, line 45: It's times like this that I wish Rust had currying

crates/lib/src/qpu/execution.rs
```diff
@@ -219,99 +127,60 @@ impl<'a> Execution<'a> {
...
+    pub(crate) async fn submit(&mut self, params: &Parameters) -> Result<JobHandle, Error> {
+        // todo: return / persist readout sources
+        let EncryptedTranslationResult { job, readout_map } = translate(
+            self.quantum_processor_id,
+            &self.program.to_string().0,
+            self.shots.into(),
+            self.client.as_ref(),
+        )
+        .await?;
+        println!("done translating...");
```
Shadow53, line 146: Debug print?

crates/lib/src/qpu/execution.rs
```diff
@@ -219,99 +127,60 @@ impl<'a> Execution<'a> {
...
-        &mut self,
-        params: &Parameters,
-        readouts: &[&str],
-        config: &Configuration,
-    ) -> Result<JobId, Error> {
-        let qcs = self.refresh_qcs(readouts, config).await?;
-        let qcs_for_thread = qcs.clone();
+    /// needs to also return readout_sources
+    pub(crate) async fn submit(&mut self, params: &Parameters) -> Result<JobHandle, Error> {
+        // todo: return / persist readout sources
```
Shadow53, line 138: Still applies?

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -78,7 +77,7 @@ pub struct NativeQuil(String);
 
 impl NativeQuil {
     /// Cast a String to `NativeQuil` without checking or transforming it via `quilc`.
-    pub(super) fn assume_native_quil(quil: String) -> Self {
+    pub fn assume_native_quil(quil: String) -> Self {
```
Shadow53, line 80: Is this safe to make public?

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -195,7 +195,7 @@ mod tests {
         let output = compile_program(
             "MEASURE 0",
             TargetDevice::try_from(qvm_isa()).expect("Couldn't build target device from ISA"),
-            &Configuration::default(),
+            &QcsClient::with_config(ClientConfiguration::default()),
```
Shadow53, line 198: Is this different from `&QcsClient::default()`?

Shadow53: Also, IIRC, a default `ClientConfiguration` doesn't have any credentials associated with it.

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -212,16 +212,17 @@ MEASURE 1 ro[1]
 
     #[tokio::test]
     async fn run_compiled_bell_state_on_qvm() {
-        let config = Configuration::load().await.unwrap_or_default();
+        let config = ClientConfiguration::load().await.unwrap_or_default();
+        let client = QcsClient::with_config(config);
```
Shadow53, line 216: This will require `#[derive(Default)]` on `QcsClient`

```suggestion
        let client = QcsClient::load().await.unwrap_or_default();
```

crates/lib/src/qpu/runner.rs
```diff
@@ -9,132 +6,128 @@ use num::complex::Complex32;
...
+    params: &Parameters,
+) -> JobExecutionConfiguration {
+    let memory_values = params
+        .iter()
+        .map(|(str, value)| {
+            (
+                str.as_ref().into(),
+                DataValue {
+                    value: Some(Value::Real(RealDataValue {
+                        data: value.to_vec(),
```
Shadow53, line 34: Not sure if `value: &Vec<_>`. If so, there's a chance `value.clone()` would be more performant here.

crates/lib/src/qpu/runner.rs
```diff
@@ -9,132 +6,128 @@ use num::complex::Complex32;
...
     client
-        .run_request(&RPCRequest::from(&get_buffers_request))
-        .map_err(Error::from)
+        .get_controller_client(quantum_processor_id)
+        .await?
+        .get_controller_job_results(request)
+        .await?
+        .get_ref()
+        .result
+        .to_owned()
```
Shadow53, line 99: This seems redundant. Is there not a way to get the owned value directly?

crates/python/src/lib.rs
```diff
@@ -10,24 +11,31 @@ create_exception!(qcs, InvalidConfigError, PyRuntimeError);
...
 create_exception!(qcs, CompilationError, PyRuntimeError);
+create_exception!(qcs, RewriteArithmeticError, PyRuntimeError);
 
 #[pyfunction]
 fn compile(py: Python<'_>, quil: String, target_device: String) -> PyResult<&PyAny> {
     let target_device: TargetDevice = serde_json::from_str(&target_device)
         .map_err(|e| CompilationError::new_err(e.to_string()))?;
     pyo3_asyncio::tokio::future_into_py(py, async move {
-        let config = Configuration::load()
+        // TODO This doesn't need to be async.
```
Shadow53, line 21: Does this still apply?

crates/python/src/lib.rs
```diff
@@ -10,24 +11,31 @@ create_exception!(qcs, InvalidConfigError, PyRuntimeError);
...
+        let config = ClientConfiguration::load()
             .await
             .map_err(|e| InvalidConfigError::new_err(e.to_string()))?;
-        let result = api::compile(&quil, target_device, &config)
+        let client = QcsClient::with_config(config);
```
Shadow53, line 25: Can just use `QcsClient::load()` instead

crates/python/src/lib.rs
```diff
@@ -10,24 +11,31 @@ create_exception!(qcs, InvalidConfigError, PyRuntimeError);
...
 }
 
 #[pyfunction]
 fn rewrite_arithmetic(py: Python<'_>, native_quil: String) -> PyResult<PyObject> {
-    let result = api::rewrite_arithmetic(&native_quil).map_err(TranslationError::new_err)?;
+    let native_program = NativeQuil::assume_native_quil(native_quil)
+        .try_into()
+        .map_err(TranslationError::new_err)?;
+    let result = api::rewrite_arithmetic(native_program)
+        .map_err(|e| RewriteArithmeticError::new_err(e.to_string()))?;
```
Shadow53, line 38: I believe this takes anything implementing `Display`

```suggestion
        .map_err(RewriteArithmeticError::new_err)?;
```

crates/python/src/lib.rs
```diff
@@ -57,10 +65,11 @@ fn translate(
     quantum_processor_id: String,
 ) -> PyResult<&PyAny> {
     pyo3_asyncio::tokio::future_into_py(py, async move {
-        let config = Configuration::load()
+        let config = ClientConfiguration::load()
             .await
             .map_err(|e| InvalidConfigError::new_err(e.to_string()))?;
-        let result = api::translate(&native_quil, num_shots, &quantum_processor_id, &config)
+        let client = QcsClient::with_config(config);
```
Shadow53, line 71: `QcsClient::load`

crates/python/src/lib.rs
```diff
@@ -78,10 +87,11 @@ fn submit(
     quantum_processor_id: String,
 ) -> PyResult<&PyAny> {
     pyo3_asyncio::tokio::future_into_py(py, async move {
-        let config = Configuration::load()
+        let config = ClientConfiguration::load()
             .await
             .map_err(|e| InvalidConfigError::new_err(e.to_string()))?;
-        let job_id = api::submit(&program, patch_values, &quantum_processor_id, &config)
+        let client = QcsClient::with_config(config);
```
Shadow53, line 93: `QcsClient::load`

crates/python/src/lib.rs
```diff
@@ -94,11 +104,12 @@ fn retrieve_results(
...
     quantum_processor_id: String,
 ) -> PyResult<&PyAny> {
-    pyo3_asyncio::tokio::future_into_py(py, async move {
-        let config = Configuration::load()
+    pyo3_asyncio::tokio::local_future_into_py(py, async move {
+        let config = ClientConfiguration::load()
             .await
             .map_err(|e| InvalidConfigError::new_err(e.to_string()))?;
-        let results = api::retrieve_results(&job_id, &quantum_processor_id, &config)
+        let client = QcsClient::with_config(config);
```
Shadow53, line 111: `QcsClient::load`

crates/lib/Cargo.toml
```diff
@@ -21,6 +21,9 @@ lazy_static = "1.4.0"
 log = "0.4.17"
 num = "0.4.0"
 qcs-api = "0.2.1"
+qcs-api-client-common = "0.2.3"
+qcs-api-client-openapi = "0.3.4"
+qcs-api-client-grpc = { version = "0.2.4", features = ["server"] }
```
Shadow53, line 26: The feature is only used by tests. 

Change this to `qcs-api-client-grpc = "0.2.4"`

Then under `dev-dependencies`, put

```
qcs-api-client-grpc = { version = "0.2.4", features = ["server"] }
```

You might be able to use `version = "*"`, which would avoid duplicating version numbers. Not sure if that will work correctly, though

Shadow53 (changes requested):

crates/lib/src/execution_data.rs
```diff
@@ -18,3 +23,192 @@ pub struct ExecutionData {
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct ExecutionDataQPU {
+    /// The data of all readout data that were read from
+    /// (via [`Executable::read_from`][crate::Executable::read_from]). Key is the name of the
+    /// register, value is the data of the register after execution.
+    pub readout_data: ReadoutMap,
+    /// The time it took to execute the program on the QPU, not including any network or queueing
+    /// time. If paying for on-demand execution, this is the amount you will be billed for.
+    ///
+    /// This will always be `None` for QVM execution.
```
kalzoo, line 109: This line no longer applies

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -98,7 +97,7 @@ impl AsRef<str> for NativeQuil {
 /// A wrapper around [`Program`] which indicates it has been converted to `NativeQuil` (has been run
 /// through `quilc` and therefore is suitable to use on QPUs.
 #[derive(Debug, PartialEq, Clone)]
-pub(super) struct NativeQuilProgram(Program);
```
kalzoo, line 101: I don't understand why this struct is needed - a program is a program, and what is "native" is actually up to the remote API to determine. Errors should make it clear when a program is native or not, and so I think this type wrapper adds noise rather than signal.

kalzoo (changes requested):
This needs:

- A description
- Lints re-enabled and fixed

Substantively, it looks fine - I think we'll want to get away from the `&str -> String` public API but that can be a later PR, and this is consistent with the existing API.

I've marked it as draft in the meantime. Please tag me when this is ready for review.

crates/python/src/lib.rs
notmgsk: Doesn't seem to work out, but maybe I'm missing something. LMK if you have success.

crates/lib/src/qpu/quilc/mod.rs
notmgsk: There is no `QcsClient::default()`, and it's not much of a concern that there are no credentials since this test doesn't need them. But I'll replace this with `QcsClient::load()` anyway.

crates/lib/src/qpu/runner.rs
notmgsk: Can you point me to something to read on why cloning is more performant?

crates/lib/src/qpu/runner.rs
notmgsk: Yeah, I think we can do `get_controller_job_results(request).await?.into_inner()`

crates/lib/src/qpu/quilc/mod.rs
notmgsk: Agreed. It was added long ago (unrelated to this MR or the Python bindings MRs) but I'll remove it anyhow.

crates/lib/src/execution_data.rs
notmgsk: It doesn't? From `executable.rs`
```rust
    pub async fn execute_on_qvm(&mut self) -> ExecuteResultQVM {
        let config = self.get_client().await?.get_config();

        let mut qvm = if let Some(qvm) = self.qvm.take() {
            qvm
        } else {
            qvm::Execution::new(&self.quil)?
        };
        let result = qvm
            .run(self.shots, self.get_readouts(), &self.params, &config)
            .await;
        self.qvm = Some(qvm);
        result
            .map_err(Error::from)
            .map(|registers| ExecutionDataQVM {
                registers,
                duration: None,
            })
    }
```

crates/lib/src/qpu/execution.rs
```diff
@@ -211,107 +121,64 @@ impl<'a> Execution<'a> {
...
     /// Run on a real QPU and wait for the results.
-    pub(crate) async fn submit(
-        &mut self,
-        params: &Parameters,
-        readouts: &[&str],
-        config: &Configuration,
-    ) -> Result<JobId, Error> {
-        let qcs = self.refresh_qcs(readouts, config).await?;
-        let qcs_for_thread = qcs.clone();
+    /// needs to also return readout_sources
```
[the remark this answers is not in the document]
notmgsk: removed. out of date

PR #172, opened 2022-10-06
feat: upgrade quil-rs dependency and fix issues

notmgsk:
Does what it says on the tin. `quil-rs` has had some breaking changes which are fixed herein. Mostly this required updating/improving errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,9 +116,9 @@ dependencies = [
 
 [[package]]
 name = "digest"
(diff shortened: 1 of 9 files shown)


Shadow53:
[This](https://github.com/rigetti/quil-rs/pull/108) should fix the issue with the error not being `Send`

Shadow53 (approved):
LGTM

PR #180, opened 2022-10-11
fix: deserialize null fidelity and duration as "perfect" values

### LINKED ISSUE #179: Deserialize `null` fidelity and duration as "perfect" values
When pyQuil provides a `TargetDevice` that is based on a physical device (e.g. `Aspen-11`) it will have numeric fidelity and duration values. If instead pyQuil provides a `TargetDevice` that is a non-physical device (e.g. `8q-qvm`) it will have `null` fidelity and duration values. This is generally handled by patching `null` with "perfect" values.

notmgsk:
See attached issue for details.

Closes #179.

### DIFF
--- a/crates/lib/src/qpu/quilc/isa/operator.rs
+++ b/crates/lib/src/qpu/quilc/isa/operator.rs
@@ -31,9 +31,16 @@ pub(crate) enum Operator {
         duration: f64,
         #[serde(deserialize_with = "deserialize_fidelity_null_default")]
         fidelity: f64,
-        qubit: i32,
+        qubit: Qubit,
         target: Option<String>,
     },
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(untagged)]
+pub(crate) enum Qubit {
+    Int(i32),
+    String(String),
 }
 
 /// Swap a ``null`` duration value with [`PERFECT_DURATION`]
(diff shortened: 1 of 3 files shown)


MarquessV (approved):

crates/lib/src/qpu/quilc/isa/operator.rs
```diff
@@ -18,23 +18,47 @@ pub(crate) fn wildcard(node_id: Option<i32>) -> Operator {
 pub(crate) enum Operator {
     Gate {
         operator: String,
+        #[serde(deserialize_with = "deserialize_duration_null_default")]
```
Shadow53, line 21: Can you not just do this and leave out the manual deserializer stuff?
```suggestion
        #[serde(default = "PERFECT_DURATION")]
```

Shadow53 (commented):
Seems like this could be simplified...

crates/lib/src/qpu/quilc/isa/operator.rs
notmgsk: `default` expects a function, so it's not quite _that_ simple, but it can be improved with e.g.
```rust
fn deserialize_duration_null_default() -> f64 {
    PERFECT_DURATION
}
```

Shadow53: I saw you added `simplify default` and `Revert "simplify default"` commits, so I tried using a default function locally. Looks like an explicit `null` is considered invalid for a `f64` field. So it looks like the deserializer is needed after all :/

PR #198, opened 2022-10-24
fix(python): remove Awaitable from non-async function return types

MarquessV:
Making the type hint on these functions more accurate - they are sync functions, so they don't return Awaitables.


jselig-rigetti (approved):
Looks right 👍

PR #199, opened 2022-10-24
fix(python): remove awaitable annotation on async functions

jselig-rigetti:
See #198

Missed removing the type annotations


MarquessV:
Nice catch! When you squash this update this commit message to use the python scope, that way it only bumps and releases the python library:

```
fix(python): remove awaitable annotation on async functions
```

MarquessV (approved):

PR #204, opened 2022-10-27
feat: support protoquil flag in compilation

notmgsk:
This change makes it possible to configure the compiler such that it produces "protoquil" as its output.

BREAKING: Adds arguments to some functions.


jselig-rigetti:
Apologies for breaking this MR... I'll close this in lieu of https://github.com/rigetti/qcs-sdk-rust/pull/243 instead.

PR #206, opened 2022-11-01
feat: add method for getting the version of the running quilc server

### LINKED ISSUE #205: Python: Add method for getting version info from quilc

MarquessV:
Closes #205

### DIFF
--- a/crates/lib/src/api.rs
+++ b/crates/lib/src/api.rs
@@ -24,7 +24,7 @@ use crate::qpu::{
 };
 
 /// Uses quilc to convert a Quil program to native Quil
-/// TODO: Add `+ Send + Sync` to the error type once quil-rs supports it: https://github.com/rigetti/quil-rs/issues/122
+/// TODO: Add `+ Send + Sync` to the error type once quil-rs supports it: <https://github.com/rigetti/quil-rs/issues/122>
 pub fn compile(
     quil: &str,
     target: TargetDevice,
@@ -36,7 +36,7 @@ pub fn compile(
 }
 
 /// Gets the quilc version
-/// TODO: Add `+ Send + Sync` to the error type once quil-rs supports it: https://github.com/rigetti/quil-rs/issues/122
+/// TODO: Add `+ Send + Sync` to the error type once quil-rs supports it: <https://github.com/rigetti/quil-rs/issues/122>
 pub fn get_quilc_version(client: &Qcs) -> Result<String, Box<dyn std::error::Error + 'static>> {
     quilc::get_version_info(client).map_err(Into::into)
 }
--- a/crates/lib/src/api.rs
+++ b/crates/lib/src/api.rs
@@ -34,6 +34,11 @@ pub fn compile(
         .map(|p| p.to_string(true))
 }
 
+/// Gets the quilc version
+pub fn get_quilc_version(client: &Qcs) -> Result<String, Box<dyn std::error::Error>> {
+    quilc::get_version_info(client).map_err(Into::into)
+}
+
 /// Collection of errors that can result from rewriting arithmetic.
 #[derive(thiserror::Error, Debug)]
 pub enum RewriteArithmeticError {
--- a/crates/lib/src/api.rs
+++ b/crates/lib/src/api.rs
@@ -24,7 +24,9 @@ use crate::qpu::{
 };
 
 /// Uses quilc to convert a Quil program to native Quil
-/// TODO: Add `+ Send + Sync` to the error type once quil-rs supports it: <https://github.com/rigetti/quil-rs/issues/122>
+/// TODO: Add `+ Send + Sync` to the error type once quil-rs supports it:
+/// <https://github.com/rigetti/quil-rs/issues/122>
+/// <https://github.com/rigetti/qcs-sdk-rust/issues/210>
 pub fn compile(
     quil: &str,
     target: TargetDevice,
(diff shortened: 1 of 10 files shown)


crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -50,6 +50,19 @@ pub(crate) fn compile_program(
     }
 }
 
+pub(crate) fn get_version_info(client: &Qcs) -> Result<String, Error> {
+    let config = client.get_config();
+    let endpoint = config.quilc_url();
+    let binding: HashMap<&str, &str> = HashMap::new();
+    let request = rpcq::RPCRequest::new("get_version_info", &binding);
```
MarquessV, line 57: What is the best way to pass in an empty set of params to `rpcq::RPCRequest::new`? I don't love initializing this empty `HashMap` just for that purpose, but the function needs something, and the empty types I tried crash quilc. cc'ing: @Shadow53 for their Rust expertise.

Shadow53: `HashMap::new` should be fine. The standard library generally tries to avoid allocating until necessary, which [is the case](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.new) with `HashMap::new`.

Could also try passing in `()` if you haven't already. I think that implements `Serialize`. Only other option I can think to try is passing `None`, but I'm pretty sure the compiler would complain about not knowing what type the `Option` should contain.

crates/python/tests/test_api.py
```diff
@@ -1,11 +1,12 @@
 from unittest.mock import patch
 import pytest
 
+import re
+
 import qcs_sdk
 
 
 @pytest.mark.asyncio
-@pytest.mark.skip
```
Shadow53, line 8: This seems unrelated to getting quilc version. Was this change intended?

Shadow53 (changes requested):

crates/lib/src/qpu/quilc/mod.rs
MarquessV: I tried `()` first, which also unfortunately doesn't work. Rust doesn't mind, but `quilc` throws a pretty gnarly error, dies, and leaves the Rust library hanging.

```
Unhandled TYPE-ERROR in thread #<SB-THREAD:THREAD "RPC-server-thread-0" RUNNING
                                  {1005A834B3}>:
  The value
    NIL
  is not of type
    HASH-TABLE
```

crates/python/tests/test_api.py
MarquessV: Yes, when I added a test for the new quilc method, I also updated `conftest.py` so that these tests will load the config from `tests/qcs_config`. That way the test would work without having a QCS config in `~/.qcs` and will work in CI. That change also allowed this test to run successfully on CI so I removed the `skip` decoration. 

The other tests with `skip` would require a valid QCS config so we can't run those on CI just yet.

crates/lib/src/api.rs
```diff
@@ -34,6 +34,11 @@ pub fn compile(
         .map(|p| p.to_string(true))
 }
 
+/// Gets the quilc version
+pub fn get_quilc_version(client: &Qcs) -> Result<String, Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
MarquessV: The compiler doesn't like that:

image

crates/lib/src/qpu/quilc/mod.rs
Shadow53: Sounds good. Good to know that it requires a hash map.

It's strictly unrelated to this change, but it would be a good idea to update the documentation and/or signature for making a new `RPCRequest` to only allow hash maps, e.g. using (pseudo code) `binding: HashMap<K=String, V=String>` rather than `binding: T = HashMap<String, String>`

crates/lib/src/api.rs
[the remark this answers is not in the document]
Shadow53: If `Sync` is the only thing failing, that's fine to remove. I couldn't remember if it's something usually used here or not -- the more important ones are `'static` and `Send`.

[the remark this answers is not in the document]
MarquessV: It'll take `Error + 'static` but not ` + Send` or ` + Sync`

crates/lib/src/qpu/quilc/mod.rs
MarquessV: I like the idea of hardening the type that gets accepted, but I think only allowing HashMap would be too restrictive. For example, the compile function uses the `QuilcParams` struct. We only use a HashMap here as a quick way of creating an empty set of params.

crates/lib/src/api.rs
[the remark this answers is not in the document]
MarquessV: Michael and I discussed this offline - he found the issue within quil-rs:

```
1. Box has two From impls related to errors: From<E> for Box<dyn Error + 'a> and From<E> for Box<dyn Error + Send + Sync + 'a>. That is, there's no impl for Box<dyn Error + Send + 'a.
2. quilc::Error is not Sync because quil_rs::program::SyntaxError is not Sync.
3. SyntaxError is not Sync because quil_rs::parser::error::error::Error is not Sync
4. quil_rs::parser::error::error::Error is not Sync because it contains a field of type Option<Box<dyn Error + Send + 'static>>.
```

We have to add `Sync` to the parser error in quil-rs, then we can do it here. Adding a TODO since there is another method in this API with the same problem.

crates/lib/src/api.rs
```diff
@@ -24,16 +24,23 @@ use crate::qpu::{
 };
 
 /// Uses quilc to convert a Quil program to native Quil
+/// TODO: Add `+ Send + Sync` to the error type once quil-rs supports it: <https://github.com/rigetti/quil-rs/issues/122>
```
Shadow53, line 27: Should add a `TODO` for this repo as well.

crates/lib/src/api.rs
```diff
@@ -24,16 +24,23 @@ use crate::qpu::{
...
     client: &Qcs,
-) -> Result<String, Box<dyn std::error::Error>> {
+) -> Result<String, Box<dyn std::error::Error + 'static>> {
     quilc::compile_program(quil, target, client)
         .map_err(Into::into)
         .map(|p| p.to_string(true))
 }
 
+/// Gets the quilc version
+/// TODO: Add `+ Send + Sync` to the error type once quil-rs supports it: <https://github.com/rigetti/quil-rs/issues/122>
```
Shadow53, line 39: Same here

Shadow53 (changes requested):
Add links to an issue in this repository, then this looks good.

Shadow53 (approved):
lgtm

PR #211, opened 2022-11-03
feat!: make compiler timeout configurable

### LINKED ISSUE #209: The compiler timeout should be configurable
When `compile_program` is called, it builds it's own RPCRequest [here](https://github.com/rigetti/qcs-sdk-rust/blob/128922b0792be48d9b07a6e5b467e5aa8edb56ff/crates/lib/src/qpu/quilc/mod.rs#L42), which has a default timeout of 10.0s.

This timeout isn't sufficient for all programs, and should be configurable by the user. This configuration should be available for users that hit `compile_program` directly _and_ for users who create a new `Executable` (ie. with `Executable::from_quil`) and then use the `submit_to_qpu` method.

MarquessV:
Closes #209 

* `quilc::compile_program` now takes a `timeout: Option<u8>` parameter for overriding the default timeout.
* `Execution::new` now takes a `compiler_timeout: Option<u8>` parameter, since it calls out to `compile_program`.

### DIFF
--- a/crates/lib/src/qpu/quilc/mod.rs
+++ b/crates/lib/src/qpu/quilc/mod.rs
@@ -21,6 +21,7 @@ mod isa;
 /// * `program`: The Quil program to compile.
 /// * `isa`: The [`InstructionSetArchitecture`] of the targeted platform. Get this using
 ///     [`super::get_isa`].
+/// * `timeout`: The number of seconds to wait before timing out. If not set, defaults to 10.
 ///
 /// returns: `eyre::Result<quil_rs::Program>`
 ///
--- a/crates/lib/src/qpu/quilc/mod.rs
+++ b/crates/lib/src/qpu/quilc/mod.rs
@@ -34,11 +34,15 @@ pub(crate) fn compile_program(
     quil: &str,
     isa: TargetDevice,
     client: &Qcs,
+    timeout: Option<u8>,
 ) -> Result<quil_rs::Program, Error> {
     let config = client.get_config();
     let endpoint = config.quilc_url();
     let params = QuilcParams::new(quil, isa);
-    let request = rpcq::RPCRequest::new("quil_to_native_quil", &params);
+    let mut request = rpcq::RPCRequest::new("quil_to_native_quil", &params);
+    if let Some(seconds) = timeout {
+        request.with_timeout(seconds);
+    }
     let rpcq_client = rpcq::Client::new(endpoint)
         .map_err(|source| Error::from_quilc_error(endpoint.into(), source))?;
     match rpcq_client.run_request::<_, QuilcCompileProgramResponse>(&request) {
@@ -175,6 +179,7 @@ mod tests {
             "MEASURE 0",
             TargetDevice::try_from(qvm_isa()).expect("Couldn't build target device from ISA"),
             &Qcs::load().await.unwrap_or_default(),
+            None,
         )
         .expect("Could not compile");
         assert_eq!(output.to_string(true), EXPECTED_H0_OUTPUT);
@@ -196,6 +201,7 @@ MEASURE 1 ro[1]
             BELL_STATE,
             TargetDevice::try_from(aspen_9_isa()).expect("Couldn't build target device from ISA"),
             &client,
+            None,
         )
         .expect("Could not compile");
         let mut results = crate::qvm::Execution::new(&output.to_string(true))
--- a/crates/lib/src/qpu/quilc/mod.rs
+++ b/crates/lib/src/qpu/quilc/mod.rs
@@ -14,6 +14,9 @@ use super::{rpcq, Qcs};
 
 mod isa;
 
+/// Number of seconds to wait before timing out.
+pub const DEFAULT_COMPILER_TIMEOUT: u8 = 30;
+
 /// Take in a Quil program and produce a "native quil" output from quilc
 ///
 /// # Arguments
@@ -21,7 +24,7 @@ mod isa;
 /// * `program`: The Quil program to compile.
 /// * `isa`: The [`InstructionSetArchitecture`] of the targeted platform. Get this using
 ///     [`super::get_isa`].
-/// * `timeout`: The number of seconds to wait before timing out. If not set, defaults to 10.
+/// * `timeout`: The number of seconds to wait before timing out. If not set, defaults to 30.
 ///
 /// returns: `eyre::Result<quil_rs::Program>`
 ///
@@ -81,7 +84,9 @@ impl Default for CompilerOpts {
     /// Default compiler options
     /// * `timeout`: 30 seconds
     fn default() -> Self {
-        Self { timeout: Some(30) }
+        Self {
+            timeout: Some(DEFAULT_COMPILER_TIMEOUT),
+        }
     }
 }
 
--- a/crates/lib/src/qpu/quilc/mod.rs
+++ b/crates/lib/src/qpu/quilc/mod.rs
@@ -69,7 +69,7 @@ impl CompilerOpts {
         Self { timeout: None }
     }
 
-    /// Set the number of seconds to wait before timing out. If set to None, there is no timeout.
+    /// Set the number of seconds to wait before timing out. If set to None, the timeout is disabled.
     #[must_use]
     pub fn with_timeout(&mut self, seconds: Option<u8>) -> Self {
         self.timeout = seconds;
--- a/crates/lib/src/qpu/quilc/mod.rs
+++ b/crates/lib/src/qpu/quilc/mod.rs
@@ -35,7 +35,7 @@ pub(crate) fn compile_program(
     quil: &str,
     isa: TargetDevice,
     client: &Qcs,
-    options: &CompilerOpts,
+    options: CompilerOpts,
 ) -> Result<quil_rs::Program, Error> {
     let config = client.get_config();
     let endpoint = config.quilc_url();
@@ -64,11 +64,13 @@ pub struct CompilerOpts {
 impl CompilerOpts {
     /// Creates a new instance of [`CompilerOpts`] with zero values for each option.
     /// Consider using [`CompilerOpts::default()`] to create an instance with recommended defaults.
+    #[must_use]
     pub fn new() -> Self {
         Self { timeout: None }
     }
 
     /// Set the number of seconds to wait before timing out. If set to None, there is no timeout.
+    #[must_use]
     pub fn with_timeout(&mut self, seconds: Option<u8>) -> Self {
         self.timeout = seconds;
         *self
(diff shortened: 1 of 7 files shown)


crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -34,11 +35,15 @@ pub(crate) fn compile_program(
     quil: &str,
     isa: TargetDevice,
     client: &Qcs,
+    timeout: Option<u8>,
```
MarquessV, line 38: Is it an abuse of `Option`s in Rust to use them for pseudo-optional parameters like this? I don't love it. In this case, it would be easy to think setting `None` here would set no timeout at all, but it doesn't.

At first, I opted for a `compile_with_timeout` method but then the plain `compile` method would be misleading in the same way this parameter is. In addition the plain `compile` method would fall out of use, as we want to support configuring timeouts everywhere compilation is used. 

We _could_ get rid of the default timeout of 10s, but as a user, I'd rather run into a timeout that would give me a warning rather then have the program hang until I quit it without any obvious insight into why.

jselig-rigetti: Besides the other breaking changes such as using a builder pattern or a custom parameter struct, the only other option would be an environment variable and that seems pretty inflexible, so your change seems like the best move, especially since the rust comments and `.pyi` file make it clear what the default value is.

Does setting a value of `Some(0)` make the timeout indefinite?

---

Side thought - I'm guessing the following would be overkill if not an antipattern (but a nifty thing I found [here](https://vidify.org/blog/rust-parameters/#b-with-intooptiont))

```rust
pub(crate) fn compile_program(
    // ...
    timeout: impl Into<Option<u8>>,
) -> Result<quil_rs::Program, Error> {
    // ...
    if let Some(seconds) = timeout.into() {
        request.with_timeout(seconds);
    }
    // ...
}

// allows all of these:
compile_program(/* ... */, 10);
compile_program(/* ... */, Some(10));
compile_program(/* ... */, None);
```

notmgsk: `None` _should_, in effect, be no timeout... at least that's how quilc treats it. (I think it can also technically take a `false` value which indicates the request should use the _server's_ timeout setting.)

So I think the type of `client_timeout` _should_ be `Option<u8>` everywhere.

notmgsk: And I'm also in favour of bumping the default timeouts to 30s here and in pyquil.

MarquessV: @jselig-rigetti - That is a nifty trick. I do like that it allows passing in an int directly. I feel similarly though, not sure if it's an antipattern or not.

@notmgsk - 30s is what I was thinking for the Azure integration too, so we're aligned on setting a new default. We can make it Option<u8> all the way down. That said, if `None` is supposed to mean no timeout, how should we approach having a default of 30s? When calling `compile_program` should a `timeout: Option<u8>` of `None` mean default to 30s, or no timeout?

The latter _feels_ right, but it doesn't allow us to have a "default" timeout for the function.

MarquessV: After thinking about this a little more, I think I'm just gonna wrap this in a `CompilerOptions` struct that implements `default`. It's extendable, protects us from future breaking changes, and allows us to have our 30s default without being potentially misleading to the user.

notmgsk: Good idea!

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -50,6 +53,38 @@ pub(crate) fn compile_program(
...
+/// Functions for building a [`CompilerOpts`] instance
+impl CompilerOpts {
+    /// Creates a new instance of [`CompilerOpts`] with zero values for each option.
+    /// Consider using [`CompilerOpts::default()`] to create an instance with recommended defaults.
+    #[must_use]
+    pub fn new() -> Self {
+        Self { timeout: None }
+    }
+
+    /// Set the number of seconds to wait before timing out. If set to None, there is no timeout.
```
Shadow53, line 72:
```suggestion
    /// Set the number of seconds to wait before timing out. If set to None, the timeout is disabled.
```

crates/lib/src/qpu/rpcq.rs
```diff
@@ -176,10 +176,22 @@ impl<'params, T: Serialize> RPCRequest<'params, T> {
             params,
             id: Uuid::new_v4().to_string(),
             jsonrpc: "2.0",
-            client_timeout: 10,
+            client_timeout: Some(30),
```
Shadow53, line 179: I've seen this number used twice in code and once (with wrong value) in docs. Should make this a constant, maybe even public so the default value is visible in the public API.

crates/python/src/lib.rs
```diff
@@ -17,15 +18,32 @@ create_exception!(qcs, CompilationError, PyRuntimeError);
...
+fn compile<'a>(
+    py: Python<'a>,
+    quil: String,
+    target_device: String,
+    kwds: Option<&PyDict>,
+) -> PyResult<&'a PyAny> {
     let target_device: TargetDevice =
         serde_json::from_str(&target_device).map_err(|e| DeviceIsaError::new_err(e.to_string()))?;
+
+    let mut compiler_timeout = Some(30);
```
Shadow53, line 31: Constant would be used here too.

Shadow53 (changes requested):
Looks good overall, but I strongly suggest making the default timeout into a public constant (so the python lib can use it and so it's visible in API docs) and referencing that everywhere an int literal is used.

crates/lib/src/qpu/rpcq.rs
MarquessV: Fair enough. I created a constant for it - though I created a separate one for this RPCQ client and the quilc module because I don't think it makes sense for them to be coupled.

crates/python/src/lib.rs
MarquessV: Done

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -50,6 +56,40 @@ pub(crate) fn compile_program(
...
+    #[must_use]
+    pub fn with_timeout(&mut self, seconds: Option<u8>) -> Self {
+        self.timeout = seconds;
+        *self
+    }
+}
+
+impl Default for CompilerOpts {
+    /// Default compiler options
+    /// * `timeout`: 30 seconds
```
Shadow53, line 85:
```suggestion
    /// * `timeout`: See [`DEFAULT_COMPILER_TIMEOUT`]
```

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -14,13 +14,17 @@ use super::{rpcq, Qcs};
 
 mod isa;
 
+/// Number of seconds to wait before timing out.
+pub const DEFAULT_COMPILER_TIMEOUT: u8 = 30;
```
Shadow53, line 18:
```suggestion
pub use super::rpcq::DEFAULT_CLIENT_TIMEOUT;
```'
Or
```suggestion
/// Number of seconds to wait before timing out.
pub const DEFAULT_COMPILER_TIMEOUT: u8 = super::rpcq::DEFAULT_CLIENT_TIMEOUT;
```

Unless there's a specific reason for having these set separately?

crates/python/src/lib.rs
```diff
@@ -17,15 +18,32 @@ create_exception!(qcs, CompilationError, PyRuntimeError);
...
+fn compile<'a>(
+    py: Python<'a>,
+    quil: String,
+    target_device: String,
+    kwds: Option<&PyDict>,
+) -> PyResult<&'a PyAny> {
     let target_device: TargetDevice =
         serde_json::from_str(&target_device).map_err(|e| DeviceIsaError::new_err(e.to_string()))?;
+
+    let mut compiler_timeout = Some(DEFAULT_COMPILER_TIMEOUT);
```
Shadow53, line 31: Depending on how you handle the duplication of `DEFAULT_COMPILER_TIMEOUT` and `DEFAULT_CLIENT_TIMEOUT`, you may want to change this.

Shadow53 (changes requested):
Noticed a duplication of `DEFAULT_CLIENT_TIMEOUT` and `DEFAULT_COMPILER_TIMEOUT`. Left a couple different suggestions, but basically there should be only one thing that's `= 30`;

crates/lib/src/qpu/quilc/mod.rs
MarquessV: I know they are the same value, but I see no reason to couple the RPCQ client with the compiler logic. Even if they are used exclusively together at this point.

Shadow53: So my understanding is that the compiler timeout is the timeout for the rpcq client when talking with quilc, which would make this one specific use of a client timeout.

If I understand you right, you're wanting to keep them separate because we might want a compiler timeout even if we move away from rpcq in this case, and so setting it to the same value as the rpcq timeout would no longer make sense?

In that case, it makes sense to keep them separate, even if the current use seems duplicative.

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -14,13 +14,17 @@ use super::{rpcq, Qcs};
...
+
 /// Take in a Quil program and produce a "native quil" output from quilc
 ///
 /// # Arguments
 ///
 /// * `program`: The Quil program to compile.
 /// * `isa`: The [`InstructionSetArchitecture`] of the targeted platform. Get this using
 ///     [`super::get_isa`].
+/// * `timeout`: The number of seconds to wait before timing out. If not set, defaults to 30.
```
Shadow53, line 27:
```suggestion
/// * `timeout`: The number of seconds to wait before timing out. If not set, defaults to [`DEFAULT_COMPILER_TIMEOUT`].
```

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -50,6 +56,40 @@ pub(crate) fn compile_program(
     }
 }
 
+/// A set of options that determine the behavior of compiling programs with quilc
+#[derive(Clone, Copy, Debug)]
+pub struct CompilerOpts {
+    /// The number of seconds to wait before timing out. If unset, there is no timeout
```
Shadow53, line 62:
```suggestion
    /// The number of seconds to wait before timing out. If `None`, there is no timeout.
```

Shadow53 (approved):
Left a couple suggestions for improving docs. Once those are made, you're good to merge.

crates/lib/src/qpu/quilc/mod.rs
MarquessV: Exactly, we're on the same page now.

Shadow53 (approved):
LGTM

PR #219, opened 2022-11-16
Make public types actually public, etc.

### LINKED ISSUE #217: Improve type discovery/importing
While trying to use this library in another project, I ran into issues with not being able to import `execution_data::Qpu` and not being able to view documentation for the type in `rustdoc`.

I'm not sure if there are other types that have a similar problem, but it's worth a look.

Shadow53:
Resolves #217.

Added a suite of lints to try to help catch this sort of thing in the future.


MarquessV (approved):
Looks good!

kalzoo (approved):
clean

PR #223, opened 2022-11-22
feat!: Use a common struct for both QPU and QVM execution results

### LINKED ISSUE #215: execution results from QVM and QPU should be exposed through a common struct
The changes made [here](https://github.com/rigetti/qcs-sdk-rust/pull/171/files#diff-2281d28e7ecc3103d60e15be625ee0ed848d19230272184746aa986d36b4ea43L10) broke `ExecutionData` into two structs: `Qvm` and `Qpu`. This is because the structure of data returned on a `QVM`s `registers` and a `QPU`s `readout_data` are different. However, from a users perspective, it should be easy to switch between running their programs on a `QVM` and a `QPU`, and having to account for two different return types adds a lot of friction to that. 

We should re-introduce something akin to the `ExecutionData` struct that is returned as the public facing type for both `QVM` and `QPU` execution results. `Qpu.readout_data` really is just an extension of `Qvm.registers`, so we should be able to rename the current `Qpu` struct, and do some data wrangling to make the `QVM` register data fit.

MarquessV:
The original version of this PR attempted to abstract away the differences between QVM memory and QPU readout data by fitting them both into a common representation, where each register name mapped to a corresponding rectangular 2d-matrix where-in each row corresponded to the final state of that register after that shot.

It turns out QPU readout data is more complicated in that it treats each memory reference as a stream. These streams aren't delimited per-shot, so if the program isn't always performing exactly one readout to each memory reference per shot, it isn't clear from the readout data alone what the last value in each shot is. This is often the case for programs using features like mid-circuit measurement and dynamic control flow where a measure to a memory reference could occur more than once, or be skipped conditionally.

After some discussion internally, it became clear that the rectangular matrix this PR was originally attempting to build was indeed the preferred form, as it fits the type of math that is generally done on it. However, we can't escape from the realities of dealing with QPU readout data. In light of this, we've chosen to keep the abstraction, but only allow it if we can build it using the underlying readout data without making potentially erroneous assumptions. This gives users what they want in many cases, but forces them to build exactly what they need if their program calls for it.

This PR now stores the original data returned from either the QPU or QVM in a `ResultData` enum. Users can choose to turn this into a `RegisterMap`, which will build the "ideal" mapping of register names to rectangular matrices. If the underlying QPU data would be jagged, an error is returned, signaling to the user that they should build their own based on the program that was run.

The matrices themselves are built on the `ndarray` crate. This is not only convenient for working with the data in Rust, but it can also be passed to and from Python as a `numpy` `ndarray` using the `numpy` crate.

Closes #215


crates/lib/src/execution_data.rs
```diff
@@ -1,58 +1,130 @@
...
     /// The time it took to execute the program on the QPU, not including any network or queueing
     /// time. If paying for on-demand execution, this is the amount you will be billed for.
     ///
     /// This will always be `None` for QVM execution.
     pub duration: Option<Duration>,
 }
 
+/// An enum representing every possible readout type
+#[derive(Clone, Copy, Debug, EnumAsInner, PartialEq)]
+pub enum ReadoutTypes {
```
MarquessV: Should we simplify this enum to something like `Integer(i32)`, `Real(f64)`, and `Complex(Complex64)`?

The benefits are a simpler API, especially if you are using this in Rust. The tradeoff is memory efficiency, but users that would be concerned with that kind of performance are likely using data from a real `QPU`, and that's already returned in the biggest possible types anyways.

crates/lib/src/execution_data.rs
```diff
@@ -1,58 +1,121 @@
...
     /// The time it took to execute the program on the QPU, not including any network or queueing
     /// time. If paying for on-demand execution, this is the amount you will be billed for.
     ///
     /// This will always be `None` for QVM execution.
     pub duration: Option<Duration>,
 }
 
+/// An enum representing every possible readout type
+#[derive(Clone, Copy, Debug, EnumAsInner, PartialEq)]
+pub enum ReadoutValue {
```
MarquessV, line 30: Should we simplify this to something like: `Integer(i32), Complex(Complex64), Real(f64)`? It would be a lot easier to deal with as a user, but less efficient.

crates/lib/Cargo.toml
```diff
@@ -19,6 +19,7 @@ futures = "0.3.24"
 indexmap = "1.9.1"
 lazy_static = "1.4.0"
 log = "0.4.17"
+ndarray = "0.15.6"
```
Shadow53, line 22: FWIW, I think pulling in `ndarray` is the right choice here, especially if pyQuil already integrates with `numpy`.

The only possible "better" solution I can think of would be to use const generics to return a stack-allocated 2d array. I don't know how well that could be implemented, but it would mean one less dependency for people who don't want to use it.

crates/lib/examples/parametric_compilation.rs
```diff
@@ -37,21 +36,18 @@ async fn main() {
...
-            parametric_measurements.append(&mut values)
+            .get_values_by_shot("ro", 0)
+            .expect("readout values should contain 'ro'");
+
+        for value in first_ro_values.iter() {
+            parametric_measurements.push(
+                value
+                    .expect("expect ro to have values")
+                    .into_i32()
+                    .expect(" values should be i32"),
```
Shadow53, line 49:
```suggestion
                    .expect("values should be i32"),
```

crates/lib/examples/parametric_compilation.rs
```diff
@@ -37,21 +36,18 @@ async fn main() {
...
-            .expect("'ro' should have readout values");
-        if let Values::IntegerValues(IntegerReadoutValues { mut values }) = first_ro_values {
-            parametric_measurements.append(&mut values)
+            .get_values_by_shot("ro", 0)
+            .expect("readout values should contain 'ro'");
+
+        for value in first_ro_values.iter() {
+            parametric_measurements.push(
+                value
+                    .expect("expect ro to have values")
```
Shadow53, line 47:
```suggestion
                    .expect("ro should have values")
```

crates/lib/examples/quil_t.rs
```diff
@@ -24,13 +23,13 @@ async fn main() {
...
+        .expect("Readout data should include 'ro'");
 
-    match result {
-        Values::IntegerValues(IntegerReadoutValues { values }) => assert!(!values.is_empty()),
-        _ => panic!("expected IntegerReadoutValues, got {:?}", result),
+    for value in result.iter() {
+        value
+            .expect("values should exist")
+            .into_i8()
+            .expect("should be i32");
```
Shadow53, line 33:
```suggestion
            .expect("should be i8");
```

crates/lib/src/executable.rs
```diff
@@ -236,8 +236,7 @@ impl<'executable> Executable<'executable, '_> {
     }
 }
 
-type ExecuteResultQVM = Result<execution_data::Qvm, Error>;
-type ExecuteResultQPU = Result<execution_data::Qpu, Error>;
+type ExecuteResult = Result<execution_data::ExecutionData, Error>;
```
Shadow53, line 239: Could we change this to `ExecutionResult`? `ExecuteResult` seems off to me somehow, like we are executing the result instead of it being the result of an execution.

crates/lib/src/execution_data.rs
Shadow53: I'll let someone with more QC domain knowledge chime in, but from a Rust perspective I don't see a problem with having only one integer type, etc. If you expect the data to be a smaller integer, it's not much extra work to convert it while you make sure it *is* an integer, for example.

crates/lib/src/execution_data.rs
```diff
@@ -1,58 +1,121 @@
...
+    /// doesn't exist
+    pub fn get_value(
+        &self,
+        register_name: &str,
+        index: usize,
+        shot_num: usize,
+    ) -> Option<ReadoutValue> {
+        let register = self.0.get(register_name);
+        if let Some(matrix) = register {
+            if index > matrix.nrows() || shot_num > matrix.ncols() {
```
Shadow53, line 64: Is there an off by one error here?
```suggestion
            if index >= matrix.nrows() || shot_num >= matrix.ncols() {
```

I think the Rusty thing to do is something more like

```rust
self.0.get(register_name)
    .and_then(|matrix| matrix.get((index, shot_num)))
    .copied()
```

No manual bounds checks, no accidental panics.

crates/lib/src/execution_data.rs
```diff
@@ -1,58 +1,121 @@
...
+    /// Returns a vector containing all the values in the given register at a particular memory
+    /// index across all shots.
+    pub fn get_values_by_memory_index(
+        &self,
+        register_name: &str,
+        index: usize,
+    ) -> Option<Vec<Option<ReadoutValue>>> {
+        let register = self.0.get(register_name);
+        if let Some(matrix) = register {
+            if index > matrix.nrows() {
```
Shadow53, line 82: I was going to make a similar suggestion to above, but I couldn't find a method that returns an `Option`. Maybe I missed something?

Either way, double check for off by one:
```suggestion
            if index >= matrix.nrows() {
```

crates/lib/src/execution_data.rs
```diff
@@ -1,58 +1,121 @@
...
-                if readout_values.len() <= index {
-                    readout_values.resize(index + 1, None);
-                }
-                readout_values[index] = Some(values.clone());
+        register_name: &str,
+        shot: usize,
+    ) -> Option<Vec<Option<ReadoutValue>>> {
+        let register = self.0.get(register_name);
+        if let Some(matrix) = register {
+            if shot > matrix.ncols() {
```
Shadow53, line 99: Same as rows: maybe I missed something, seems like you should be able to avoid bounds checking.

```suggestion
            if shot >= matrix.ncols() {
```

crates/lib/src/execution_data.rs
```diff
@@ -63,44 +126,95 @@ impl ReadoutMap {
...
-            .flatten()
-            .collect::<HashMap<_, _>>()
-            .into();
+        let mut result = ReadoutMap(HashMap::new());
+        for (readout_name, values) in readout_values {
+            readout_mappings
+                .iter()
+                .filter(|(_, program_alias)| *program_alias == readout_name)
+                .map(|(program_name, _)| program_name.as_ref())
+                .map(parse_readout_register)
```
Shadow53, line 135: Could combine these into a single `filter_map`. Something like

```rust
.filter_map(|(program_name, program_alias)| {
    (*program_alias == readout_name).then(|| parse_readout_register(&program_name))
})
```

Or maybe just the first two:

```rust
.filter_map(|(program_name, program_alias)| {
    (*program_alias == readout_name).then(|| program_name.as_ref())
})
```

Shadow53: Mentioning as something that I think a clippy lint would point out. Not sure if one method is better than another.

crates/lib/src/execution_data.rs
```diff
@@ -63,44 +126,95 @@ impl ReadoutMap {
...
+                        .0
+                        .entry(reference.name)
+                        .and_modify(|m| {
+                            if shape.0 > m.nrows() {
+                                *m = Array2::from_shape_fn(shape, |(r, c)| {
+                                    *m.get((r, c)).unwrap_or(&None)
+                                });
+                            }
+                        })
+                        .or_insert(Array2::from_elem(shape, None));
```
Shadow53, line 166:
```suggestion
                        .or_insert_with(|| Array2::from_elem(shape, None));
```

crates/lib/src/execution_data.rs
```diff
@@ -63,44 +126,95 @@ impl ReadoutMap {
...
+                            .collect(),
+                        None => Vec::new(),
+                    };
+                    // TODO handle possible truncation
+                    let shape = (reference.index as usize + 1, row.len());
+                    let matrix = result
+                        .0
+                        .entry(reference.name)
+                        .and_modify(|m| {
+                            if shape.0 > m.nrows() {
```
Shadow53, line 160: Possible off by one?

crates/lib/src/execution_data.rs
```diff
@@ -63,44 +126,95 @@ impl ReadoutMap {
...
+                                *m = Array2::from_shape_fn(shape, |(r, c)| {
+                                    *m.get((r, c)).unwrap_or(&None)
```
Shadow53, line 162: What is the logic here? Looks like you're converting an `Option<Option<T>>` to an `Option<T>`. Would `Option::flatten` make more sense here?

crates/lib/src/execution_data.rs
MarquessV: Thank you! I could feel that there was a more Rusty thing to do here and I knew you would call it out. Applied your suggestion, just had to throw a `?` at the end as well.

crates/lib/src/execution_data.rs
MarquessV: On an `ndarray` `get` works like it does on a vec, so we are able to get away without bounds checking because it'll just return None if the index is out of bounds. The `row` and `column` methods will panic if the index is out of range, so I think we have to check the bounds - but I could be missing something, I would love to avoid having to check it if possible.

crates/lib/src/execution_data.rs
MarquessV: Neat! Went with the second option, it just felt a little more clear to have `parse_readout_register` be a distinct step.

crates/lib/src/execution_data.rs
MarquessV: I think this one is okay, the purpose of `shape` is to describe the number of rows and columns the matrix should have, so on l160 we define the number of rows as `reference.index + 1`. If that number is equal to the existing matrices `nrows`, then there is already enough space to accommodate a row at `reference.index`

crates/lib/src/execution_data.rs
jselig-rigetti: Based on [readout.proto](https://github.com/rigetti/qcs-api-client-rust/blob/main/qcs-api-client-grpc/proto/controller/readout.proto) I don't even see a `Real` case

crates/lib/src/execution_data.rs
MarquessV: We visit the memory references in arbitrary order, so we initialize the matrix with dimensions using the first memory index and row length we see for a register. The logic here is to grow the matrix if we encounter a reference with a higher memory index than we've previously seen. This `from_shape_fn` should populate a new matrix with the new dimension using the values in the existing matrix, and pad in any new indices with `None`.

To answer your question, we can indeed use `flatten()` in this case, and looks cleaner than defaulting to `&None` - thanks for letting me know of its existence.

crates/lib/Cargo.toml
MarquessV: I appreciate that - I went through iterations of rolling our own data structure but nothing felt quite as right as just using `ndarray`. Especially since, as you say, `pyQuil` already integrates with `numpy`. It feels like a bonus to be able to leverage all the linalg capabilities of the crate as well.

crates/lib/src/execution_data.rs
MarquessV: Yeah, I'm probably just gonna do it unless I hear otherwise, it's a bit of a pain to have to worry about 3 integer types every time you want to `match` a readout value. Python uses 32 bit ints and 64 bit floats anyways.

MarquessV: > Based on [readout.proto](https://github.com/rigetti/qcs-api-client-rust/blob/main/qcs-api-client-grpc/proto/controller/readout.proto) I don't even see a `Real` case

That bit confused me too, but Mark eventually helped me make sense of it. Real QPUs only return integers or complex numbers because those are the two only types that make sense to describe a Qubit (either a Qubit was measured into `0` or `1`, or it's superposition can be described as a complex number).

The QVM complicates things, because it'll also return parameter values back to you, so if your Quil has `DECLARE theta REAL[1]`, you'll get a "readout" register like: `{"theta": 1.0}`. That's why we have to include a "Real" as a possible readout type.

crates/lib/src/executable.rs
MarquessV: Yeah, that is weird, changed it.

crates/lib/src/execution_data.rs
kalzoo: So the three flavors of data here are:

- Raw ADC values - i16
- Unclassified, integrated - Complex64 IIRC (this is a single value collected from a sequence of the raw values)
- Classified - i1 in practice today, but we should leave room for higher values (i.e. 2)

@jselig-rigetti is on the money here: those are the only values we need to support *today*, in the file he linked.

However, here's the catch - that's not _inherently_ the case, because these are built & returned from a post-processing pipeline running within the execution service. The values defined above are the types which are returned today. However, it's not unthinkable that we'd want to return different data types in the future. So, it seems reasonable to me - if it doesn't add tech debt here - to leave room for `f64`.

> that's already returned in the biggest possible types anyways.

True - over the wire, this is helped by protobuf's [varint](https://developers.google.com/protocol-buffers/docs/encoding), but once it's unpacked in Rust, it is expanded to a larger value, so we might as well just pick that same size and go with it.

The three types that @MarquessV proposes seem wise to me.

MarquessV: I simplified these to the types described above.

crates/lib/src/execution_data.rs
```diff
@@ -65,44 +120,112 @@ impl ReadoutMap {
...
+                            .values
+                            .clone()
+                            .into_iter()
+                            .map(|c| Complex64::new(c.real().into(), c.imaginary().into()))
+                            .map(|c| Some(ReadoutValue::Complex(c)))
+                            .collect(),
+                        None => Vec::new(),
+                    };
+                    // TODO handle possible truncation
+                    let shape = (reference.index as usize + 1, row.len());
```
MarquessV, line 150: @Shadow53 does it make sense for `quil-rs` to use `usize` for a memory index? Would keep us from having to bubble errors up here.

Shadow53: I'm not the most familiar with the non-parsing bits of `quil-rs`, buuuuut....

I did a search for uses of `MemoryReference::index` in `quil-rs` and [this is the only time](https://github.com/rigetti/quil-rs/blob/e6e42bf04f34babd154f23ae81c599868fc6dbe2/src/expression.rs#L365) I see the field being used rather than set:

```rust
let value = values.get(memory_reference.index as usize)?;
```

Yeah, I'd say it makes sense. I would guess that the reason it is currently a `u64` is because the parser uses `Token::Integer(u64)`. Not sure if it would need a new `Token` variant for this use case, but that can be left for discussion in its own issue.

Shadow53: cc @kalzoo and @notmgsk

Shadow53 (commented):
- One style suggestion
- One note about why `ndarray` doesn't have `get_row(index) -> Option`, etc. methods
- One response to the question of using `usize` vs `u64` for `MemoryReference::index` in `quil-rs`

crates/lib/src/execution_data.rs
```diff
@@ -1,60 +1,115 @@
...
-                readout_values[index] = Some(values.clone());
+        register_name: &str,
+        shot: usize,
+    ) -> Option<Vec<Option<ReadoutValue>>> {
+        let register = self.0.get(register_name);
+        if let Some(matrix) = register {
+            if shot >= matrix.ncols() {
+                return None;
             }
+            Some(matrix.column(shot).to_vec())
```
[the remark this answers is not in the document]
MarquessV: Huh, yeah, I would expect to at least have an alternative method that gave you that safety if you wanted it. Good to know the justification for the existing method though.

I'll look into adding a trait - I also got around to adding tests that would catch these kinds of errors, which hopefully sets a pattern of checking for off-by-one errors in these kinds of methods.

crates/lib/src/execution_data.rs
MarquessV: For now I recreated the `MemoryReference` struct that uses a `usize` for the index instead of `u64`. We were already doing the parsing of the reference as a string in the library, so it was simple enough to cast the index to a `usize` and let the existing error handling do the rest. 

`MemoryReference` parsing errors are ignored by the `from_mapping_and_values` method, so any `index` that exceeds `usize` just wouldn't be populated in the matrix. This seems like a nearly impossible edge case, how could one write and run a Quil program that exceeded their systems `usize`? Even on 32-bit systems, would anyone write a serious program with `2^32` indices on a register?

notmgsk: See this comment for a reason why we might want to avoid `usize`: https://github.com/rigetti/quil-rs/issues/8#issuecomment-885265144

crates/lib/Cargo.toml
```diff
@@ -48,3 +49,4 @@ tempfile = "3.3.0"
 tokio = { version = "1.21.2", features = ["macros", "rt-multi-thread"] }
 warp = "0.3.3"
 regex = "1.7.0"
+test-case = "2.2.2"
```
notmgsk, line 52: TIL

crates/lib/src/executable.rs
```diff
@@ -40,12 +40,13 @@ use quil_rs::Program;
 /// #[tokio::main]
 /// async fn main() {
 ///     let mut result = Executable::from_quil(PROGRAM).with_config(ClientConfiguration::default()).with_shots(4).execute_on_qvm().await.unwrap();
-///     // We know it's i8 because we declared the memory as `BIT` in Quil.
 ///     // "ro" is the only source read from by default if you don't specify a .read_from()
-///     let data = result.registers.remove("ro").expect("Did not receive ro data").into_i8().unwrap();
-///     // In this case, we ran the program for 4 shots, so we know the length is 4.
-///     assert_eq!(data.len(), 4);
-///     for shot in data {
+///     // We get the data as a shot wise matrix so we can iterate over rows containing all the values in
```
notmgsk, line 44: I think this is clearer as
> We get the data as a matrix `M` whose elements `M[shot][index]` are the readout values for memory offset `index` during shot `shot`.

crates/lib/src/execution_data.rs
```diff
@@ -1,60 +1,114 @@
...
+    /// index across all shots.
+    #[must_use]
+    pub fn get_values_by_memory_index(
+        &self,
+        register_name: &str,
+        index: usize,
+    ) -> Option<Vec<Option<ReadoutValue>>> {
+        let register = self.0.get(register_name);
+        if let Some(matrix) = register {
+            if index >= matrix.nrows() {
```
notmgsk, line 72: Why isn't this an error?

crates/lib/src/execution_data.rs
```diff
@@ -1,60 +1,114 @@
...
 
-        Ok((!readout_values.is_empty()).then_some(readout_values))
+    /// Returns the matrix as a 2-dimensional array where the row is the memory index and the
+    /// column is the shot number.
+    #[must_use]
+    pub fn get_index_wise_matrix(&self, register_name: &str) -> Option<&RegisterMatrix> {
+        self.0.get(register_name)
+    }
+
+    /// Returns the matrix as a 2-dimensional array where the row is the shot number, and the
```
notmgsk, line 106: What matrix?

notmgsk (commented):
Looking good. Need to spend some more time with it, when I'm not tired :)

crates/lib/src/execution_data.rs
Shadow53: That comment makes sense, but in practicality there is going to be *somewhere* that the value turns into a `usize`. In that case, one of two things must happen:

- Some arches have the value truncated (`as usize`)
- Some arches hit an error during conversion (`usize::try_from()`)

IMO, the most backwards-compatible solution without unexpected side effects (i.e. truncation) is to have a new `Token` type in `quil-rs` for indices specifically, and surface an error if the parsed index does not fit a `usize` for some reason.

This is something that should never happen in practice, but I'd prefer to surface an error if someone decides to try this anyways.

crates/lib/src/api.rs
```diff
@@ -292,7 +292,7 @@ impl From<ControllerJobExecutionResult> for ExecutionResults {
 ///
 /// # Errors
 ///
-/// May error if a [`gRPC`] client cannot be constructed, or a [`gRPC`]
+/// May error if a [`Qcs`] client cannot be constructed, or if the ``gRPC``
 /// call fails.
```
Shadow53, line 296: Shouldn't this be \`gRPC\`?

Alternatively, assuming that clippy flagged this in the first place, create a `clippy.toml` file with the [following contents](https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown):

```toml
doc-valid-idents = ["..", "gRPC"]
```

crates/lib/src/execution_data.rs
```diff
@@ -228,6 +225,11 @@ impl ReadoutMap {
     }
 }
 
+struct MemoryReference {
+    name: String,
+    index: usize,
+}
+
```
Shadow53, line 232: I'm wary of a temporary solution of creating a new `MemoryReference` type. I appreciate that this is module private, at least. If/when we find a better solution for the `u64`/`usize` debate, I'd like to see this removed in favor of the `quil-rs` one again.

Which is a long way to say... is there an issue for replacing this in the future?

crates/lib/src/execution_data.rs
```diff
@@ -343,4 +346,83 @@ mod describe_readout_map {
         ]);
         assert_eq!(ro, expected);
     }
+
+    #[test_case(0, 0 => Some(ReadoutValue::Integer(1)) ; "returns value when both indices are in bounds")]
+    #[test_case(1, 0 => None ; "returns None when row index is off by 1")]
+    #[test_case(0, 1 => None ; "returns None when column index is off by 1")]
```
Shadow53, line 352: Is this a new syntax that compares the return value for you? 🤯

crates/lib/src/qpu/client.rs
```diff
@@ -193,7 +193,7 @@ pub enum GrpcClientError {
     GrpcError(#[from] GrpcError),
 }
 
-/// Errors that may occur while trying to use a [`OpenAPI`] client
+/// Errors that may occur while trying to use an ``OpenAPI`` client
```
Shadow53, line 196: Same note as `gRPC` -- this could also be added to `clippy.toml`

crates/lib/src/executable.rs
MarquessV: Thanks for the suggestion, updated it.

crates/lib/src/execution_data.rs
MarquessV: Updated the docstring, and made it reference `RegisterMatrix`

crates/lib/src/execution_data.rs
MarquessV: I suppose it could be, and I'm open to that idea. I went with this method because it aligns with the default behavior of standard library containers like `Vec`. `Vec<T>.get(i)` returns `Option<T>` and returns `None` if `i` is out of bounds.

crates/lib/src/api.rs
MarquessV: I went with double backticks because that is how it was done in other places. It doesn't appear to make a difference in the rendered docs, so I updated them all to single backticks.

crates/lib/src/qpu/client.rs
MarquessV: Switched to single backticks for these.

crates/lib/src/execution_data.rs
MarquessV: Yeah, it's pretty amazing.

crates/lib/src/execution_data.rs
MarquessV: Yeah, I don't love it and would like to rely on `quil-rs` for this. That said I don't think it's _that_ bad. At no point does any of this code flow through `quil-rs`, we're really just parsing the strings we get back from controller service into a structure that works for building a `ReadmoutMap`. Would be nice to not have to duplicate the parsing logic, but there is a separation of concerns and no communication with `quil-rs` that might break on account of the structs drifting.

Regardless, I created #224 and added a comment.

crates/python/Cargo.toml
```diff
@@ -26,6 +26,7 @@ tokio = "1.21"
 qcs-api = "0.2.1"
 rigetti-pyo3 = { version = "0.1.0-rc.0", features = ["extension-module", "complex"] }
 serde_json = "1.0.86"
+numpy = "0.17"
```
MarquessV, line 29: note to self, `numpy` re-exports `ndarray` which is also needed by the `lib` crate. `numpy`, `pyo3`, and `pyo3-asyncio` all need to be version matched as well, so we should probably lift them into workspace dependencies with a comment about how to keep them in sync

crates/python/src/execution_data.rs
```diff
@@ -37,25 +78,77 @@ py_wrap_type! {
     PyReadoutMap(ReadoutMap) as "ReadoutMap";
 }
 
+// py_wrap_union_enum! {
```
MarquessV, line 81: Note to self, need to remove this or figure out how to get the macro to support this. `PyArray2` is not clone, hence the manual implementation below.

crates/python/Cargo.toml
MarquessV: done

crates/lib/src/executable.rs
```diff
@@ -303,8 +344,8 @@ impl Executable<'_, '_> {
         self.qvm = Some(qvm);
         result
             .map_err(Error::from)
-            .map(|registers| execution_data::Qvm {
-                registers,
+            .map(|registers| execution_data::ExecutionData {
+                readout_data: ReadoutData::Qvm(registers),
```
kalzoo, line 348: One thing I've learned from the past few weeks of discussions is that we should label readout data and memory accurately. This was a mistake I introduced to pyQuil v3 and so it's no surprise that it's in the Rust SDK here ... but the QVM does not have readout data. This is memory.

Where things get weird is that we _only_ really have readout data for QPU right now when it comes to reconstructing per-shot memory. So, for QPU data, there should be a method like `build_memory_matrix_from_readout_data` (or better name) which we'd deprecate when we have a better option.

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,345 @@
...
+use ndarray::prelude::*;
 
-use crate::RegisterData;
+use crate::qpu::readout_data::ReadoutValues;
+use crate::{qpu::readout_data::QpuReadout, qvm::QvmMemory, RegisterData};
 
-/// The result of executing an [`Executable`](crate::Executable) via
-/// [`Executable::execute_on_qvm`](crate::Executable::execute_on_qvm).
+/// Represents the two possible types of readout data returned from either the QVM or a real QPU.
+/// Each variant contains the original data returned from it's respective executor.
```
kalzoo, line 15:
```suggestion
/// Each variant contains the original data returned from its respective executor.
```

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,345 @@
...
+/// (ie. "ro\[0\]") is more like a stream than a value in memory. Every `MEASURE` to a memory
+/// reference emits a new value to said stream. This means that the number of values per memory
+/// reference can vary per shot. For this reason, it's not always clear what the final value in
+/// each shot was for a particular reference. When this is the case, `to_readout_map()` will return
+/// an error as it's impossible to build a correct [`RegisterMatrix`]  from the data without
+/// knowing the intent of the program that was run. Instead, it's recommended to build the
+/// [`RegisterMatrix`] you need from the inner [`QpuReadout`] data using the knowledge of your
+/// program to choose the correct readout values for each shot.
+#[derive(Debug, Clone, PartialEq, EnumAsInner)]
+pub enum ReadoutData {
```
kalzoo, line 46: Yep. This is what I had in mind from the previous comment - this might better be `ResultData`, which can wrap:

* QVM memory
* QPU memory _and_ readout data

and then, like you have it here, present an interface for accessing those things in a common way.

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,345 @@
...
+    MissingRow { register: String, index: usize },
+
+    /// The memory reference could not be parsed
+    #[error("{0}")]
+    MemoryReferenceParseError(MemoryReferenceParseError),
+}
+
+impl ReadoutData {
+    /// Convert [`ReadoutData`] from its inner representation as [`QvmMemory`] or
+    /// [`QpuReadout`] into a [`ReadoutMap`]. The [`RegisterMatrix`] for each register will be
```
kalzoo, line 108: Likewise, not readout

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,345 @@
...
-        }
+    /// Returns a [`ReadoutMap`] with the underlying [`RegisterMatrix`] data
+    #[must_use]
+    pub fn from_hashmap(map: HashMap<String, RegisterMatrix>) -> Self {
+        Self(map)
+    }
 
-        Ok((!readout_values.is_empty()).then_some(readout_values))
+    /// Returns a [`ReadoutMap`] built from [`QvmMemory`]
+    pub fn from_qvm_memory(memory: &QvmMemory) -> Result<Self, RegisterMatrixConversionError> {
```
kalzoo, line 144: does this need to be pub?

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,345 @@
...
-            .iter()
-            .flat_map(|(readout_name, values)| {
-                readout_mappings
+    /// Attempts to build a [`ReadoutMap`] from [`QpuReadout`].
+    ///
+    /// # Errors
+    ///
+    /// This fails if the underlying [`QpuReadout`] data is jagged. See [`ReadoutData`] for more
+    /// detailed explanations of why and when this occurs.
+    pub fn from_qpu_readout_data(
```
kalzoo, line 212: does this need to be pub?

kalzoo (changes requested):
Leaving a quick / partial review. Will give a more detailed look later into the result manipulation logic and documentation, but the top-level interface looks right (aside from naming)

crates/lib/examples/quil_t.rs
```diff
@@ -20,17 +19,19 @@ async fn main() {
...
+        .expect("ro should be a register of integer values")
+        .to_owned();
 
-    match result {
-        Values::IntegerValues(IntegerReadoutValues { values }) => assert!(!values.is_empty()),
-        _ => panic!("expected IntegerReadoutValues, got {:?}", result),
-    }
+    let serialized = serde_json::to_string(&result).unwrap();
+    println!("{serialized}");
+    dbg!(&result);
```
Shadow53, line 36: Not the most important, but I think this fits the intent more closely. I don't think line number and file name are important for this output.

```suggestion
    println!("{result:?}");
```

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,345 @@
...
+    }
 
-        Ok((!readout_values.is_empty()).then_some(readout_values))
+    /// Returns a [`ReadoutMap`] built from [`QvmMemory`]
+    pub fn from_qvm_memory(memory: &QvmMemory) -> Result<Self, RegisterMatrixConversionError> {
+        Ok(Self(
+            memory
+                .iter()
+                .map(|(name, register)| {
+                    let register_matrix = match register {
```
Shadow53, line 149: I really want to suggest a macro to deduplicate the code in each branch, but not sure how much that would help maintainability.

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,345 @@
...
+                            )?,
+                        ))
+                    })
+                    // Collect into a type that will sort them by memory reference, this allows us
+                    // to make sure indices are sequential.
+                    .collect::<Result<
+                        BTreeMap<MemoryReference, &ReadoutValues>,
+                        RegisterMatrixConversionError,
+                    >>()?
+                    .iter()
```
Shadow53, line 240: Also useful to avoid unnecessary cloning later

```suggestion
                    .into_iter()
```

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,345 @@
...
-        result
-    }
-}
+                    // Pair all the memory references with their readout values
+                    .map(|(memory_reference, alias)| {
+                        Ok((
+                            parse_readout_register(memory_reference).map_err(|e| {
+                                RegisterMatrixConversionError::MemoryReferenceParseError(e)
+                            })?,
+                            readout_data.readout_values.get(alias).ok_or(
```
Shadow53, line 226: Why not this, to avoid cloning later on?

```suggestion
                            readout_data.readout_values.remove(alias).ok_or(
```

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,345 @@
...
+                                        register: reference.name.clone(),
+                                    })
+                                }
+                            }
+
+                            let previous_reference = if reference.index == 0 {
+                                None
+                            } else {
+                                Some(reference)
+                            };
```
Shadow53, line 312: As an alternative:

```suggestion
                            let previous_reference = (reference.index != 0).then_some(reference);
```

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,345 @@
...
+                            // If we haven't started on a new register, make sure we aren't
+                            // skipping an index. For example, if we jumped from ro[5] to ro[3], or
+                            // ro[2] to theta[1], we skipped a column.
+                            if let Some(previous) = previous_reference {
+                                if previous.name != reference.name
+                                    || previous.index != reference.index + 1
+                                {
+                                    return Err(RegisterMatrixConversionError::MissingRow {
+                                        register: reference.name.clone(),
+                                        index: previous.index + 1,
```
Shadow53, line 260: The condition is `previous.index != reference.index + 1` but the returned index is `previous.index + 1`. Looks like a potential off by one error to me.

Also, would it be worth doing this check for all registers before trying to build the matrix? You use a `BTreeMap` to sort by key, but collecting into `Vec` would also allow you to sort, and then `Vec::windows(2)` would allow you to iterate over sequential pairs of registers and make this check before doing any of the below work, including any cloning it may do.

May be overthinking this, but the logic here is somewhat complex and I'm looking for ways to make it easier to follow. Maybe intermediate values would be enough to make it easier to follow.

crates/lib/src/register_data.rs
```diff
@@ -2,21 +2,26 @@ use enum_as_inner::EnumAsInner;
...
 ///
 /// This represents a single vector (or "register") of typed memory across some number of shots.
 /// The register corresponds to the usage of a `DECLARE` instruction in Quil, and the name of that
 /// register should be provided with [`Executable::read_from`](`crate::Executable::read_from`).
 ///
-/// There is a variant of this enum for each type of data that a register could hold.
-/// Any variant of an instance of `ExecutionResult` will contain a `Vec` with one entry for each shot,
-/// where each entry represents the entire register.
+/// There is a variant of this enum for each type of data that a register could hold. The register
+/// is represented as a 2-dimensional, array `M` where the value `M[shot_number][memory_index]`
```
Shadow53, line 12:
```suggestion
/// is represented as a 2-dimensional array `M` where the value `M[shot_number][memory_index]`
```

crates/python/src/execution_data.rs
Shadow53: I don't think `PyArray2` should need to be `Clone`. It's the wrapped types, like `RegisterMatrix` that need to be `Clone`.

Most likely, you should be using `Py<PyArray2<_>>`?

Shadow53 (changes requested):
There were a couple parts where I had trouble grokking what was happening, so I may not have reviewed those as well as I could have. Hopefully I can come back to it later (tomorrow) and give it another look.

crates/python/src/execution_data.rs
MarquessV: Yeah, the clone thing was a red herring. Revisited it, and the issue I haven't figured out how to work around is that `PyArray2` doesn't implement `PyTryFrom` for `Array2` and since neither of which are defined in this crate, I can't implement it myself.

Cargo.toml
```diff
@@ -1,3 +1,23 @@
...
+qcs-api-client-openapi = "0.4.0"
+quil-rs = "0.15"
+serde_json = "1.0.86"
+tokio = "1.24.2"
+
+# ndarray is used by the `qcs` crate, but it is also used in the `python` crate via a 
+# re-export through the numpy crate. They should be updated as a pair to keep both 
+# crates version of ndarray in sync.
+# Similarly, pyo3 packages (`numpy`, `pyo3*`) track versions together and need to be
+# updated together.
```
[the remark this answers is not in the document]
MarquessV: Right, will update the comment.

It feels a bit weird to re-export crates `rigetti-pyo3` doesn't use, especially since that crate is open-source and feels like a nice expansion to `pyo3` as is. I wouldn't want to dilute its usefulness by re-exporting a bunch of crates tied to just our use case. I don't love this dependency management either, but it seems to be the reality of working with the pyo3 ecosystem at the moment.

crates/lib/examples/quil_t.rs
MarquessV: Oops, yeah, will clean up. I was doing some testing with this example and didn't fully sweep behind me.

crates/lib/src/execution_data.rs
```diff
@@ -65,44 +119,115 @@ impl ReadoutMap {
...
 
         result
     }
-}
 
-impl From<HashMap<MemoryReference, ReadoutValues>> for ReadoutMap {
-    fn from(map: HashMap<MemoryReference, ReadoutValues>) -> Self {
-        Self(map)
+    /// Creates a new `ReadoutMap` from a mapping of register names (ie. "ro") to a 2-dimensional
+    /// vector containing rows of that registers memory values for each shot.
```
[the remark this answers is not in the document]
MarquessV: Looks like this one was based on an old commit.

crates/lib/src/execution_data.rs
MarquessV: Eh, probably not. I was thinking it was effectively made public by the `to_readout_map` method, but we can keep the public api leaner by making these private.

crates/lib/src/execution_data.rs
MarquessV: Right, fixed the condition by using `previous.index - 1` in both places instead.

crates/lib/src/execution_data.rs
MarquessV: Nice!

crates/lib/src/execution_data.rs
MarquessV: We'd have to make `QpuReadout` mutable and consume it to do so right? If so, I don't think we can, since we don't want to destroy any of the data in `QpuReadout` if the transformation could ultimately fail.

crates/lib/src/execution_data.rs
MarquessV: Yeah, the logic here is complex. I initially wanted to a check like you suggest with `windows` and `any`, but stopped short when I realized you couldn't use the former with iterators. 

I can try collecting it into an intermediate value, but wasn't sure if that was worth breaking the chain.

crates/lib/src/execution_data.rs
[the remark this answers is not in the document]
MarquessV: Good question, I'm not sure, but if we were keeping the methods public, I think that very well could've been the right thing to do. I just made them private though, so I think we want to avoid the trait implementation, since it would effectively make them public again.

crates/lib/src/executable.rs
MarquessV: Can you help clarify your idea for `build_memory_matrix_from_readout_data` with a function signature? Trying to understand what the ask is.

The to be renamed `to_readout_map` method effectively builds memory matrices from readout data if it is possible for every register. Would `build_memory_matrix_from_readout_data` do it for a single register?

crates/lib/src/execution_data.rs
MarquessV: > QPU memory _and_ readout data

Would you suggest renaming `QpuReadout` if it is supposed to represent memory as well?

crates/lib/src/execution_data.rs
MarquessV: Yeah, if we were doing this in more than one place, a could see an argument for a macro. Just using one for this though feels like it might hurt maintainability more than help.

crates/lib/examples/parametric_compilation.rs
```diff
@@ -30,32 +29,30 @@ async fn main() {
...
-            .expect("Aspen-11 should always report duration");
+            .expect("Aspen-M-3 should always report duration");
 
-        let ro_readout_data = data
-            .readout_data
-            .get_readout_values_for_field("ro")
+        let first_ro_values = data
+            .result_data
+            .to_register_map()
+            .expect("should be able to create a ReadoutMap")
```
kalzoo, line 42:
```suggestion
            .expect("should be able to create a RegisterMap")
```

crates/lib/examples/quil_t.rs
```diff
@@ -20,17 +19,17 @@ async fn main() {
...
         .compile_with_quilc(false)
-        .execute_on_qpu("Aspen-11")
+        .execute_on_qpu("Aspen-M-3")
         .await
         .expect("Program should execute successfully")
-        .readout_data
-        .get_readout_values("ro".to_string(), 0)
+        .result_data
+        .to_register_map()
+        .expect("should be able to convert execution data to ReadoutMap")
```
kalzoo, line 27: ditto

crates/lib/src/qpu/mod.rs
```diff
@@ -12,6 +12,7 @@ use qcs_api_client_openapi::{
 pub mod client;
 mod execution;
 pub mod quilc;
+pub mod readout_data;
```
kalzoo, line 15: needs rename too

crates/lib/src/qpu/readout_data.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Real numbered readout values
+    Real(Vec<f64>),
+    /// Complex readout values
+    Complex(Vec<Complex64>),
+}
+
+/// This struct encapsulates the readout data returned from the QPU after executing a job.
+#[derive(Debug, Clone, PartialEq)]
+pub struct QpuReadout {
+    /// Mappings of a memory region (ie. "ro\[0\]") to it's key name in `readout_values` (ie. "q0")
```
kalzoo, line 26:
```suggestion
    /// Mappings of a memory region (ie. "ro\[0\]") to its key name in `readout_values` (ie. "q0")
```

crates/lib/src/qpu/readout_data.rs
```diff
@@ -0,0 +1,83 @@
...
+    Integer(Vec<i64>),
+    /// Real numbered readout values
+    Real(Vec<f64>),
+    /// Complex readout values
+    Complex(Vec<Complex64>),
+}
+
+/// This struct encapsulates the readout data returned from the QPU after executing a job.
+#[derive(Debug, Clone, PartialEq)]
+pub struct QpuReadout {
```
kalzoo, line 25: When we start to handle memory values alongside readout, both coming back from QPU execution - which we could start to do today, and will do more in the future - would those live on this struct, or another one? If they'd be here, which makes sense to me, then this should have a name that expresses it's the resulting data of QPU execution and not just readout.

crates/lib/src/qpu/readout_data.rs
```diff
@@ -0,0 +1,83 @@
...
+                            }
+                            None => ReadoutValues::Integer(Vec::new()),
+                        },
+                    )
+                })
+                .collect(),
+        }
+    }
+
+    /// Returns the [`ReadoutValues`] for a [`MemoryReference`], or None if a mapping to the
```
kalzoo, line 72:
```suggestion
    /// Returns the [`ReadoutValues`] for a [`MemoryReference`], or `None` if a mapping to the
```

crates/lib/src/qvm/mod.rs
```diff
@@ -11,6 +11,10 @@ use crate::RegisterData;
 
 mod execution;
 
+/// A map of register names (ie. "ro") to a [`RegisterData`] containing their values.
+#[allow(clippy::module_name_repetitions)]
+pub type QvmMemory = HashMap<String, RegisterData>;
```
kalzoo, line 16: @randall-fulton could speak to this, but we ran into a problem a while back where the execution service was just returning data like this - a string-keyed map - and there was no room left to be able to add another top-level field.

Basically, I think that this should:

1. Have a name that accommodates data other than memory in the future (as much as it seems stable and settled now)
2. Be a struct that allows for additional fields, and so contains the `HashMap` as only one field within.

crates/lib/src/register_data.rs
```diff
@@ -2,21 +2,26 @@ use enum_as_inner::EnumAsInner;
...
-/// where each entry represents the entire register.
+/// There is a variant of this enum for each type of data that a register could hold. The register
+/// is represented as a 2-dimensional array `M` where the value `M[shot_number][memory_index]`
+/// represents the value at `memory_index` for `shot_number`.
 ///
 /// # Usage
 ///
-/// Typically you will already know what type of data the `ExecutionResult` _should_ have, so you can
-/// use the [`mod@enum_as_inner`] methods (e.g. [`ExecutionResult::into_i8`]) in order to
+/// Typically, you will be interacting with this data through the [`crate::ReadoutData`] of an
```
kalzoo, line 17: Before merge, please give the docs (and whole crate) a sweep for the old names

kalzoo (dismissed):
All the changes from the last review look good, just a little more to go.

It's fine to refer to "readout" and "memory" (when that's accurate to do so), we just don't want to bind ourselves in to thinking the QPU is only readout and QVM is only memory and that's how it'll always be. Let's leave ourselves as flexible as we can be, and let users call your new methods here to get at data in particular, forward-compatible ways.

crates/lib/src/qpu/readout_data.rs
MarquessV: Renamed this `QpuResultData` and updated the module to match.

crates/lib/src/qvm/mod.rs
MarquessV: Sure, I guess we shouldn't take the response shape for granted. I replaced it with a `QvmResultData`

crates/lib/src/execution_data.rs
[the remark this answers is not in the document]
Shadow53: FWIW, I think there's a decent split on this on what's more "rusty". Pretty sure I've seen libraries that do both a dedicated method and `TryFrom`, and one probably delegates to the other.

Cargo.toml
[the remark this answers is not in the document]
Shadow53: I can see an argument for including the re-exports under feature flags, so they aren't dependencies if they aren't being used. I see your point though.

crates/python/src/execution_data.rs
Shadow53: Might be worth investigating blanket implementations in `rigetti-pyo3` based on `pyo3::{FromPyObject, IntoPy}`.

crates/lib/src/executable.rs
```diff
@@ -225,21 +227,24 @@ impl<'executable> Executable<'executable, '_> {
...
+    ///         let theta_register = result
+    ///             .result_data
+    ///             .to_register_map()
     ///             .expect("should fit readout map")
     ///             .get_register_matrix("theta")
     ///             .expect("should have theta")
     ///             .as_real()
     ///             .expect("should be real valued register")
-    ///             .to_owned()
+    ///             .to_owned();
```
Shadow53, line 238: Consistency note: this has `to_owned()`, while the others use `clone()`.

crates/lib/src/execution_data.rs
```diff
@@ -232,27 +243,27 @@ impl ReadoutMap {
...
                         ),
                         |(mut readout_data, previous_reference), (reference, values)| {
                             // If we haven't started on a new register, make sure we aren't
                             // skipping an index. For example, if we jumped from ro[5] to ro[3], or
-                            // ro[2] to theta[1], we skipped a column.
+                            // ro[2] to theta[0], we skipped a column.
                             if let Some(previous) = previous_reference {
                                 if previous.name != reference.name
-                                    || previous.index != reference.index + 1
+                                    || previous.index - 1 != reference.index
```
Shadow53, line 262: Still unsure about the logic here. Shouldn't it be this?
```suggestion
                                    || previous.index + 1 != reference.index
```

i.e., we expect `previous.index == reference.index - 1` or `previous.index + 1 == reference.index`?

crates/lib/src/execution_data.rs
```diff
@@ -232,27 +243,27 @@ impl ReadoutMap {
...
                                 {
                                     return Err(RegisterMatrixConversionError::MissingRow {
-                                        register: reference.name.clone(),
-                                        index: previous.index + 1,
+                                        register: reference.name,
+                                        index: previous.index - 1,
```
Shadow53, line 266: Same here

crates/lib/src/qvm/mod.rs
```diff
@@ -11,9 +11,21 @@ use crate::RegisterData;
...
 
-/// A map of register names (ie. "ro") to a [`RegisterData`] containing their values.
+/// Encapsulates data returned after running a program on the QVM
 #[allow(clippy::module_name_repetitions)]
-pub type QvmMemory = HashMap<String, RegisterData>;
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct QvmResultData {
+    /// A map of register names (ie. "ro") to a [`RegisterData`] containing their values.
+    pub memory: HashMap<String, RegisterData>,
+}
```
Shadow53, line 20: Was there a specific reason to move from a type alias to a wrapping struct? Just curious.

Shadow53 (changes requested):
Left a few comments. I think the change to fix the off by one error may have just made the calculation off in the other direction...

crates/lib/src/execution_data.rs
MarquessV: We are iterating over the references in reverse, so if we evaluate some index `n`, then the correct index to see next would be `n - 1`.  Hence, the check that the current `reference.index` is equal to `previous.index - 1`. If not, then we return an error with `previous.index - 1`, since that is the index we expected to see.

crates/lib/src/qvm/mod.rs
MarquessV: Per [this recommendation](https://github.com/rigetti/qcs-sdk-rust/pull/223#discussion_r1092549376) to make sure we can accommodate any additional data the QVM may return in the future.

crates/lib/src/executable.rs
MarquessV: I went ahead and made them all use `to_owned()`

crates/lib/src/execution_data.rs
kalzoo: So the [Quil spec](https://quil-lang.github.io/) doesn't specify the bitness of integers. However, along the lines of that comment that @notmgsk linked - I don't think the concerns of using these integers should leak backwards into the parser. 

Imagine that someone _does_ write a program with such a large register and happens to parse and run it on a 32-bit machine. Should that fail in parsing, or in usage with our client library? Of the two, the latter seems like the better choice. It's also the easiest to implement - rather than truncating here, I'd `try_from` ... although that will have to make this function fallible.

crates/lib/src/execution_data.rs
MarquessV: I used itertools to break the sequence check into it's own step.

crates/lib/src/execution_data.rs
MarquessV: Cool, that aligns with what I settled on for the PR. The memory reference parser now bubbles up errors and uses `usize::from_str()` which blows up if it would overflow the usize.

jselig-rigetti (approved):
Some light nitpicking, but overall LGTM after the changes resulting from updating to the latest `main` - IIRC we were waiting on updating the `.pyi` files until taking on #237

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,343 @@
...
+                                    .map(|c| Complex64::new(c.re.into(), c.im.into()))
+                                    .collect(),
+                            )
+                            .map_err(|_| {
+                                RegisterMatrixConversionError::InvalidShape {
+                                    register: name.to_string(),
+                                }
+                            })?,
+                        ),
+                    };
```
jselig-rigetti, line 205: Potentially move `map_error` outside of the match statement?

```suggestion
                    let register_matrix = match register {
                        RegisterData::I8(data) => Array::from_shape_vec(
                            (data.len(), data.first().map_or(0, Vec::len)),
                            data.iter().flatten().copied().map(i64::from).collect(),
                        )
                        .map(RegisterMatrix::Integer),
                        RegisterData::I16(data) => Array::from_shape_vec(
                            (data.len(), data.first().map_or(0, Vec::len)),
                            data.iter().flatten().copied().map(i64::from).collect(),
                        )
                        .map(RegisterMatrix::Integer),
                        RegisterData::F64(data) => Array::from_shape_vec(
                            (data.len(), data.first().map_or(0, Vec::len)),
                            data.iter().flatten().copied().collect(),
                        )
                        .map(RegisterMatrix::Real),
                        RegisterData::Complex32(data) => Array::from_shape_vec(
                            (data.len(), data.first().map_or(0, Vec::len)),
                            data.iter()
                                .flatten()
                                .copied()
                                .map(|c| Complex64::new(c.re.into(), c.im.into()))
                                .collect(),
                        )
                        .map(RegisterMatrix::Complex),
                    }
                    .map_err(|_| {
                        RegisterMatrixConversionError::InvalidShape {
                            register: name.to_string(),
                        }
                    })?;
```

I wanted to do this with other common stuff too, but didn't see a clean way to do this
```rs
let arr2 = |data| {
    Array::from_shape_vec(
        (data.len(), data.first().map_or(0, Vec::len)),
        data
    )
};
```

crates/python/src/execution_data.rs
```diff
@@ -34,28 +71,72 @@ py_wrap_data_struct! {
...
+            Err(PyValueError::new_err("not a real numbered register"))
+        }
+    }
+
+    fn is_real(&self) -> bool {
+        matches!(self.as_inner(), RegisterMatrix::Real(_))
+    }
+
+    #[classmethod]
+    fn from_complex(_cls: &PyType, matrix: &PyArray2<Complex64>) -> PyRegisterMatrix {
```
jselig-rigetti, line 118: should this just be `staticmethod`?

crates/python/src/execution_data.rs
```diff
@@ -34,28 +71,72 @@ py_wrap_data_struct! {
...
+            Err(PyValueError::new_err("not a integer register"))
+        }
+    }
+
+    fn is_integer(&self) -> bool {
+        matches!(self.as_inner(), RegisterMatrix::Integer(_))
+    }
+
+    #[classmethod]
+    fn from_real(_cls: &PyType, matrix: &PyArray2<f64>) -> PyRegisterMatrix {
```
jselig-rigetti, line 101: should this just be `staticmethod`?

crates/python/src/execution_data.rs
```diff
@@ -34,28 +71,72 @@ py_wrap_data_struct! {
...
+}
+
+py_wrap_type! {
+    PyRegisterMatrix(RegisterMatrix) as "RegisterMatrix"
+}
+
+#[pymethods]
+impl PyRegisterMatrix {
+    #[classmethod]
+    fn from_integer(_cls: &PyType, matrix: &PyArray2<i64>) -> PyRegisterMatrix {
```
jselig-rigetti, line 84: should this just be `staticmethod`?

jselig-rigetti (commented):
Overall looks great, really like using a btreemap to sort the data.

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,343 @@
...
+            if reference_a.name == reference_b.name {
+                if reference_a.index + 1 != reference_b.index {
+                    return Err(RegisterMatrixConversionError::MissingRow {
+                        register: reference_a.name.clone(),
+                        index: reference_a.index + 1,
+                    });
                 }
-                readout_values[index] = Some(values.clone());
+            } else if reference_b.index != 0 {
+                return Err(RegisterMatrixConversionError::MissingRow {
```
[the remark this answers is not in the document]
MarquessV: Nice catch, that edge case wasn't properly accounted for when I switched to using `itertools`. I added a peek into the first value to ensure the first index was zero.

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,343 @@
...
-            .collect::<HashMap<_, _>>()
-            .into();
-
-        result
-    }
-}
+        Ok(Self(
+            // Iterate over them in reverse so we  can initialize each RegisterMatrix with the
+            // correct number of rows
+            register_map.into_iter().rev().try_fold(
```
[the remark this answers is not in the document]
MarquessV: Nice! I am continually impressed by the versatility of the iterator API

crates/lib/src/execution_data.rs
MarquessV: Good idea, applied your change.

crates/lib/src/qpu/result_data.rs
```diff
@@ -0,0 +1,84 @@
...
+use enum_as_inner::EnumAsInner;
+use num::complex::Complex64;
+use quil_rs::instruction::MemoryReference;
+use std::collections::HashMap;
+
+use qcs_api_client_grpc::models::controller::{
+    readout_values as controller_readout_values, ReadoutValues as ControllerReadoutValues,
+};
+
+/// A row of readout values from the QPU.
```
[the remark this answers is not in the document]
MarquessV: This would be a vector containing the values emitted for a particular memory reference (ie "ro[0]") across all shots. I added clarification to the docstring.

MarquessV:
Here is the squashed commit message I plan to use:

```
feat!: The `execution_data` module now provides `ExecutionData` as a replacement for both `Qvm` and `Qpu` structs. It serves a common interface for interacting with both result shapes when possible. See the `ExecutionData` documentation for more details on how to use it.
```

crates/python/qcs_sdk/_execution_data.pyi
```diff
@@ -6,67 +6,142 @@ It is only here to represent the structure of the rust source code 1:1
...
+    def is_integer(self) -> bool: ...
+    def is_real(self) -> bool: ...
+    def is_complex(self) -> bool: ...
+    def as_integer(self) -> Optional[NDArray[np.int64]]: ...
+    def as_real(self) -> Optional[NDArray[np.float64]]: ...
+    # In numpy `complex128` is a complex number made up of two `f64`s.
+    def as_complex(self) -> Optional[NDArray[np.complex128]]: ...
+    def to_integer(self) -> Optional[NDArray[np.int64]]: ...
+    def to_real(self) -> Optional[NDArray[np.float64]]: ...
+    def to_complex(self) -> Optional[NDArray[np.complex128]]: ...
```
jselig-rigetti, line 47: I think the `to_` variants are not `Optional` since they throw if that type isn't found, right?

```suggestion
    def to_integer(self) -> NDArray[np.int64]: ...
    def to_real(self) -> NDArray[np.float64]: ...
    def to_complex(self) -> NDArray[np.complex128]: ...
```

crates/python/qcs_sdk/_execution_data.pyi
```diff
@@ -6,67 +6,142 @@ It is only here to represent the structure of the rust source code 1:1
...
+    Instead, it's recommended to manually build a matrix from ``QPUResultData`` that accurately
+    selects the last value per-shot based on the program that was run.
+    """
+
+    def is_qvm(self) -> bool: ...
+    def is_qpu(self) -> bool: ...
+    def as_qvm(self) -> Optional[QVMResultData]: ...
+    def as_qpu(self) -> Optional[QPUResultData]: ...
+    def to_qvm(self) -> Optional[QVMResultData]: ...
+    def to_qpu(self) -> Optional[QPUResultData]: ...
```
jselig-rigetti, line 133: Similar to above comment, this throws if the inner type is not found right?
```suggestion
    def to_qvm(self) -> QVMResultData: ...
    def to_qpu(self) -> QPUResultData: ...
```

crates/python/qcs_sdk/qpu/result_data.pyi
```diff
@@ -0,0 +1,54 @@
...
+
+    def is_integer(self) -> bool: ...
+    def is_real(self) -> bool: ...
+    def is_complex(self) -> bool: ...
+    def as_integer(self) -> Optional[List[int]]: ...
+    def as_real(self) -> Optional[List[float]]: ...
+    def as_f64(self) -> Optional[List[complex]]: ...
+    def to_integer(self) -> Optional[List[int]]: ...
+    def to_real(self) -> Optional[List[float]]: ...
+    def to_complex(self) -> Optional[List[complex]]: ...
```
jselig-rigetti, line 30: Similar to above, `to_` will throw if the value isn't found right?
```suggestion
    def to_integer(self) -> List[int]: ...
    def to_real(self) -> List[float]: ...
    def to_complex(self) -> List[complex]: ...
```

jselig-rigetti (approved):
LGTM, just some `pyi` questions

crates/lib/src/executable.rs
```diff
@@ -41,12 +41,27 @@ use quil_rs::Program;
...
 ///     // "ro" is the only source read from by default if you don't specify a .read_from()
-///     let data = result.registers.remove("ro").expect("Did not receive ro data").into_i8().unwrap();
-///     // In this case, we ran the program for 4 shots, so we know the length is 4.
-///     assert_eq!(data.len(), 4);
-///     for shot in data {
+///
+///     // We first convert the readout data to a [`RegisterMap`] to get a mapping of registers
+///     // (ie. "ro") to a [`RegisterMatrix`], `M`, where M[`shot`][`index`] is the value for
+///     // the memory offset `index` during shot `shot`.
+///     // There are some programs where QPU readout data does not fit into a [`RegisterMap`]. In
```
kalzoo, line 49:
```suggestion
///     // There are some programs where QPU readout data does not fit into a [`RegisterMap`], in
```

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,334 @@
...
+/// register name is mapped to a 2-dimensional rectangular [`RegisterMatrix`] where each row
+/// represents the final values in each register index for a particular shot. This is often the
+/// desired form of the data and it is _probably_ what you want. This transformation isn't always
+/// possible, in which case `to_register_map()` will return an error.
+///
+/// To understand why this transformation can fail, we need to understand a bit about how readout data is
+/// returned from the QVM and from a real QPU:
+///
+/// The QVM treats each `DECLARE` statement as initialzing some amount of memory. This memory works
+/// as one might expect it to. It is zero-initalized, and subsequent writes to the same region
```
kalzoo, line 34:
```suggestion
/// as one might expect it to. It is zero-initialized on each shot, and subsequent writes to the same region
```

crates/lib/src/execution_data.rs
```diff
@@ -1,120 +1,334 @@
...
+pub enum MemoryReferenceParseError {
     #[error("Could not parse memory reference: {reason}")]
     InvalidFormat { reason: String },
 
     #[error("Could not parse index from reference name: {0}")]
     InvalidIndex(#[from] ParseIntError),
 }
 
-// Note: MemoryReference may have a from_string in the fututre
 fn parse_readout_register(
```
kalzoo, line 332: this _should_ be [this](https://github.com/rigetti/quil-rs/blob/ba5f28987a1c64efbde1c621ec219e7d85f7128c/src/instruction.rs#L340) - no custom function or error type needed.

crates/lib/src/qpu/result_data.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/// This struct encapsulates data returned from the QPU after executing a job.
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug, Clone, PartialEq)]
+pub struct QpuResultData {
+    /// Mappings of a memory region (ie. "ro\[0\]") to it's key name in `readout_values` (ie. "q0")
+    pub mappings: HashMap<String, String>,
+    /// Mapping of a readout values identifier (ie. "q0") to a set of [`ReadoutValues`]
+    pub readout_values: HashMap<String, ReadoutValues>,
```
kalzoo, line 31: I could be convinced otherwise - but I think that these (and the QVM equivalents) should remain private (i.e. `pub(crate)` and be accessed via methods. We're going to want to change these as the API evolves and our ideas of result data evolve, and if they're public - we're stuck.

I think the SDK should treat the data coming back from execution as a black box that can be queried for data in various shapes. That way, even if the shape in which they are stored changes - we can continue to uphold our API until we _choose_ to break it.

crates/lib/src/qvm/mod.rs
```diff
@@ -11,6 +11,22 @@ use crate::RegisterData;
 
 mod execution;
 
+/// Encapsulates data returned after running a program on the QVM
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct QvmResultData {
+    /// A map of register names (ie. "ro") to a [`RegisterData`] containing their values.
+    pub memory: HashMap<String, RegisterData>,
```
kalzoo, line 19: Likewise re: visibility

kalzoo (commented):
It's almost there! This is excellent work. 

Some small typos and a note re: visibility; otherwise, this is ready to go.

crates/lib/src/execution_data.rs
MarquessV: Fair, I'll commit this, though there is an [open QVM issue](https://github.com/quil-lang/qvm/issues/313) related to this.

Shadow53 (changes requested):
Left a handful of stylistic comments and a question or two.

Main reason for requesting changes is that I ran `cargo check --all-targets --all-features` locally and it surfaced some errors inside of tests. It may also be worth running `cargo clippy` with the same flags, just to be safe.

crates/lib/src/qpu/result_data.rs
MarquessV: I've implemented this, but a practical problem that arises is whether or not we want to expose constructors for these types. If we do, we are committing to some data shape for initializing the type. We can choose to omit constructors, but that makes the library hard to work with from the outside. Ie. you can't build your own `QpuResultData` for testing without actually submitting a job and retrieving the results.

crates/python/qcs_sdk/_execution_data.pyi
```diff
@@ -6,67 +6,142 @@ It is only here to represent the structure of the rust source code 1:1
...
-
-class QPU:
-    @property
-    def readout_data(self) -> ReadoutMap: ...
-    @readout_data.setter
-    def readout_data(self, value: ReadoutMap): ...
-
+    def result_data(self) -> ResultData: ...
+    @result_data.setter
+    def registers(self, result_data: ResultData): ...
```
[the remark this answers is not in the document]
MarquessV: Yup, fixed

crates/python/qcs_sdk/_execution_data.pyi
```diff
@@ -6,67 +6,142 @@ It is only here to represent the structure of the rust source code 1:1
 import datetime
 from typing import List, Optional
 
-class _IntegerReadoutValues:
+import numpy as np
+from numpy.typing import NDArray
+
+from crates.python.qcs_sdk.qpu.result_data import QPUResultData
+from crates.python.qcs_sdk.qvm.result_data import QVMResultData
```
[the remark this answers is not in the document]
MarquessV: I'm not sure, but they weren't intentional. I think these were auto imports from pyright. Fixed.

crates/python/src/execution_data.rs
```diff
@@ -34,28 +66,96 @@ py_wrap_data_struct! {
...
-    PyReadoutMap(ReadoutMap) as "ReadoutMap";
+    PyRegisterMap(RegisterMap) as "RegisterMap";
+}
+
+py_wrap_type! {
+    PyRegisterMatrix(RegisterMatrix) as "RegisterMatrix"
+}
+
+#[pymethods]
+impl PyRegisterMatrix {
```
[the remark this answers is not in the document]
MarquessV: `numpy::PyArray2` doesn't implement PyTryFrom for `ndarray:Array2` and it can't be defined because neither type exists in this crate.

crates/lib/src/qpu/result_data.rs
MarquessV: I've implemented constructors, but if we want to avoid that, then the python unit tests targeting these will have to become integration tests.

crates/lib/src/execution_data.rs
MarquessV: It appears the future is here.

I updated the function to use `from_str`, and the error type to match, but I ultimately kept them, because it's a useful abstraction for the readout data to register map conversion. We need the `index` of a memory reference to be `usize` since it is used to initialize matrices, and catching that here instead of every other place it's used.

crates/python/src/execution_data.rs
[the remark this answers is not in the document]
Shadow53: Right, this is something that would need to be done in `rigetti-pyo3` directly, if anywhere.

Shadow53 (commented):
Three things:

1. Enum variants got turned into tuples in a `pyi` file
2. `cargo deny check` fails, in part due to the addition of `numpy`
3. `cargo doc` and `cargo deadlinks --check-intra-doc-links` show warnings. Not sure how many were there previously, but worth at least looking at.

crates/python/qcs_sdk/_executable.pyi
```diff
@@ -64,46 +60,41 @@ class JobHandle:
...
     def value(self) -> float: ...
     @value.setter
     def value(self, value: float): ...
 
-
 class Service(Enum):
-    Quilc = "Quilc",
-    Qvm = "Qvm",
-    Qcs = "Qcs",
-    Qpu = "Qpu",
```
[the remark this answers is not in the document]
MarquessV: That isn't an error with black, rather it highlighted an error with how the enum was defined. Both forms of the enum are equivalent. The trailing comma after each value makes it a tuple with or without the parentheses. It's a silly Python thing, but black was just trying to standardize it. I fixed it by removing the extra comma.

MarquessV:
I added the numpy license to `deny.toml` to satisfy cargo deny. The doc warnings and deadlinks are all pre-existing. The latter looks like it has to do with code generated by the pyo3 bindings.

kalzoo (approved):

crates/lib/src/execution_data.rs
kalzoo: Would it still make sense to use the official parser and then convert from one MemoryReference to the other? Unlikely to make a difference in practice, but that saves us two possibly-differing parser implementations. One difference even in this simple parser is that `ro` is a valid memory reference but, lacking brackets, would not parse correctly here.

Shadow53 (approved):
LGTM

PR #225, opened 2022-11-30
fix: Fetch Gateway address with pagination

randall-fulton:
Remove `GATEWAY_ADDR` environment variable and fully paginate accessor results to find the highest priority Gateway.

### DIFF
--- a/crates/lib/src/qpu/client.rs
+++ b/crates/lib/src/qpu/client.rs
@@ -136,6 +136,7 @@ impl Qcs {
             .await?;
             gateways.extend(accessors.accessors.into_iter().filter(|acc| {
                 acc.live
+                    // `as_deref` needed to work around the `Option<Box<_>>` type.
                     && acc.access_type.as_deref() == Some(&QuantumProcessorAccessorType::GatewayV1)
             }));
             next_page_token = accessors.next_page_token.clone();
@@ -144,7 +145,9 @@ impl Qcs {
             }
         }
         gateways.sort_by_key(|acc| acc.rank);
-        let target = gateways.first().ok_or_else(|| GrpcEndpointError::NoEndpoint(quantum_processor_id.to_string()))?;
+        let target = gateways
+            .first()
+            .ok_or_else(|| GrpcEndpointError::NoEndpoint(quantum_processor_id.to_string()))?;
         parse_uri(&target.url).map_err(GrpcEndpointError::BadUri)
     }
 }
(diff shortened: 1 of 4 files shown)


crates/lib/src/qpu/client.rs
```diff
@@ -124,32 +124,28 @@ impl Qcs {
...
-                        && acc.access_type.as_ref().unwrap().as_ref()
-                            == &QuantumProcessorAccessorType::GatewayV1
-                })
-                .ok_or_else(|| GrpcEndpointError::NoEndpoint(quantum_processor_id.into()))?;
-            target.url
-        };
-        parse_uri(&gateway_addr).map_err(GrpcEndpointError::BadUri)
+            gateways.extend(accessors.accessors.into_iter().filter(|acc| {
+                acc.live
+                    && acc.access_type.as_deref() == Some(&QuantumProcessorAccessorType::GatewayV1)
```
Shadow53, line 139: ~~You probably don't need the `as_deref` here.~~

Cloned locally and figured out why this is the way it is. I'd suggest leaving a note for the next person to look at this code.

```suggestion
                    // `as_deref` to work around the `Option<Box<_>>` type.
                    && acc.access_type.as_deref() == Some(&QuantumProcessorAccessorType::GatewayV1)
```

Shadow53 (approved):
The logic here seems fine, barring my note about how the API should filter/sort for us.

One style thing confused me, so I suggested a comment to help explain it.

crates/lib/src/qpu/client.rs
```diff
@@ -124,32 +124,28 @@ impl Qcs {
...
+            }));
```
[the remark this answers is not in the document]
randall-fulton: Definitely something that we can improve in the future. In actuality, this will only ever pull a single page right now, so the impact is minimal.

crates/lib/src/qpu/client.rs
```diff
@@ -124,32 +124,31 @@ impl Qcs {
...
-                        && acc.access_type.as_ref().unwrap().as_ref()
-                            == &QuantumProcessorAccessorType::GatewayV1
-                })
-                .ok_or_else(|| GrpcEndpointError::NoEndpoint(quantum_processor_id.into()))?;
-            target.url
-        };
-        parse_uri(&gateway_addr).map_err(GrpcEndpointError::BadUri)
+            gateways.extend(accessors.accessors.into_iter().filter(|acc| {
+                acc.live
+                    // `as_deref` needed to work around the `Option<Box<_>>` type.
```
notmgsk, line 139: Comment not strictly necessary

notmgsk (approved):

PR #230, opened 2023-01-09
feat!: implement Python wrappers for the full Rust API

### LINKED ISSUE #208: The Python bindings should return opaque pointers instead of dictionaries
Currently, any object we pass through to Python are serialized and returned as a dictionary. This isn't performant nor ergonomic. We'll generally want to wrap the SDKs internal types as opaque pointers. We should be mindful of doing this in a way that could be extended to other language bindings too.

Shadow53:
Resolves #208.

This does not include tests for the new Python stuff, but no clippy lints are triggered. There are changes made to the Rust crate in order to make this work, but everything should just work as before.


crates/lib/src/qvm/execution.rs
```diff
@@ -129,11 +129,7 @@ impl Execution {
                 qvm_url: config.qvm_url().into(),
                 source,
             }),
-            Ok(Response::Success(response)) => Ok(response
-                .registers
-                .into_iter()
-                .map(|(key, value)| (key.into_boxed_str(), value))
-                .collect()),
```
Shadow53, line 136: This transformation turned a `HashMap<String, RegisterData>` into a `HashMap<Box<str>, RegisterData>`. Because the output type was changed, this conversion is no longer necessary.

crates/python/src/api.rs
```diff
@@ -0,0 +1,223 @@
+use std::collections::HashMap;
```
Shadow53, line 1: Moved this stuff from `lib.rs` to `api.rs`. It is re-exported at the top level to avoid breakages.

crates/python/src/executable.rs
```diff
@@ -0,0 +1,154 @@
...
+    QcsExecutionError,
+    PyRuntimeError
+);
+
+// Because Python is garbage-collected, no lifetimes can be guaranteed except `'static`.
+//
+// `Arc<Mutex<>>` to work around https://github.com/awestlake87/pyo3-asyncio/issues/50
+py_wrap_type! {
+    PyExecutable(Arc<Mutex<Executable<'static, 'static>>>) as "Executable";
+}
```
Shadow53, line 30: `pyo3_asyncio` requires that all futures be `'static`, which does not work when the future has a reference to `&self`. The workaround for this is to wrap the inner data in an `Arc<Mutex<_>>`, which effectively gives the inner data the `'static` lifetime.

crates/python/src/executable.rs
```diff
@@ -0,0 +1,162 @@
...
+impl PyExecutable {
+    #[new]
+    #[args(
+        "/",
+        registers = "Vec::new()",
+        parameters = "Vec::new()",
+        shots = "None",
+        compile_with_quilc = "None",
+        compiler_options = "None"
+    )]
```
Shadow53, line 53: This does two things:

1) Defines all parameters after the "/" as keyword-only
2) Sets defaults for all of the listed parameters (non-listed are required)

crates/python/src/register_data.rs
```diff
@@ -0,0 +1,15 @@
...
+};
+use qcs::RegisterData;
+use rigetti_pyo3::py_wrap_union_enum;
+
+py_wrap_union_enum! {
+    PyRegisterData(RegisterData) as "RegisterData" {
+        i8: I8 => Vec<Vec<Py<PyInt>>>,
+        f64: F64 => Vec<Vec<Py<PyFloat>>>,
+        i16: I16 => Vec<Vec<Py<PyInt>>>,
+        complex32: Complex32 => Vec<Vec<Py<PyComplex>>>
```
jselig-rigetti, line 13: Does this `complex32` have to agree with the one in [crates/lib/src/api.rs#Register](https://github.com/rigetti/qcs-sdk-rust/pull/230/files#diff-0444195b31be2fc4d234f4a0834b34600c41e782bee4e376391658cb369a1987L216) ?

crates/python/src/api.rs
```diff
@@ -0,0 +1,223 @@
...
+    PyExecutionResults(ExecutionResults) as "ExecutionResults";
+}
+
+py_wrap_union_enum! {
+    PyRegister(Register) as "Register" {
+        f64: F64 => Vec<Py<PyFloat>>,
+        i16: I16 => Vec<Py<PyInt>>,
+        i32: I32 => Vec<Py<PyInt>>,
+        i8: I8 => Vec<Py<PyInt>>,
+        complex64: Complex64 => Vec<Py<PyComplex>>
```
jselig-rigetti, line 77: What's the reason to have this and [crates/python/src/register_data.rs#PyRegisterData](https://github.com/rigetti/qcs-sdk-rust/pull/230/files#diff-13066fc74c4b4c1cf4df7c4976361ec3a2ccdb990406e46b2b3fe27f7e2bb66eR8) and why do the `complex` types disagree?

crates/python/src/qpu/client.rs
```diff
@@ -0,0 +1,68 @@
...
+                .await
+                .map_err(LoadError)
+                .map_err(ToPythonError::to_py_err)?;
+
+            let client = match use_gateway {
+                None => client,
+                Some(use_gateway) => client.with_use_gateway(use_gateway),
+            };
+
+            Python::with_gil(|py| <_ as ToPython<Py<PyAny>>>::to_python(&Self(client), py))
```
jselig-rigetti, line 64: Is this any different than the following? Why invoke this way?

```suggestion
            Python::with_gil(|py| Self(client).to_python(py))
```

crates/python/src/executable.rs
jselig-rigetti: Will require at least python 3.8 - so I think this needs to be updated https://github.com/rigetti/qcs-sdk-rust/blob/2722a59c9f2b472d5d607458be0d310e5d0a503d/crates/python/pyproject.toml#L43

jselig-rigetti (commented):
Overall looks great!
- We should merge the `rigetti-pyo3:support-fieldless-enums-data-structs` branch first, right?
- Do we still need to hand-write the `.pyi` files?

crates/lib/src/api.rs
```diff
@@ -213,7 +210,7 @@ pub enum Register {
     /// A register of 32-bit integers
     I32(Vec<i32>),
     /// A register of 64-bit complex numbers
-    Complex64(Vec<Complex64>),
+    Complex64(Vec<Complex32>),
```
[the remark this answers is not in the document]
MarquessV: This is also a breaking change, so we should have an associated conventional commit so that it shows up in the change logs as such

MarquessV (commented):
This is a big PR, so I'll take another pass later, but looking good so far. I would make sure we use conventional commits to properly call out breaking changes. I called out one on an enum, but it looks like there are others related to the shift to `Cow`

crates/python/src/executable.rs
Shadow53: I'm curious if the `pyo3` stuff uses that PEP or if it does something else. I'm okay with bumping the Python version though, if @kalzoo does not see any issues.

Shadow53:
From @jselig-rigetti:

> We should merge the `rigetti-pyo3:support-fieldless-enums-data-structs` branch first, right?

Correct.

> Do we still need to hand-write the `.pyi` files?

I think so. [This issue](https://github.com/PyO3/pyo3/issues/510) is still open. The `#[args()]` attribute sets `__text_signature__`, which is helpful, but not the same as type stubs.

crates/python/src/api.rs
```diff
@@ -0,0 +1,223 @@
...
+        if let Some(timeout_arg) = kwargs.get_item("timeout") {
+            let timeout: Result<Option<u8>, _> = timeout_arg.extract();
+            if let Ok(option) = timeout {
+                compiler_timeout = option
+            }
+        }
+    }
+
+    pyo3_asyncio::tokio::future_into_py(py, async move {
+        let client = Qcs::load()
```
jselig-rigetti, line 116: Do we want / need to support passing this in as a parameter? If so, what's the right pattern for reconstructing it from a (guessing here) `PyDict` or `PyAny`?

kalzoo: Shouldn't we follow the same pattern as with the rest of the Rust-defined structs and just let Python users manipulate something that is otherwise kept on the Rust side? If the RUst SDK is making all the network calls, it makes sense to me to stick with that as the canonical ClientConfiguration rather than using the one from the `qcs-api-client-python` side. That Python crate shouldn't even need to be imported by `pyquil` once we're done with v4.

crates/python/src/qpu/client.rs
Shadow53: I think this was to specify that it's `ToPython<PyAny>`, not `ToPython<SomethingElse>`. Looks like your suggestion works fine though, so 🤷🏻

crates/lib/src/api.rs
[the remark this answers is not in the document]
Shadow53: Addressed in ef9a28e7

crates/python/Cargo.toml
```diff
@@ -18,12 +18,13 @@ crate-type = ["cdylib"]
...
 qcs = { path = "../lib" }
 qcs-api-client-common = "0.2.7"
+qcs-api-client-grpc = "0.2.7"
 pyo3 = { version = "0.17", features = ["extension-module"] }
 pyo3-asyncio = { version = "0.17", features = ["tokio-runtime"] }
-pythonize = "0.17"
 quil-rs = "0.15"
 tokio = "1.21"
 qcs-api = "0.2.1"
+rigetti-pyo3 = { version = "0.0.1", features = ["extension-module", "complex"], git = "ssh://git@github.com/rigetti/rigetti-pyo3", branch = "support-fieldless-enums-data-structs" }
```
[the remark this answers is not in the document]
Shadow53: Yes

crates/python/src/api.rs
jselig-rigetti: Fix is in https://github.com/rigetti/qcs-sdk-rust/pull/235

crates/python/src/api.rs
Shadow53: Having both is because the Rust library uses both. I am not sure why the variant names disagree.

crates/python/src/register_data.rs
Shadow53: Not sure what you mean. The `Complex32` inner type in both Rust types is the same, just the variant name is different.

crates/python/src/executable.rs
jselig-rigetti: I hope so, since support EOL is in 5 months 😬

crates/python/src/api.rs
jselig-rigetti: As you say, something that would need to be addressed upstream, so IMO resolved in terms of this MR.

crates/python/src/register_data.rs
jselig-rigetti: As you say, something that would need to be addressed upstream, so IMO resolved in terms of this MR.

crates/python/src/executable.rs
kalzoo: I'm good with 3.8.

crates/lib/src/api.rs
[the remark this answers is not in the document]
kalzoo: This is the kind of comment that, while writing it, you ask yourself, "wouldn't it be better to just not have to explain this?"

I'm glad @jselig-rigetti brought it to our attention though, so we can sort it out before it's too late and while we can break things.

Considering the type `[f32;2]` rom the bottom of the stack upwards:

* Our [proto definitions](https://github.com/rigetti/qcs-api-client-rust/blob/f4f693b413fbd530b76e2b7bdd11de67d9b590d5/qcs-api-client-grpc/proto/controller/readout.proto#L20) call this `Complex64`. That's not something we'd want to break.
* `num_complex` uses Complex32
* `numpy`, used by `pyQuil`, uses [`complex64`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.complex64)

To me, 64 makes more sense; it uses 64 bits per unit, and so happens to be two smaller types packed together. You don't call f32 by the size of the mantissa; it takes up 32 bits in memory and you name it as such.

I like the easy-to-follow suggestion in [this relevant issue](https://github.com/rust-num/num-complex/issues/77): just don't use the alias at all, but rather `Complex<f32>`. Then there's no apparent conflict here to confuse folks later on or to require a comment.

crates/python/src/lib.rs
```diff
@@ -1,165 +1,46 @@
...
+pub mod api;
+pub mod executable;
+pub mod execution_data;
+pub mod grpc;
+pub mod qpu;
+pub mod register_data;
+
+// pub use executable::{Error, Executable, ExecuteResultQPU, ExecuteResultQVM, JobHandle, Service};
+// pub use execution_data::{Qpu, Qvm, ReadoutMap};
+// pub use register_data::RegisterData;
```
kalzoo, line 13: remove pls

crates/python/src/api.rs
```diff
@@ -0,0 +1,223 @@
...
+create_exception!(qcs, TranslationError, PyRuntimeError);
+create_exception!(qcs, CompilationError, PyRuntimeError);
+create_exception!(qcs, RewriteArithmeticError, PyRuntimeError);
+create_exception!(qcs, DeviceIsaError, PyValueError);
+
+wrap_error!(SubmitError(qcs::api::SubmitError));
+py_wrap_error!(api, SubmitError, QcsSubmitError, PyRuntimeError);
+
+#[pyfunction(kwds = "**")]
+pub fn compile<'a>(
```
kalzoo, line 96: Could this file use some namespacing so it's not just a collection of functions at the top level?

I'm thinking especially of `get_quilc_version` which seems like it might be suited under a `quilc` module.

It's not urgent to solve now because there are breaking changes on the horizon, but I'd like to solicit opinions before it is urgent.

kalzoo (approved):
Aside from the version bump following upstream PR, this meets the bar to merge. Noting, of course, that we're about to break just about every function in the Python bindings once @MarquessV 's `Program` changes land.

crates/python/src/api.rs
Shadow53: Pinging @MarquessV, @notmgsk, and @randall-fulton, as they are the ones who wrote this code.

This used to be the entirety of the `lib.rs` file, when we were working on a minimum viable product for this. With writing bindings for the entire SDK, do we want to just remove this entirely?

MarquessV: Yeah, everything here was thrown together as part of the initial PoC. I agree that this should all be re-organized into submodules before we pyQuil v4 is launched.

What are you proposing we remove? I think these functions have value still. This one could be lifted into the quilc module you've created, for example.

Shadow53: Not proposing we keep or remove anything specific. When I made my last comment, I was thinking everything in `api` existed already in the SDK in some fashion and we'd prefer to use that instead of the PoC functions.

Since there is existing work built on top of this branch, I'd prefer to create an issue to revisit this and otherwise merge this PR. That work for you?

MarquessV: Yeah, that is fine with me.

MarquessV (approved):
Looks good!

crates/python/src/api.rs
Shadow53: Created #237 to track this.

PR #232, opened 2023-01-13
Re-export quil-rs Python bindings

MarquessV:
This re-exports [quil-rs](https://github.com/rigetti/quil-rs)'s Python bindings as a `quil` submodule. This allows the SDK to manage which version of `quil-rs` it uses and simplifies
dependency management for downstream users who just want to use this SDK.


Shadow53:
> Does anything bad happen when the qcs-sdk and quil are installed in the same environment? What if the versions are different?

Judging from previous experience with two sources of the same code (at least in Rust), if Rust cannot determine that two types are from the same library version, it considers them incompatible, even if they have the same definitions.

In other words, my guess is that the two types would be considered different types with the same name.

MarquessV:
> > Does anything bad happen when the qcs-sdk and quil are installed in the same environment? What if the versions are different?
> 
> Judging from previous experience with two sources of the same code (at least in Rust), if Rust cannot determine that two types are from the same library version, it considers them incompatible, even if they have the same definitions.
> 
> In other words, my guess is that the two types would be considered different types with the same name.

Yeah, doing some discovery on this to confirm the behavior. But I think enforcing a strict `quil` version as a dependency (both as a workspace dependency for the Rust crate and as part of Python package spec) will effectively allow us to enforce a specific version of `quil` be used.

MarquessV:
Closing in favor of #261

PR #234, opened 2023-01-13
Add `list_quantum_processors` helper

### LINKED ISSUE #233: Add `list_quantum_processors` helper
From https://github.com/rigetti/pyquil/issues/1495

jselig-rigetti:
Closes #233
Related to https://github.com/rigetti/pyquil/issues/1496


crates/lib/src/api.rs
```diff
@@ -311,3 +312,42 @@ pub async fn retrieve_results(
...
+/// being found.
+pub async fn list_quantum_processor_names(
+    client: &Qcs,
+) -> Result<Vec<String>, Box<dyn std::error::Error + Send + Sync + 'static>> {
+    let mut quantum_processors = vec![];
+    let mut page_token = None;
+
+    // sanity check
+    let mut calls = 0;
+    loop {
```
Shadow53, line 327: Why not a `for` loop? `for calls in (0..10)` Can error on `page_token.is_some()` instead.

But also, do we expect to ever have 10+ pages of results?

crates/python/src/api.rs
```diff
@@ -88,6 +91,7 @@ create_exception!(qcs, TranslationError, PyRuntimeError);
 create_exception!(qcs, CompilationError, PyRuntimeError);
 create_exception!(qcs, RewriteArithmeticError, PyRuntimeError);
 create_exception!(qcs, DeviceIsaError, PyValueError);
+create_exception!(qcs, QcsApiError, PyRuntimeError);
```
Shadow53, line 94: Along with using a non-boxed Rust error, the `wrap_error`/`py_wrap_error` method should be used here instead.

Shadow53 (changes requested):

crates/lib/src/api.rs
jselig-rigetti: No (or even more than one), but I figured it'd be some protection against endlessly spinning against the API incase there's a bug that makes the pages loop or something. I may just remove pagination.

kalzoo: I think it would be more conventional to just accept a timeout value, which also accounts for more failure modes. Accounting for one particular API failure mode in the client, in a way that introduces a new failure mode, seems unwise to me. Imagine that at one point we do have >1000 QPU IDs (maybe because our tests create them, and an internal user is running this) - now this version of the client will permanently fail until we delete some of those. With a user-configurable timeout, there's at least a client-side workaround to keep this client version usable.

crates/lib/src/api.rs
```diff
@@ -311,3 +312,42 @@ pub async fn retrieve_results(
         .map(ExecutionResults::from)
         .ok_or_else(|| GrpcClientError::ResponseEmpty("Controller Job Execution Results".into()))
 }
+
+/// Use the qcs api client to fetch all available quantum processors.
+/// Will fail if more than 10 pages are requested without all names
+/// being found.
+pub async fn list_quantum_processor_names(
+    client: &Qcs,
+) -> Result<Vec<String>, Box<dyn std::error::Error + Send + Sync + 'static>> {
```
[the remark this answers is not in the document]
jselig-rigetti: Done, just returns the underlying error from the api client

crates/lib/src/api.rs
jselig-rigetti: Removed the pagination.

crates/lib/src/api.rs
```diff
@@ -311,3 +312,42 @@ pub async fn retrieve_results(
         .map(ExecutionResults::from)
         .ok_or_else(|| GrpcClientError::ResponseEmpty("Controller Job Execution Results".into()))
 }
+
+/// Use the qcs api client to fetch all available quantum processors.
```
kalzoo, line 321:
```suggestion
/// Query the QCS API for the names of all available quantum processors.
```

kalzoo (changes requested):

crates/lib/src/api.rs
```diff
@@ -311,3 +319,55 @@ pub async fn retrieve_results(
...
+
+    let mut quantum_processors = vec![];
+    let mut page_token = None;
+
+    loop {
+        if end_time > SystemTime::now() {
+            return Err(ListQuantumProcessorNamesError::TimeoutError(timeout));
+        }
+
+        let result = list_quantum_processors(
```
MarquessV, line 352: It looks like we only effectively check the timeout between calls to the API, but the API call itself could take an arbitrarily long amount of time since it doesn't have it's own timeout. This would lead to us disrespecting the timeout value.

I think we need to wrap this loop in a construct like [tokio::timer::Timeout](https://docs.rs/tokio/0.1.22/tokio/timer/struct.Timeout.html) with the provided timeout.

Shadow53: Adding on to this, I think the best thing to do would be wrap most of the body of the function in an async block and pass the entire thing to `timeout`, i.e.

```rust
let timeout = timeout.unwrap_or_else(|| Duration::from_secs(10));

tokio::time::timeout(timeout, async move {
    let mut quantum_processors = vec![];
    let mut page_token = None;
    loop {
        // .. snip ...
   }

   Ok(quantum_processors)
}).await
```

That way, we don't need to pass a timeout specifically to the API call and don't need to calculate a new timeout for the 2nd+ iterations of the loop.

The only argument against this that I can think of is if we decide we want to return a partial list on timeout, which doesn't seem to be the case here.

crates/python/src/api.rs
```diff
@@ -221,3 +234,16 @@ pub fn get_quilc_version(py: Python<'_>) -> PyResult<&PyAny> {
...
+
+#[pyfunction]
+pub fn list_quantum_processors(py: Python<'_>) -> PyResult<&PyAny> {
+    pyo3_asyncio::tokio::future_into_py(py, async move {
+        let client = Qcs::load()
+            .await
+            .map_err(|e| InvalidConfigError::new_err(e.to_string()))?;
+        let names = list_quantum_processor_names(&client, None)
+            .await
+            .map_err(|e| QcsListQuantumProcessorsError::new_err(e.to_string()))?;
```
Shadow53, line 246: This method is provided by `rigetti-pyo3` and handles all of the wrapper shenanigans. Basically the same effect, but a little shorter.

```suggestion
            .map_err(ListQuantumProcessorsError::to_py_err)?;
```

crates/lib/src/api.rs
jselig-rigetti: Ah, I wasn't sure this would actually cancel the underlying loop or just stop waiting for it - prefer to do it that way 👍

crates/lib/src/api.rs
```diff
@@ -311,3 +316,53 @@ pub async fn retrieve_results(
         .map(ExecutionResults::from)
         .ok_or_else(|| GrpcClientError::ResponseEmpty("Controller Job Execution Results".into()))
 }
+
+/// API Errors encountered when trying to list available quantum processors.
+#[derive(Debug, thiserror::Error)]
+pub enum ListQuantumProcessorNamesError {
+    /// Failed an http call
+    #[error("Failed an http call: {0}")]
```
Shadow53, line 324: This should be more descriptive, something like "List processors API returned an error". Not sure if there should be a hyphen in "List(-)processors" or not.

Shadow53 (commented):
I think the message for an OpenAPI error could be more specific, but otherwise looks good.

crates/python/src/api.rs
```diff
@@ -221,3 +234,16 @@ pub fn get_quilc_version(py: Python<'_>) -> PyResult<&PyAny> {
         Ok(version)
     })
 }
+
+#[pyfunction]
+pub fn list_quantum_processors(py: Python<'_>) -> PyResult<&PyAny> {
```
kalzoo, line 239: Why is this name different than the Rust function, if they do the same thing?

Also, these are IDs rather than names.

kalzoo (changes requested):
Small change in naming - otherwise, approved 🚢

MarquessV (approved):
Thanks for tackling this!

PR #235, opened 2023-01-19
Export rust QCS ClientConfiguration to python

### LINKED ISSUE #231: Export rust QCS ClientConfiguration to python
From https://github.com/rigetti/qcs-sdk-rust/pull/230/files#r1068297511

As a `pyquil` user, I want to be able to provide a custom QCS client configuration when making API calls.

Since [`qcs-api-client`](https://pypi.org/project/qcs-api-client/) is being removed from pyquil as part of https://github.com/rigetti/pyquil/issues/1485, we'll need to replace the python [`QCSClientConfiguration`](https://github.com/rigetti/qcs-api-client-python/blob/main/qcs_api_client/client/_configuration/configuration.py#L25) functionality with the rust [`ClientConfiguration`](https://docs.rs/qcs-api-client-common/latest/qcs_api_client_common/configuration/struct.ClientConfiguration.html).

So far in [qcs-sdk-rust/pull/230](https://github.com/rigetti/qcs-sdk-rust/pull/230/files#r1068297511) there is only support for the environmentally-bootstrapped configuration.

---

<details>
<summary>previous suggestion, no longer relevant</summary>

Since `QCSClientConfiguration` is an object with getters, I'm thinking the solution looks like:
- Convert the object into a `dict`
- Pass it to rust as a `PyDict`
- Construct [`ClientConfiguration`](https://github.com/rigetti/qcs-api-client-rust/blob/main/qcs-api-client-common/src/configuration/mod.rs#L73) as closely as possible (likely requires changes to the crate)

</details>

jselig-rigetti:
Closes #231 

~~Open~~ questions:
- ~~The `info` method helps with testing, but one could mistake the output for being serializable when it isn't really. Any thoughts on if we should provide a serialized view of the client, or what should be inspectable by the client?~~
Implemented the `==` operator overload without exposing `Debug` output.
- ~~How do we add `pyi` type stubs for nested submodules? Should everything be exported at the top level?~~ Use https://www.maturin.rs/project_layout.html#mixed-rustpython-project but only have `.pyi` files.


crates/python/src/qpu/client.rs
```diff
@@ -61,8 +197,11 @@ impl PyQcsClient {
                 Some(use_gateway) => client.with_use_gateway(use_gateway),
             };
 
-            Python::with_gil(|py| <_ as ToPython<Py<PyAny>>>::to_python(&Self(client), py))
-        })?
-        .extract()
+            Ok(Self(client))
+        })
+    }
```
jselig-rigetti, line 202: I had to remove the `extract` or else python would throw the following:

```
TypeError: 'Future' object cannot be converted to QcsClient
```

crates/python/src/qpu/client.rs
```diff
@@ -61,8 +197,11 @@ impl PyQcsClient {
...
+
+    fn info(&self) -> String {
```
MarquessV, line 204: There is an `impl_repr!` macro that will implement this as a `__repr__` method for you.

Is this method replacing some functionality in its Python counterpart?

crates/python/tests/qcs_config/settings.toml
```diff
@@ -3,8 +3,8 @@ default_profile_name = "default"
 [auth_servers]
 
   [auth_servers.default]
-    client_id = ""
-    issuer = "https://rigetticomputing.okta.com/oauth2/ausmoidj5qavd4v310h7"
+    client_id = "0oa3ykoirzDKpkfzk357"
+    issuer = "https://auth.qcs.rigetti.com/oauth2/aus8jcovzG0gW2TUG355"
```
jselig-rigetti, line 7: to match https://github.com/rigetti/qcs-api-client-rust/blob/main/qcs-api-client-common/src/configuration/settings.rs#L157-L159

crates/python/tests/test_client.py
```diff
@@ -0,0 +1,31 @@
...
+def default_client_info():
+    return QcsClient().info()
+
+
+@pytest.mark.asyncio
+async def test_client_empty_profile_is_default(default_client_info):
+    """The profile "empty" is configured to be similar to a default client."""
+    client = await QcsClient.load(profile_name="empty")
+
+    assert client.info() == default_client_info
```
MarquessV, line 16: We could implement `__eq__` for `PyQcsClient` so we don't have to do an equality check on the debug string. Assuming the Rust QcsClient derives eq, we would just return the result of comparing those values.

MarquessV (commented):
[According to maturin](https://github.com/PyO3/maturin/blob/084cfaced651b28616aeea1f818bdc933a536bfe/guide/src/project_layout.md#adding-python-type-information) we want a `<module_name>.pyi` in the root for each module, but it isn't explicit about what to do for submodules. I would try just the module name first to see if that works. If not, maybe try the full module path? Ie. `qcs_sdk.my_module.pyi`

It's not clear about subm

crates/python/src/qpu/client.rs
jselig-rigetti: I'll use that - overall I'm hesitant to expose what may seem at first to be a serializable format but is actually just rust `Debug` output on the off chance that someone needs to parse it. I don't think serialization should be a common or recommended usecase, but I don't want support for it to be ambiguous 

```
Qcs {
  config: ClientConfiguration {
    tokens: Mutex {
      data: Tokens {
        bearer_access_token: Some("..."),
        refresh_token: Some("...")
      }
    },
    api_url: "https://api.qcs.rigetti.com", 
    auth_server: AuthServer {
      client_id: "0oa3ykoirzDKpkfzk357",
      issuer: "https://auth.qcs.rigetti.com/oauth2/aus8jcovzG0gW2TUG355"
    },
    grpc_api_url: "https://legacy.grpc.qcs.rigetti.com",
    quilc_url: "tcp://127.0.0.1:5555",
    qvm_url: "http://127.0.0.1:5000"
  },
  use_gateway: true
}
```

crates/python/tests/test_client.py
jselig-rigetti: I tried this but ran into https://github.com/PyO3/pyo3/issues/2089 and I don't think it makes sense to implement `PartialOrd` here. Will stick with `__repr__` for now.

crates/python/src/qpu/client.rs
MarquessV: Yeah, definitely a fair concern. If we switch to `impl_repr!` we can at least keep this PR moving, and consider it as a potential improvement to `rigetti-pyo3` that will in turn benefit all of these bindings.

crates/python/tests/test_client.py
MarquessV: Ah, good to know! It's not a requirement for me, but curious what you would think about implementing it through `__richcmp__` and using `py.NotImplemented()` for any non equality checking branch, as described [here](https://pyo3.rs/v0.18.0/class/object.html?highlight=__eq__#comparisons)

crates/python/src/qpu/client.rs
jselig-rigetti: I liked your idea from https://github.com/rigetti/qcs-sdk-rust/pull/235#discussion_r1083030237 👍

crates/python/qcs_sdk/qpu/client/__init__.pyi
```diff
@@ -0,0 +1,9 @@
+class QcsClient:
+    """
+    Here we are, deep in the belly of the module
```
kalzoo, line 3: Haha, I do see your PR description, but it looks like this is still getting worked out.

crates/python/tests/test_client.py
```diff
@@ -0,0 +1,31 @@
...
+@pytest.mark.asyncio
+async def test_client_default_profile_is_not_empty(default_client: QcsClient):
+    """The "default" profile is configured to have a token, unlike the default client."""
+    client = await QcsClient.load()
+
+    assert client != default_client
+
+
+@pytest.mark.asyncio
+async def test_client_broken_raises():
```
kalzoo, line 28: nice.

crates/python/src/qpu/client.rs
```diff
@@ -61,8 +203,15 @@ impl PyQcsClient {
...
-        .extract()
+            Ok(Self(client))
+        })
+    }
+
+    fn __richcmp__(&self, other: &Self, op: CompareOp, py: Python<'_>) -> PyObject {
+        match op {
+            CompareOp::Eq => (self == other).into_py(py),
+            CompareOp::Ne => (self != other).into_py(py),
+            _ => py.NotImplemented(),
```
kalzoo, line 214: TIL that this (as you've written) is the right way to do it

https://docs.python.org/3.8/c-api/typeobj.html#c.PyTypeObject.tp_richcompare

kalzoo (commented):
I don't quite follow this sentence:

> The info method helps with testing, but one could mistake the output for being serializable when it isn't really. Any thoughts on if we should provide a serialized view of the client, or what should be inspectable by the client?

Has that already been removed, or am I missing it in review? I think we're okay without serializing the client; it's just a handle on file data anyway (which is a neat serialized format) and there are getters and setters for everything, so I can't think of anything the user would need to do but couldn't.

Overall, this looks good to me, and neat 🗃️ . I'd like to ask for a look from @Shadow53 at the macro invocation for a full approval.

There's also the pyi type hints - I don't have any insight there and would likely just try the same things you are. If nothing else, make sure to clean up any experiments/cruft prior to final review & merge.

Other than that - I'd be happy to approve this.

crates/python/qcs_sdk/qpu/client.pyi
```diff
@@ -0,0 +1,65 @@
...
+
+class QcsClientAuthServer:
+    """Authentication server configuration for the QCS API."""
+
+    client_id: str
+    issuer: str
+
+
+
+class QcsClientTokens:
```
MarquessV, line 45: Does this type hint line up with the implementation? It looks like the conversion to the Rust type is from `PyDict`. Maybe [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) is a better fit?

MarquessV: Same for `QcsClientAuthServer`

MarquessV (commented):
Minor comment on the type hints, but looks good!

crates/python/qcs_sdk/qpu/client.pyi
jselig-rigetti: Good catch, fixed!

MarquessV (approved):
🚀

kalzoo (approved):

crates/python/qcs_sdk/_execution_data.pyi
```diff
@@ -0,0 +1,39 @@
...
+    values: List[int]
+
+
+class ComplexReadoutValues:
+    values: List[complex]
+
+
+class ReadoutValuesValues:
+    integer_values: IntegerReadoutValues
+    complex_values: ComplexReadoutValues
```
Shadow53, line 19: I'm not sure these are actually being exported to Python in the Rust source. Need to go through the types that _are_ exported and make sure all the types they reference are _also_ exported.

crates/python/qcs_sdk/_register_data.pyi
```diff
@@ -0,0 +1,45 @@
...
+
+    def as_i8() -> Optional[List[List[int]]]: ...
+    def as_i16() -> Optional[List[List[int]]]: ...
+    def as_f64() -> Optional[List[List[float]]]: ...
+    def as_complex32() -> Optional[List[List[complex]]]: ...
+
+    def to_i8() -> List[List[int]]: ...
+    def to_i16() -> List[List[int]]: ...
+    def to_f64() -> List[List[float]]: ...
+    def to_complex32() -> List[List[complex]]: ...
```
Shadow53, line 40: These need a `self` parameter

crates/python/qcs_sdk/_register_data.pyi
```diff
@@ -0,0 +1,45 @@
...
+
+    def from_i8(inner: List[List[int]]) -> "RegisterData": ...
+    def from_i16(inner: List[List[int]]) -> "RegisterData": ...
+    def from_f64(inner: List[List[float]]) -> "RegisterData": ...
+    def from_complex32(inner: List[List[complex]]) -> "RegisterData": ...
```
Shadow53, line 45: These need to be marked `@staticmethod`

crates/python/qcs_sdk/api.pyi
```diff
@@ -0,0 +1,326 @@
...
+    """Error encountered rewriting arithmetic for program."""
+    ...
+
+
+class DeviceIsaError(ValueError):
+    """Error while building Instruction Set Architecture."""
+    ...
+
+
+class RewriteArithmeticResults(TypedDict):
```
Shadow53, line 39: This class and below don't actually inherit from `TypeDict` (though we could update the macros to do that if we wanted to)

crates/python/src/qpu/client.rs
```diff
@@ -40,18 +51,150 @@ py_wrap_error!(
...
+        rs -> py {
+            rs_struct: AuthServer => Py<PyDict> {
+                let obj = PyDict::new(py);
+                obj.set_item("client_id", rs_struct.client_id())?;
+                obj.set_item("issuer", rs_struct.issuer())?;
+                Ok(obj.into_py(py))
+            }
+        }
+    }
+}
```
Shadow53, line 90: As written, this is a completely opaque type. Unless Rust pyfunctions call `TryFrom`/`TryInto` on it, objects of this type will be returned to Python as a `QcsClientAuthServer` with no methods or accessible fields. The type has a `__new__(cls, Dict[str, Any])` constructor, but that is it.

Just making sure that is what is intended here. Same for below.

crates/python/qcs_sdk/_executable.pyi
```diff
@@ -0,0 +1,89 @@
+"""
+Do not import this file, it has no exports.
+It is only here to represent the structure of the rust source code 1:1
+"""
+
+from typing import Dict, List, Optional
+from .qpu.quilc import CompilerOpts
```
Shadow53, line 7:
```suggestion
from .qpu.quilc import CompilerOpts
from .. import QVM, QPU
```

Shadow53 (changes requested):
The Rust looks good, except for a couple clarifications. There are things to be corrected with the `pyi` files, though.

I started to comment each instance, but I was basically repeating myself for the same 3-4 things, so I instead left a comment pointing out how a thing should be fixed and then left it as an exercise for the reader to find the others.

I'm open to pairing on this if there are any questions about how the pyi files should be written.

crates/python/qcs_sdk/__init__.py
```diff
@@ -0,0 +1,2 @@
+# See https://www.maturin.rs/project_layout.html#adding-python-type-information
+from .qcs_sdk import *
```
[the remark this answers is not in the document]
jselig-rigetti: The `.so` file that gets generated from all the `pyo3` needs this to export them https://pyo3.rs/v0.18.0/python_typing_hints#__init__py-content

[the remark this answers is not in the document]
jselig-rigetti:
```suggestion
# See https://pyo3.rs/v0.18.0/python_typing_hints#__init__py-content for why this is necessary
from .qcs_sdk import *
```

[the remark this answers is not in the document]
Shadow53: That URL makes more sense.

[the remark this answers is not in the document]
Shadow53: Related, it looks like we need `py.typed` still? That got deleted as part of this PR.

crates/lib/src/api.rs
```diff
@@ -313,3 +315,53 @@ pub async fn retrieve_results(
...
+    TimeoutError(#[from] Elapsed),
+}
+
+/// Query the QCS API for the names of all available quantum processors.
+/// If `None`, the default `timeout` used is 10 seconds.
+pub async fn list_quantum_processors(
+    client: &Qcs,
+    timeout: Option<Duration>,
+) -> Result<Vec<String>, ListQuantumProcessorsError> {
+    let timeout = timeout.unwrap_or_else(|| Duration::from_secs(10));
```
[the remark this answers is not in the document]
jselig-rigetti: I imagine we will, though I'm not sure the timeout value will be the same everywhere - if it is, it should probably live on `qpu::client::Qcs`

crates/python/qcs_sdk/__init__.py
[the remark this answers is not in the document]
jselig-rigetti: Aye - `maturin develop` is functioning happily without it so I figured it wasn't necessary, but it doesn't hurt to add it back.

crates/python/src/qpu/client.rs
jselig-rigetti: I'll make [Tokens](https://github.com/rigetti/qcs-api-client-rust/blob/main/qcs-api-client-common/src/configuration/mod.rs#L65-L70) a `py_wrap_data_struct` since the fields are `pub`, but the [`AuthServer`](https://github.com/rigetti/qcs-api-client-rust/blob/main/qcs-api-client-common/src/configuration/settings.rs#L121-L126) fields are private so this needs a proxy `#[pyclass]` I think

crates/python/qcs_sdk/_executable.pyi
```diff
@@ -0,0 +1,89 @@
...
+        compile_with_quilc: Optional[bool],
+        compiler_options: Optional[CompilerOpts]
+    ) -> "Executable": ...
+
+    async def execute_on_qvm():
+        """
+        Execute on a QVM which must be available at the configured URL (default http://localhost:5000).
+
+        Raises:
+            - ``QcsExecutionError``: When the job failed to execute.
```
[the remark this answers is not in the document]
jselig-rigetti: Made `when` -> `if` changes, and fixed tense.

crates/python/qcs_sdk/_executable.pyi
```diff
@@ -0,0 +1,89 @@
...
+        ...
+
+    async def execute_on_qpu(
+        quantum_processor_id: str
+    ):
+        """
+        Compile the program and execute it on a QPU, waiting for results.
+
+        Raises:
+            - ``QcsExecutionError``: When the job failed to execute.
```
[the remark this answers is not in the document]
jselig-rigetti: Made `when` -> `if` changes, and fixed tense.

crates/python/qcs_sdk/_execution_data.pyi
jselig-rigetti: That's what I was trying to get at with the `Do not import this file...` comment at the top - you aren't able to directly import these structs, but you will encounter these types from `QPU.readout_data.get_readout_values(field, index)`.

Since https://typing.readthedocs.io/en/latest/source/stubs.html#private-definitions suggest that these can be implied to be private by prefixing with an underscore, so I can do that.

crates/python/src/executable.rs
```diff
@@ -31,13 +34,29 @@ py_wrap_type! {
...
 
-#[pyclass]
-#[pyo3(name = "ExeParameter")]
-#[derive(FromPyObject)]
-pub struct PyParameter {
-    name: String,
-    index: usize,
-    value: f64,
+#[derive(Clone, Debug)]
+pub struct ExeParameter {
```
jselig-rigetti, line 38: The `#[pyclass]` wasn't playing well with `wrap_data_struct` for getters/settings and `#[pymethods]` for the constructor - this seems to work though.

crates/lib/src/api.rs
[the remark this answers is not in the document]
jselig-rigetti: Though I think it would make sense to add that to this MR, I addressed this comment but split the larger task of api defaults into a different issue in the interest of not letting this MR continue to balloon. https://github.com/rigetti/qcs-sdk-rust/issues/239

crates/python/qcs_sdk/_execution_data.pyi
```diff
@@ -0,0 +1,50 @@
+"""
+Do not import this file, it has no exports.
```
Shadow53, line 2: I think the properties in this file also have setters

crates/python/src/executable.rs
Shadow53: Yeah, `py_wrap_data_struct` creates a new `pyclass`

crates/python/tests/test_api.py
```diff
@@ -45,7 +45,22 @@ def test_build_patch_values():
     assert patch_values == expected
 
 
+def test_exe_parameters():
+    """Should be able to construct and pass exe parameters"""
+    exe_parameter = qcs_sdk.ExeParameter("a", 0, 0.25)
+    assert str(exe_parameter) == 'ExeParameter { name: "a", index: 0, value: 0.25 }'
```
Shadow53, line 51: Why not assert each field?

crates/python/qcs_sdk/_executable.pyi
```diff
@@ -0,0 +1,89 @@
...
+
+
+class JobHandle:
+    """
+    The result of submitting a job to a QPU.
+    
+    Used to retrieve the results of a job.
+    """
+
+    job_id: str
```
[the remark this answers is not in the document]
jselig-rigetti: Converted all attributes to properties, some have setters some don't

Shadow53 (changes requested):
Left a few notes. The big one is that there are types in the pyi files with `@property` getters but not setters, but I think they get setters from the macros

crates/python/src/executable.rs
jselig-rigetti: It didn't generate the getters or setters. Possibly because the fields were not public though...

crates/python/qcs_sdk/_execution_data.pyi
jselig-rigetti: I figured I'd omit them as these should generally be treated as immutable, but I can add them.

crates/python/src/executable.rs
jselig-rigetti: When I just have
```rs
#[pyclass]
#[pyo3(name = "ExeParameter")]
#[derive(FromPyObject)]
pub struct PyParameter {
    pub name: String,
    pub index: usize,
    pub value: f64,
}
```

I get the following from the test:
```
tests/test_api.py::test_exe_parameters - TypeError: No constructor defined
```

Then adding the following
```rs
#[pymethods]
impl PyParameter {
    #[new]
    pub fn new(name: String, index: usize, value: f64) -> Self {
        Self { name, index, value }
    }
}
```

I get this from the test
```py
    def test_exe_parameters():
        """Should be able to construct and pass exe parameters"""
        exe_parameter = qcs_sdk.ExeParameter("a", 1, 2.5)
>       assert exe_parameter.name == "a"
E       AttributeError: 'builtins.ExeParameter' object has no attribute 'name'
```

I figured the data struct would be the easiest way to add getters/setters, unless there is a recommended way with `#[pyclass]`

crates/python/tests/test_api.py
jselig-rigetti: good point - fixed

crates/python/src/executable.rs
Shadow53: I was thinking it would be an automatic thing, but looks like it works [if you annotate each field](https://pyo3.rs/v0.17.0/class#object-properties-using-pyo3get-set) with `#[pyo3(get, set)]`

jselig-rigetti: .... man I really need to learn to RTFM. Thank you, fixed...

crates/python/qcs_sdk/api.pyi
```diff
@@ -0,0 +1,349 @@
...
+    """Error encountered rewriting arithmetic for program."""
+    ...
+
+
+class DeviceIsaError(ValueError):
+    """Error while building Instruction Set Architecture."""
+    ...
+
+
+class RewriteArithmeticResults:
```
Shadow53, line 38: Do these types in this file still need to have setters?

crates/python/qcs_sdk/qpu/quilc.pyi
```diff
@@ -0,0 +1,29 @@
...
+
+
+class QuilcError(RuntimeError):
+    ...
+
+
+class CompilerOpts:
+    """A set of options that determine the behavior of compiling programs with quilc."""
+
+    @property
```
Shadow53, line 13: Setter?

crates/python/src/qpu/client.rs
```diff
@@ -40,18 +53,154 @@ py_wrap_error!(
...
+    QcsConfigurationBuildError,
+    PyRuntimeError
+);
+
+/// The fields on qcs_api_client_common::client::AuthServer are not public.
+#[derive(Clone)]
+pub struct QcsClientAuthServer {
+    pub client_id: Option<String>,
+    pub issuer: Option<String>,
+}
```
Shadow53, line 69: Could this be replaced with a `pyclass` and get/set?

Shadow53 (changes requested):
A couple more places that might need setters in `pyi`, and a question whether a type can be a `pyclass` instead of a wrapped type.

Shadow53 (approved):
LGTM

PR #240, opened 2023-01-30
feat(python): implement wrappers and convenience methods for getting isa

### LINKED ISSUE #236: Add `get_quantum_processor` helper
From https://github.com/rigetti/pyquil/issues/1496

jselig-rigetti:
Closes #236 

Some outstanding questions as inline comments


crates/python/tests/test_isa.py
```diff
@@ -0,0 +1,50 @@
...
+def test_isa_from_aspen_m_3_json(aspen_m_3_json: str):
+    isa = InstructionSetArchitecture.from_raw(aspen_m_3_json)
+
+    assert isa.architecture.family == Family.Aspen
+
+    node_ids = {node.node_id for node in isa.architecture.nodes}
+    assert len(node_ids) == 80
+
+    # TODO: example ISA has values that round when being converted to f32
+    # assert json.loads(isa.json()) == clean_nones(json.loads(aspen_m_3_json))
```
jselig-rigetti, line 42: Can these be changed to `f64`? https://docs.rs/qcs-api-client-openapi/latest/qcs_api_client_openapi/models/characteristic/struct.Characteristic.html

crates/python/qcs_sdk/qpu/isa.pyi
```diff
@@ -0,0 +1,311 @@
...
+
+    Value "Full"  implies that each node is connected to every other (fully-connected architecture).
+    """
+
+    Full = "Full",
+    Aspen = "Aspen",
+    Ankaa = "Ankaa",
+
+
+class Node:
```
kalzoo, line 28: Did you have to write this all up by hand?

crates/lib/src/qpu/mod.rs
```diff
@@ -27,7 +27,7 @@ pub(crate) use execution::{Error as ExecutionError, Execution};
 /// 1. Problem communicating with QCS
 /// 2. Unauthenticated
 /// 3. Expired token
-pub(crate) async fn get_isa(
+pub async fn get_isa(
```
kalzoo, line 30: I'm a little hesitant about this pattern of lightly wrapping client calls in SDK calls with dubious names (while completely acknowledging that you didn't do that). It doesn't feel like something that's going to stand up well over time.

If nothing else, I think I'd be happier with a full name, `get_instruction_set_architecture` or `fetch_instruction_set_architecture`

crates/python/src/qpu/isa.rs
```diff
@@ -0,0 +1,184 @@
...
+    errors: [
+        QcsIsaSerializationError,
+        QcsIsaError
+    ],
+    funcs: [
+        py_get_isa
+    ],
+}
+
+wrap_error!(IsaSerializationError(serde_json::Error));
```
kalzoo, line 41: Wow, this all looks like it took some patience!

crates/python/src/qpu/isa.rs
```diff
@@ -0,0 +1,184 @@
...
+        benchmarks: Vec<Operation> => Vec<PyOperation>,
+        instructions: Vec<Operation> => Vec<PyOperation>,
+        name: String => Py<PyString>
+    }
+}
+
+#[pymethods]
+impl PyOperation {
+    fn test(&self) -> String {
+        format!("What a great test!")
```
kalzoo, line 139: lol what

crates/python/src/qpu/isa.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+        .map_err(IsaSerializationError::from)
+        .map_err(IsaSerializationError::to_py_err)?;
+        Ok(data)
+    }
+}
+
+#[pyfunction(client = "None")]
+#[pyo3(name = "get_isa")]
+pub(crate) fn py_get_isa(
```
kalzoo, line 170: If the Rust name changes please change this as well

crates/python/tests/test_isa.py
kalzoo: We could try to use the `double` type in the schema and see if Rust picks that up as `f64`. I'm not seeing how best to specify that using `pydantic` -  there is the schema post-processing available in `Config`, but otherwise it seems like `float` types in Python end up as `float` in the schema which seems to be single-precision, and thus `f32`.

JSON numbers themselves are `f64`, so this change would make sense to me.

crates/python/src/qpu/isa.rs
jselig-rigetti: Definitely glad to have these macros though, and it would probably be difficult to have a clean interface around something that wraps nested structs.

crates/python/qcs_sdk/qpu/isa.pyi
jselig-rigetti: Lot of copy pasting, with some polishing

crates/python/tests/test_isa.py
jselig-rigetti: Finally fixed this with https://crates.io/crates/qcs-api-client-openapi/0.5.2 dropping

crates/lib/src/qpu/quilc/isa/qubit.rs
```diff
@@ -239,28 +239,28 @@ mod describe_rx_gates {
             Operator::Gate {
                 arguments: vec![Argument::Int(1)],
                 duration: 50.0,
-                fidelity: 0.996_832_609_176_635_7,
+                fidelity: 0.996_832_6,
```
jselig-rigetti, line 242: @notmgsk The original characteristic from line 221 has this value exactly, the extra digits on the pre-change number I figure is due to `i32 -> i64` conversion which this MR gets rid of, but wanted to double check with you that I wasn't missing anything

```
            Characteristic {
                name: "fRB".to_string(),
                value: 0.996_832_6,
                error: Some(0.000_100_896_78),
                node_ids: Some(vec![1]),
                timestamp: "1970-01-01T00:00:00+00:00".to_string(),
                parameter_values: None,
            },
```

kalzoo: Assuming you mean `f32 -> f64` and so the loss of precision in parsing, then this seems good to me.

kalzoo (approved):

PR #241, opened 2023-02-03
feat: support protoquil flag in compilation (merge latest from main)

jselig-rigetti:
Merges `main` into #204 and resolves conflicts.

Also moves `protoquil` into `CompilerOptions`

### Since this diff is unusable, I recommend reviewing the differences between https://github.com/rigetti/qcs-sdk-rust/compare/main...python-protoquil--v2 and #204 which should be minimal.


MarquessV:
Looks good! I like the `get_kwd` function, I'm sure that's going to be very useful as we build this package out. I'm also glad `CompilerOpts` worked out to be extensible.

I know using `Option<bool>` for the `protoquil` is a carry over from @notmgsk iteration of this PR, but I'm wondering if it should just be a `bool`? Do we get anything meaningful from `None`?

jselig-rigetti:
> I know using `Option<bool>` for `protoquil` is a carry over from @notmgsk iteration of this PR, but I'm wondering if it should just be a `bool`? Do we get anything meaningful from `None`?

@MarquessV since the `struct QuilcParams` uses `Option<bool>` for the RPCQ request and I don't understand the ramifications of defaulting to `false` for the [lisp handler](https://github.com/quil-lang/quilc/blob/9ef4b362fd998b17467831bf90a95fa824cf6eb7/app/src/rpc-server.lisp#L48) or the lisp server, I'm reluctant to change it without guidance

notmgsk:
> I know using Option<bool> for protoquil is a carry over from @notmgsk iteration of this PR, but I'm wondering if it should just be a bool? Do we get anything meaningful from None?

Good question. Yes, it _is_ meaningful, because a value of `None` indicates that the _client_ has no preference and so the _server_ can choose. That is, the server can be configured (at startup) to compile programs as protoquil, but it also allows the client to override that setting by specifically sending a true/false value.

@jselig-rigetti you could add a comment to that effect (as I should have originally) so we don't unintentionally remove that functionality.

notmgsk:
@jselig-rigetti Rather than extending `QuilcParams::new()` to take more arguments, we might make a builder for it (a la `CompilerOpts`). [`derive_builder`](https://crates.io/crates/derive_builder) would make that easy.

PR #243, opened 2023-02-06
feat: support protoquil flag in compilation (rebase fix)

### LINKED ISSUE #242: feat: support protoquil flag from rust
Originally done in https://github.com/rigetti/qcs-sdk-rust/pull/204

Related to https://github.com/rigetti/pyquil/issues/1483

jselig-rigetti:
Fixes https://github.com/rigetti/qcs-sdk-rust/pull/204 since I broke it with https://github.com/rigetti/qcs-sdk-rust/pull/241 (sorry @notmgsk)

Closes #242

### DIFF
--- a/crates/lib/src/qpu/quilc/mod.rs
+++ b/crates/lib/src/qpu/quilc/mod.rs
@@ -42,7 +42,7 @@ pub(crate) fn compile_program(
 ) -> Result<Program, Error> {
     let config = client.get_config();
     let endpoint = config.quilc_url();
-    let params = QuilcParams::new(quil, isa);
+    let params = QuilcParams::new(quil, isa).with_protoquil(options.protoquil);
     let request =
         rpcq::RPCRequest::new("quil_to_native_quil", &params).with_timeout(options.timeout);
     let rpcq_client = rpcq::Client::new(endpoint)
@@ -58,6 +58,10 @@ pub(crate) fn compile_program(
 pub struct CompilerOpts {
     /// The number of seconds to wait before timing out. If `None`, there is no timeout.
     timeout: Option<u8>,
+
+    /// If the compiler should produce "protoquil" as output. If `None`, the default
+    /// behavior configured in the compiler service is used.
+    protoquil: Option<bool>,
 }
 
 /// Functions for building a [`CompilerOpts`] instance
@@ -66,13 +70,23 @@ impl CompilerOpts {
     /// Consider using [`CompilerOpts::default()`] to create an instance with recommended defaults.
     #[must_use]
     pub fn new() -> Self {
-        Self { timeout: None }
+        Self {
+            timeout: None,
+            protoquil: None,
+        }
     }
 
     /// Set the number of seconds to wait before timing out. If set to None, the timeout is disabled.
     #[must_use]
     pub fn with_timeout(&mut self, seconds: Option<u8>) -> Self {
         self.timeout = seconds;
+        *self
+    }
+
+    /// Set if the compiler should produce "protoquil" as output.
+    #[must_use]
+    pub fn with_protoquil(&mut self, protoquil: Option<bool>) -> Self {
+        self.protoquil = protoquil;
         *self
     }
 }
@@ -83,6 +97,7 @@ impl Default for CompilerOpts {
     fn default() -> Self {
         Self {
             timeout: Some(DEFAULT_COMPILER_TIMEOUT),
+            protoquil: None,
         }
     }
 }
@@ -150,6 +165,10 @@ impl QuilcParams {
             protoquil: None,
             args: [NativeQuilRequest::new(quil, isa)],
         }
+    }
+
+    fn with_protoquil(self, protoquil: Option<bool>) -> Self {
+        Self { protoquil, ..self }
     }
 }
 
--- a/crates/lib/src/qpu/quilc/mod.rs
+++ b/crates/lib/src/qpu/quilc/mod.rs
@@ -83,7 +83,8 @@ impl CompilerOpts {
         *self
     }
 
-    /// Set if the compiler should produce "protoquil" as output.
+    /// Set to control whether the compiler should produce "protoquil" as output.
+    /// If `None`, the default behavior configured in the compiler service is used.
     #[must_use]
     pub fn with_protoquil(&mut self, protoquil: Option<bool>) -> Self {
         self.protoquil = protoquil;
--- a/crates/python/src/api.rs
+++ b/crates/python/src/api.rs
@@ -106,8 +106,7 @@ py_wrap_error!(
 /// Returns an error if a value is present but cannot be extracted into `T`.
 fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> PyResult<Option<T>> {
     kwds.and_then(|kwds| kwds.get_item(key))
-        .map(<Option<T>>::extract)
-        .unwrap_or(Ok(None))
+        .map_or(Ok(None), PyAny::extract::<Option<T>>)
 }
 
 #[pyfunction(client = "None", kwds = "**")]
--- a/crates/python/src/api.rs
+++ b/crates/python/src/api.rs
@@ -102,13 +102,12 @@ py_wrap_error!(
     PyRuntimeError
 );
 
-/// Get the keyword `key` value from `kwds` if it is of type `Option<T>`, else `None`.
-/// Note that values present at `key` that cannot be extracted into `Option<T>` are treated as `None`.
-fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> Option<T> {
-    kwds
-        .and_then(|kwds| kwds.get_item(key))
-        .map(|value| value.extract())
-        .transpose()
+/// Get the keyword `key` value from `kwds` if it is of type `Option<T>` and it is present, else `None`.
+/// Returns an error if a value is present but cannot be extracted into `T`.
+fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> PyResult<Option<T>> {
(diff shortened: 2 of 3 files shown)


crates/python/src/api.rs
```diff
@@ -102,6 +102,16 @@ py_wrap_error!(
     PyRuntimeError
 );
 
+/// Get the keyword `key` value from `kwds` if it is of type `Option<T>`, else `None`.
+/// Note that values present at `key` that cannot be extracted into `Option<T>` are discarded.
```
Shadow53, line 106: This seems more accurate?
```suggestion
/// Note that values present at `key` that cannot be extracted into `Option<T>` are treated as `None`.
```

crates/python/src/api.rs
```diff
@@ -102,6 +102,16 @@ py_wrap_error!(
...
+fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> Option<T> {
+    if let Some(kwds) = kwds {
+        kwds.get_item(key).and_then(|value| value.extract().ok())
+    } else {
+        None
+    }
```
Shadow53, line 112:
```suggestion
    kwds
        .and_then(|kwds| kwds.get_item(key))
        .and_then(|value| value.extract().ok())
```

Alternatively, change this to returning a `PyResult<Option<T>>` and

```suggestion
    kwds
        .and_then(|kwds| kwds.get_item(key))
        .map(|value| value.extract())
        .transpose()
```

Shadow53: FWIW, I think returning a `Result` is preferred, because it should probably be an error if the wrong type is provided, rather than ignoring that anything was provided in the first place.

Shadow53 (changes requested):
A few thoughts. Mostly a suggestion for rewriting the `get_kwd` logic, plus suggesting that a `Result` should be returned.

crates/python/src/api.rs
```diff
@@ -102,6 +102,15 @@ py_wrap_error!(
     PyRuntimeError
 );
 
+/// Get the keyword `key` value from `kwds` if it is of type `Option<T>`, else `None`.
+/// Note that values present at `key` that cannot be extracted into `Option<T>` are treated as `None`.
+fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> Option<T> {
```
Shadow53, line 107:
```suggestion
fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> PyResult<Option<T>> {
```

crates/python/src/api.rs
```diff
@@ -102,6 +102,15 @@ py_wrap_error!(
...
+    kwds
+        .and_then(|kwds| kwds.get_item(key))
+        .map(|value| value.extract())
```
Shadow53, line 110: Need to double check this, but clippy may point this out:

```suggestion
        .map(PyAny::extract)
```

Shadow53 (changes requested):
Need to change a return type. Also a nonessential suggestion that I think `clippy` would suggest.

crates/python/src/api.rs
jselig-rigetti: I added this locally at `<T>::extract`

jselig-rigetti: Actually, I'm not sure how pyo3 handles this, but I want to extract it into `Option<T>` so I can handle the case of a key being present, but the value being `None`,

```rs
fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> PyResult<Option<T>> {
    kwds.and_then(|kwds| kwds.get_item(key))
        .map(<Option<T>>::extract)
        .unwrap_or(Ok(None))
}
```

Shadow53: `pyo3` should handle it just fine if you tell it to extract an `Option`. I'd suggest writing it like this:

```rust
fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> PyResult<Option<T>> {
    kwds.and_then(|kwds| kwds.get_item(key))
        .map(PyAny::extract::<Option<T>>)
        .unwrap_or(Ok(None))
}
```

The method is `PyAny::extract<D>`, so you want to specify the type to extract to _after_ the method name.

FWIW, the handling of a `None` dictionary can also be written like this (using my above suggestion as a base):

```rust
fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> PyResult<Option<T>> {
    kwds.and_then(|kwds| kwds.get_item(key))
        .map_or(Ok(None), PyAny::extract::<Option<T>>)
}
```

Shadow53: On second read, does "so I can handle the case of a key being present, but the value being None" mean that "key not present" should be a different return value from "key present, `value = None`"? I think you'd need `Option<Option<T>`` at that point...

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -58,6 +58,10 @@ pub(crate) fn compile_program(
 pub struct CompilerOpts {
     /// The number of seconds to wait before timing out. If `None`, there is no timeout.
     timeout: Option<u8>,
+
+    /// If the compiler should produce "protoquil" as output. If `None`, the default
```
notmgsk, line 62: 👍🏽

crates/lib/src/qpu/quilc/mod.rs
```diff
@@ -151,6 +167,10 @@ impl QuilcParams {
             args: [NativeQuilRequest::new(quil, isa)],
         }
     }
+
+    fn with_protoquil(self, protoquil: Option<bool>) -> Self {
+        Self { protoquil, ..self }
```
notmgsk, line 172: TIL!

https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax

crates/python/src/api.rs
```diff
@@ -102,6 +102,14 @@ py_wrap_error!(
     PyRuntimeError
 );
 
+/// Get the keyword `key` value from `kwds` if it is of type `Option<T>` and it is present, else `None`.
+/// Returns an error if a value is present but cannot be extracted into `T`.
+fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> PyResult<Option<T>> {
+    kwds.and_then(|kwds| kwds.get_item(key))
+        .map(<Option<T>>::extract)
+        .unwrap_or(Ok(None))
```
notmgsk, line 110: Can you get away with `unwrap_or_default()`?

crates/python/src/api.rs
```diff
@@ -102,6 +102,14 @@ py_wrap_error!(
     PyRuntimeError
 );
 
+/// Get the keyword `key` value from `kwds` if it is of type `Option<T>` and it is present, else `None`.
+/// Returns an error if a value is present but cannot be extracted into `T`.
```
notmgsk, line 106: Doesn't look like this returns an error due to the `unwrap_or(Ok(None))`

jselig-rigetti: After the `.map` the type is `Option<Result<Option<T>>` so the `.unrap_or` is unpacking the outer option, which is contingent on `kwds` being `Some(kwds)`

crates/python/src/api.rs
jselig-rigetti: No need to distinct the cases.

Welp, I wrote these tests before learning about [this issue](https://pyo3.rs/v0.13.1/faq.html#i-cant-run-cargo-test-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror) but these are my expecations:

```rs
#[cfg(test)]
mod tests {
    use super::get_kwd;
    use pyo3::{self, types::PyDict, PyResult, Python};

    #[test]
    fn test_get_kwd() {
        pyo3::prepare_freethreaded_python();
        let result: PyResult<()> = Python::with_gil(|py| {
            assert!(
                matches!(get_kwd::<bool>(None, "key"), Ok(None)),
                "kwds `None` should successfully return empty value."
            );

            assert!(
                matches!(get_kwd::<bool>(Some(PyDict::new(py)), "key"), Ok(None)),
                "empty `kwds` should successfully return empty value."
            );

            assert!(
                matches!(
                    get_kwd::<bool>(
                        Some({
                            let kwds = PyDict::new(py);
                            kwds.set_item("key", None::<Option<bool>>)?;
                            kwds
                        }),
                        "key"
                    ),
                    Ok(None)
                ),
                "`kwds` with key but empty value should successfully return empty value."
            );

            assert!(
                matches!(
                    get_kwd::<bool>(
                        Some({
                            let kwds = PyDict::new(py);
                            kwds.set_item("key", true)?;
                            kwds
                        }),
                        "key"
                    ),
                    Ok(Some(true))
                ),
                "`kwds` with key and value of matching type should successfully return that value."
            );

            assert!(
                matches!(
                    get_kwd::<bool>(
                        Some({
                            let kwds = PyDict::new(py);
                            kwds.set_item("key", 4u64)?;
                            kwds
                        }),
                        "key"
                    ),
                    Err(_)
                ),
                "`kwds` with a value of disagreeing type should return an error."
            );

            Ok(())
        });

        result.expect("get_kwd test should pass without issue.");
    }
}

```

crates/python/src/api.rs
jselig-rigetti: Unfortunately no - I think `Ok(None)` is probably the sanest default but makes sense that there isn't any since it's a bit arbitrary
```
`Result<std::option::Option<T>, pyo3::PyErr>: Default` is not satisfied
```

crates/python/src/api.rs
notmgsk: Oh, right. Also if it didn't return a result, type checking would fail, so was a derp moment from me :)

notmgsk (approved):

PR #247, opened 2023-02-07
feat: export get_quilt_calibrations

### LINKED ISSUE #246: feat(python): export get_quilt_calibrations
To support https://github.com/rigetti/pyquil/issues/1494

jselig-rigetti:
Closes #246

### DIFF
--- a/crates/python/qcs_sdk/api.pyi
+++ b/crates/python/qcs_sdk/api.pyi
@@ -35,6 +35,11 @@ class DeviceIsaError(ValueError):
     ...
 
 
+class QcsGetQuiltCalibrationsError(RuntimeError):
+    """Error while fetching Quil-T calibrations."""
+    ...
+
+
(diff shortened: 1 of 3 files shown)


crates/python/qcs_sdk/api.pyi
```diff
@@ -337,31 +360,22 @@ async def get_quilc_version(
     """
     ...
 
-
-async def get_quilc_version(
```
Shadow53, line 341: What happened to `get_quilc_version`? Looks like it got deleted.

Shadow53 (approved):
One item was deleted from the pyi file but not from the exported API, I think? Otherwise, LGTM.

crates/python/qcs_sdk/api.pyi
jselig-rigetti: There was actually 2 duplicate definitions, so I removed the redundant one

PR #249, opened 2023-02-08
feat: export qcs client url settings

### LINKED ISSUE #248: feat(python): export QcsClient URL values via getters
As part of https://github.com/rigetti/pyquil/issues/1494 the pure-python `QCSClientConfiguration` needs to be replaced with the pyo3-python `QcsClient` from this repo.

Parity support requires making the `quilc_url` and `qvm_url` available to pyquil, which still interact directly with these services, even though issues like https://github.com/rigetti/qcs-sdk-rust/issues/207 plan to change that in the future.

jselig-rigetti:
Closes #248 

Supports https://github.com/rigetti/pyquil/issues/1494

### DIFF
--- a/crates/python/src/qpu/client.rs
+++ b/crates/python/src/qpu/client.rs
@@ -210,6 +210,26 @@ impl PyQcsClient {
         })
     }
 
+    #[getter]
+    pub fn api_url(&self) -> String {
+        self.0.get_config().api_url().to_string()
+    }
+
+    #[getter]
+    pub fn grpc_api_url(&self) -> String {
+        self.0.get_config().grpc_api_url().to_string()
+    }
+
+    #[getter]
+    pub fn quilc_url(&self) -> String {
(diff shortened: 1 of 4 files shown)


crates/python/src/qpu/client.rs
```diff
@@ -210,6 +210,26 @@ impl PyQcsClient {
         })
     }
 
+    #[getter]
+    pub fn api_url(&self) -> String {
+        self.0.get_config().api_url().to_string()
```
MarquessV, line 215: For consistency, can we `self.as_inner()` via the `rigetti-pyo3` `PyWrapper` trait?

jselig-rigetti: Didn't find a receiver method `.as_inner`, but `.as_ref` does the same 👍

kalzoo (approved):

Issue #251, opened 2023-02-16
pyo3_runtime.PanicException: `spawn_local` called from outside of a `task::LocalSet`


jselig-rigetti:
Received the above error while testing [PyExecutable::execute_on_qvm](https://github.com/rigetti/qcs-sdk-rust/blob/41c5b0112172077a46f61acd445b974fb725bce5/crates/python/src/executable.rs#L96) by adding the following test to [tests/test_api.py](https://github.com/rigetti/qcs-sdk-rust/blob/main/crates/python/tests/test_api.py):

```py
@pytest.mark.asyncio
async def test_translate_exe(native_bitflip_program: str):
    exe = qcs_sdk.Executable(native_bitflip_program)
    results = await exe.execute_on_qvm()
```

jselig-rigetti:
Related to https://github.com/rigetti/pyquil/issues/1473

jselig-rigetti:
Turns out I was using `pyo3_asyncio::tokio::local_future_into_py` when I should have been using `pyo3_asyncio::tokio::future_into_py`

PR #252, opened 2023-02-16
fix!: convert all pyo3 async functions to sync

jselig-rigetti:
This is an alternative to https://github.com/rigetti/pyquil/pull/1531 and partially a response to #251

Since it's unclear how to get around using `async/await` from `pyquil` and recent  other issues like #251, one option is to provide all the methods as synchronous by default - then later come back and add async variants as needed, using a common pattern of postfixing `_async` to the function name, e.g. 

```py
assert do_thing() == await do_thing_asnyc()
```


crates/python/src/lib.rs
```diff
@@ -50,3 +50,39 @@ create_init_submodule! {
...
+    /// async fn say_hello(name: String) -> String {
+    ///     format!("hello {name}")
+    /// }
+    ///
+    /// #[pyo3(name="say_hello")]
+    /// pub fn py_say_hello(name: String) -> PyResult<String> {
+    ///     py_sync!(async move {
+    ///         say_hello(name)
+    ///     })
+    /// }
```
Shadow53, line 70: I feel like this should be one of these:

```suggestion
    /// #[pyo3(name="say_hello")]
    /// pub fn py_say_hello(name: String) -> PyResult<String> {
    ///     py_sync!(async move {
    ///         say_hello(name).await
    ///     })
    /// }
```

```suggestion
    /// #[pyo3(name="say_hello")]
    /// pub fn py_say_hello(name: String) -> PyResult<String> {
    ///     py_sync!(say_hello(name))
    /// }
```

crates/python/src/api.rs
```diff
@@ -131,27 +131,25 @@ fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> PyRe
...
-) -> PyResult<&'a PyAny> {
+) -> PyResult<String> {
     let target_device: TargetDevice =
         serde_json::from_str(&target_device).map_err(|e| DeviceIsaError::new_err(e.to_string()))?;
 
     let compiler_timeout = get_kwd(kwds, "timeout")?.or(Some(DEFAULT_COMPILER_TIMEOUT));
     let protoquil: Option<bool> = get_kwd(kwds, "protoquil")?;
 
-    pyo3_asyncio::tokio::future_into_py(py, async move {
+    crate::utils::py_sync!(async move {
```
Shadow53, line 146: Might prefer importing the macro instead of using the full path every time.

Shadow53 (approved):
A couple small suggestions, but this looks overall fine to me.

crates/python/src/lib.rs
kalzoo: Agreed, and the latter seems cleaner

crates/python/src/lib.rs
```diff
@@ -50,3 +50,39 @@ create_init_submodule! {
 fn qcs_sdk(py: Python<'_>, m: &PyModule) -> PyResult<()> {
     init_submodule("qcs_sdk", py, m)
 }
+
+pub(crate) mod utils {
```
kalzoo, line 54: Can this be its own file?

Also, `utils` is usually a name of last resort, "things which are useful."

crates/python/src/qpu/client.rs
```diff
@@ -185,12 +184,8 @@ impl PyQcsClient {
...
     #[staticmethod]
     #[args("/", profile_name = "None", use_gateway = "None")]
-    pub fn load(
-        py: Python<'_>,
-        profile_name: Option<String>,
-        use_gateway: Option<bool>,
-    ) -> PyResult<&PyAny> {
-        future_into_py(py, async move {
+    pub fn load(profile_name: Option<String>, use_gateway: Option<bool>) -> PyResult<Self> {
+        crate::utils::py_sync!(async move {
```
kalzoo, line 188: The pattern I'd suggest, while we're here: a `mod blocking` like `reqwest` [does](https://docs.rs/reqwest/latest/reqwest/blocking/index.html) . Could also be a crate feature in the same way.

Then the primary `async` methods can still be available. Because, in principle, `async` makes sense here for I/O blocking calls, and those should be accessible from the get-go in these bindings, even if `pyquil` doesn't actually use them yet. Don't _make_ the user use the sync wrapper, right?

kalzoo (changes requested):

crates/python/src/lib.rs
jselig-rigetti: made a `py_sync.rs` file

crates/python/src/qpu/client.rs
jselig-rigetti: By crate feature do you mean like a python "[extra](https://peps.python.org/pep-0508/#extras)"?

Since a lot of the code would be duplicated between a `blocking` and "standard" module and the underlying function implementations are identical, I was thinking something like the following. 

```rs
impl PyQcsClient {
    async fn load(profile_name: Option<String>, use_gateway: Option<bool>) -> PyResult<Self> {
        Ok(Self(/* ... */))
    }
}

#[pymethods]
impl PyQcsClient {
    // ...

    #[staticmethod]
    #[args("/", profile_name = "None", use_gateway = "None")]
    #[pyo3(name = "load")]
    pub fn py_load(profile_name: Option<String>, use_gateway: Option<bool>) -> PyResult<Self> {
        py_sync!(Self::load(profile_name, use_gateway))
    }

    #[staticmethod]
    #[args("/", profile_name = "None", use_gateway = "None")]
    #[pyo3(name = "load_async")]
    pub fn py_load_async(
        py: Python<'_>,
        profile_name: Option<String>,
        use_gateway: Option<bool>,
    ) -> PyResult<&PyAny> {
        py_async!(py, Self::load(profile_name, use_gateway))
    }
}
```

kalzoo: Ah - yeah the way you have it here should be good. Idea with the crate feature was to reduce compile time / binary size, but since we're distributing as a wheel in most cases that does no good. That might be better as an added feature in the SDK itself.

crates/python/src/py_sync.rs
```diff
@@ -0,0 +1,106 @@
...
+/// }
+/// ```
+///
+/// becomes in python:
+/// ```py
+/// from things import do_stuff, do_stuff_async
+/// assert do_stuff() == "done"
+/// assert await do_stuff() == "done"
+/// ```
+macro_rules! py_function_dual {
```
kalzoo, line 73: macros on macros!

Looks good. I'd consider a name that's more obvious in intent - maybe `py_function_sync_async` or something. No doubt we'll run into constraints with this pattern in the future, but (a) writing the functions by hand is an easy escape hatch and (b) it's valuable today and clear in what it does.

crates/python/src/executable.rs
```diff
@@ -93,54 +115,56 @@ impl PyExecutable {
...
-                .await
-                .execute_on_qvm()
-                .await
-                .map(ExecutionData::from)
-                .map(|qvm| Python::with_gil(|py| qvm.to_python(py)))
-                .map_err(ExecutionError::from)
-                .map_err(ExecutionError::to_py_err)?
-        })
+    #[pyo3(name = "execute_on_qvm")]
+    pub fn py_execute_on_qvm(&self) -> PyResult<PyExecutionData> {
```
Shadow53, line 119: All of these methods could have the `py_` prefix and `name = ` removed to get the same effect.

crates/python/src/py_sync.rs
```diff
@@ -0,0 +1,106 @@
...
+        let handle = runtime.spawn($body);
+        runtime
+            .block_on(handle)
+            .map_err(|err| PyRuntimeError::new_err(err.to_string()))?
+    }};
+}
+
+/// Convert a rust future into a Python awaitable using
+/// `pyo3_asyncio::tokio::local_future_into_py`
+macro_rules! py_async {
```
Shadow53, line 33: Thoughts on naming this `py_local_async` and have a `py_async` using `future_into_py`?

crates/python/src/qpu/client.rs
```diff
@@ -127,6 +128,25 @@ impl PyQcsClient {
...
+            .map_err(LoadError)
+            .map_err(ToPythonError::to_py_err)?;
+
+        let client = match use_gateway {
+            None => client,
+            Some(use_gateway) => client.with_use_gateway(use_gateway),
+        };
+
+        Ok(Self(client))
+    }
```
Shadow53, line 149: Could you use `py_function_dual` for this as well?

Shadow53 (commented):
A bunch of nitpicky comments, but overall this looks really good!

crates/python/src/py_sync.rs
```diff
@@ -0,0 +1,106 @@
...
+    ) => {
+        async fn $name($($arg: $kind,)*) -> $ret {
+            $body
+        }
+
+        ::paste::paste! {
+        #[::pyo3::pyfunction$((
+            $($k = $v),*
+        ))?]
+        #[pyo3(name = $name "")]
```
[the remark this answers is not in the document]
jselig-rigetti: It's part of the paste macro, https://docs.rs/paste/latest/paste/#pasting-documentation-strings

I tried the `stringify` but it wasn't a valid literal token 😞

crates/python/src/py_sync.rs
```diff
@@ -0,0 +1,106 @@
...
+        pub fn [< py_ $name >]($($arg: $kind,)*) -> $ret {
+            py_sync!($name($($arg, )*))
+        }
+
+        #[::pyo3::pyfunction$((
+            $($k = $v),*
+        ))?]
+        #[pyo3(name = $name "_async")]
+        pub fn [< py_ $name _async >](py: Python<'_>, $($arg: $kind,)*) -> PyResult<&PyAny> {
```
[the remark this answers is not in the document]
jselig-rigetti: I will take every comment I can get! 😄

crates/python/src/api.rs
```diff
@@ -130,29 +134,53 @@ fn get_kwd<'a, T: FromPyObject<'a>>(kwds: Option<&'a PyDict>, key: &str) -> PyRe
...
+pub fn py_compile_async<'py>(
+    py: Python<'py>,
+    quil: String,
+    target_device: String,
+    client: Option<PyQcsClient>,
+    kwds: Option<&'py PyDict>,
+) -> PyResult<&'py PyAny> {
+    let options = get_compiler_opts(kwds)?;
+    py_async!(py, compile(quil, target_device, client, options))
 }
```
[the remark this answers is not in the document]
jselig-rigetti: It's needing to pre-process `options` before calling an async function because `PyDict` does not implement `Send`, but I would be happy to pair on the lifetimes issue!

crates/python/src/py_sync.rs
jselig-rigetti: Just changed the `py_async` to use `future_into_py` which works great, don't know why I was using `local_future_into_py` but it's unnecessary. Good catch!

crates/python/src/qpu/client.rs
jselig-rigetti: With some modification, yes - need to add a branch for handling between `#[staticmethod]` and `#[pyfunction]`

jselig-rigetti: I tried this, but ran into an issue: since the outer `impl` block has a proc macro on it, things got complicated:

```rs
#[pymethods]
impl PyClient {
    #[staticmethod] // <- this is fine
    #[args("/", profile_name = "None", use_gateway = "None")]
    /* async fn load ... */

    py_function_sync_async! {
        #[staticmethod] // <- error, `staticmethod` not found in scope
        #[args("/", profile_name = "None", use_gateway = "None")]
        /* async fn load ... */
    }
}
```

crates/python/qcs_sdk/api.pyi
```diff
@@ -353,14 +491,16 @@ async def list_quantum_processors(
...
+def get_quilt_calibrations(
+    quantum_processor_id: str,
     client: Optional[QcsClient] = None,
     timeout: Optional[float] = None,
 ) -> QuiltCalibrations:
     """
     Retrieve the calibration data used for client-side Quil-T generation.
 
     Args:
+        quantum_processor_id: The ID of the quanutum processor the job ran on.
```
Shadow53, line 503:
```suggestion
        quantum_processor_id: The ID of the quantum processor the job ran on.
```

crates/python/qcs_sdk/api.pyi
```diff
@@ -370,17 +510,23 @@ async def get_quilt_calibrations(
...
+) -> QuiltCalibrations:
     """
-    Retrieve the InstructionSetArchitecture (ISA) for the given Quantum Processor ID.
+    Async version of `get_quilt_calibrations``
+
+    Retrieve the calibration data used for client-side Quil-T generation.
 
     Args:
-        quantum_processor_id: The ID of the quantum processor
+        quantum_processor_id: The ID of the quanutum processor the job ran on.
```
Shadow53, line 524:
```suggestion
        quantum_processor_id: The ID of the quantum processor the job ran on.
```

crates/python/qcs_sdk/_executable.pyi
```diff
@@ -22,24 +22,54 @@ class Executable:
...
         Execute on a QVM which must be available at the configured URL (default http://localhost:5000).
 
         Raises:
             - ``QcsExecutionError``: If the job fails to execute.
         """
         ...
-    async def execute_on_qpu(self, quantum_processor_id: str) -> ExecutionData:
+    async def execute_on_qvm_async(self) -> ExecutionData:
         """
+        Async version of ``execute_on_qvm``.
```
Shadow53, line 35: Is this supposed to be markdown or some other format? If markdown, it should be a single tick on each side (same in other places)

crates/python/qcs_sdk/api.pyi
```diff
@@ -216,6 +216,36 @@ async def compile(
...
+) -> str:
+    """
+    Async version of ``compile``.
+
+    Uses quilc to convert a quil program to native Quil.
+
+    Args:
+        quil: A Quil program.
+        target_device: A JSON encoded description of the Quantum Abstract Machine Architecture.
+        client: The QcsClient to use. Loads one using environment configuration if unset - see https://docs.rigetti.com/qcs/references/qcs-client-configuration
```
Shadow53, line 234:
```suggestion
        client: The ``QcsClient`` to use. Creates one using environment configuration if unset - see https://docs.rigetti.com/qcs/references/qcs-client-configuration
```

crates/python/qcs_sdk/_executable.pyi
Shadow53: I am guessing that the double backtick is used to link to other defined names (classes, functions, etc.), but I am not familiar with the syntax.

crates/python/qcs_sdk/api.pyi
Shadow53: This suggestion also applies to the rest of the file. I see that it was copied from existing comments (at least existing relative to the "view commits since your last review" view), but I think consistency with formatting class names is important, and I think it's more correct and makes more sense to say a new client is created, rather than loaded.

Shadow53: Also, with the syntax being used here, is there a way to expressly indicate that something is a link, similar to markdown with `<https://example.com>`?

crates/python/qcs_sdk/api.pyi
```diff
@@ -216,6 +216,36 @@ async def compile(
...
+
+    Keyword Args:
+        timeout: The number of seconds to wait before timing out. If set to None, there is no timeout (default: 30).
+
+    Returns:
+        An Awaitable that resolves to the native Quil program.
```
Shadow53, line 240:
```suggestion
        An ``Awaitable`` that resolves to the native Quil program.
```

crates/python/qcs_sdk/api.pyi
```diff
@@ -325,7 +407,31 @@ async def retrieve_results(
...
+    client: Optional[QcsClient] = None,
+) -> ExecutionResults:
+    """
+    Async version of ``retrieve_results``
+
+    Fetches results for the corresponding job ID.
+
+    Args:
+        job_id: The ID of the job to retrieve results for.
+        quantum_processor_id: The ID of the quanutum processor the job ran on.
```
Shadow53, line 422:
```suggestion
        quantum_processor_id: The ID of the quantum processor the job ran on.
```

crates/python/qcs_sdk/qpu/client.pyi
```diff
@@ -21,13 +21,26 @@ class QcsClient:
...
+        ...
+    @staticmethod
+    async def load_async(
+        profile_name: Optional[str] = None,
+        use_gateway: Optional[bool] = None,
+    ) -> "QcsClient":
+        """
+        Async version of ``QcsClient.load()``
+
+        Load a QcsClient configuration using an environment-based configuration.
```
Shadow53, line 42:
```suggestion
        Load a ``QcsClient`` using an environment-based configuration.
```

crates/python/qcs_sdk/qpu/isa.pyi
```diff
@@ -300,11 +300,36 @@ class InstructionSetArchitecture:
...
+    Fetch the `InstructionSetArchitecture` (ISA) for a given `quantum_processor_id` from the QCS API.
+
+    Args:
+        quantum_processor_id: The ID of the quantum processor
+        client: The QcsClient to use. Loads one using environment configuration if unset - see https://docs.rigetti.com/qcs/references/qcs-client-configuration
 
-async def get_instruction_set_architecture(quantum_processor_id: str, client: QcsClient = ...) -> InstructionSetArchitecture:
+    Raises:
+        - `QcsLoadError` if there is an issue with the QCS API client configuration.
+        - `QcsIsaError` if there is an issue fetching the ISA from the QCS API.
```
Shadow53, line 316: Inconsistency between single and double backticks?

crates/python/src/py_sync.rs
Shadow53: The difference IIRC is that `future_into_py` requires the `Future` to be either `Send` or `Sync` (or both), while `local_future_into_py` doesn't have the same restriction. I think it's the difference of whether the future can be run on a different thread or not.

I think it's good to have the option to use `local_future_into_py`, but that can also be added when needed.

crates/python/src/qpu/isa.rs
```diff
@@ -146,19 +147,18 @@ impl PyInstructionSetArchitecture {
...
-#[pyfunction(client = "None")]
-#[pyo3(name = "get_instruction_set_architecture")]
-pub(crate) fn py_get_instruction_set_architecture(
-    quantum_processor_id: String,
-    client: Option<PyQcsClient>,
-) -> PyResult<PyInstructionSetArchitecture> {
-    py_sync!(async move {
+py_function_sync_async! {
+    #[pyfunction(client = "None")]
+    async fn get_instruction_set_architecture(
```
Shadow53, line 152: I'd be curious to see how doc comments are handled by these macros. They should be captured as part of the `meta` rule, since `/// doc comment` is syntactic sugar for `#[doc = "doc comment"]`

Shadow53 (commented):
Responded to the conversation about `future_into_py` vs `local_future_into_py`, otherwise I think all my comments are about documentation. The code itself looks good.

crates/python/qcs_sdk/__init__.pyi
```diff
@@ -3,6 +3,7 @@ from .api import *
 from .qpu.client import QcsClient as QcsClient
 from .qpu.isa import (
     get_instruction_set_architecture as get_instruction_set_architecture,
+    get_instruction_set_architecture_async as get_instruction_set_architecture_async,
```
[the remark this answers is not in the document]
jselig-rigetti: Ran into this issue if I don't: https://stackoverflow.com/a/70546787/2097348

crates/python/qcs_sdk/api.pyi
jselig-rigetti: Seems to be different for different documentation frameworks - tried a bunch of variants of https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects but none of them made links in VSCode.

I'll keep the double backtics for now for consistency with the rest of the documentation

crates/python/src/qpu/isa.rs
jselig-rigetti: huh - it seems to work in that it applies the comment to the generated/public methods, but not to the internal source function

crates/python/qcs_sdk/_executable.pyi
jselig-rigetti: Same answer as https://github.com/rigetti/qcs-sdk-rust/pull/252#discussion_r1113477496 - double-backtick for a cross-reference is a standard I've seen so far, though my tooling (VSCode) doesn't do anything with it. I'll make the changes to bring inconsistencies in line with the pattern, I see no clear answer.

Shadow53 (approved):
Gave this a quick skim with focus on the hygiene of the new macros. The use of `::` prefixed paths seems right to me, and the fixes to docs look good too. Assuming things work correctly, this LGTM

crates/python/src/py_sync.rs
```diff
@@ -0,0 +1,100 @@
...
+///         Ok(String::from("done"))
+///     }
+/// }
+/// ```
+///
+/// becomes in python:
+/// ```py
+/// from things import do_stuff, do_stuff_async
+/// assert do_stuff() == "done"
+/// assert await do_stuff() == "done"
```
MarquessV, line 71: Minor doc note: the function name changes from `do_thing` to `do_stuff`

MarquessV (approved):

crates/python/src/py_sync.rs
jselig-rigetti: 🦅 👁️

Issue #237, opened 2023-01-24
Refactor `api` module into appropriate modules


Shadow53:
The `api` module was written as a proof of concept for writing Python bindings. Now that we've proved the concept, we should look at removing the `api` module and moving the useful parts into appropriate modules.

MarquessV:
Goals:

1. Consistent style
2. Consensus on organization
3. A summary of our API would seem cohesive and well thought out

Idea: Model after the rust crate.

jselig-rigetti:
Currently the python `qcs_sdk` module fairly closely reflects the library `qcs` module, and I'm not sure any drastic changes are warranted.

Here's a potential way to split up the current `qcs_sdk::api`:

These make calls to the QCS API,  
maybe these are the "true" `qcs_sdk::api`
- `py_get_quilt_calibrations`
- `py_list_quantum_processors` 
- `py_retrieve_results`
- `py_submit`
- `py_translate`

These make calls to the local quil compiler,  
maybe `qcs_sdk::compiler` or `qcs_sdk::quilc`
- `py_compile`
- `py_get_quilc_version` 

These are in-process, part of `rewrite_arithmetic`,
maybe `qcs_sdk::program`
- `build_patch_values` 
- `rewrite_arithmetic`

MarquessV:
Maybe we continue to just map everything 1-to-1 with the Rust crate? I think the API module can just go away entirely. The API module on the Rust side existed solely to map things lazily across the FFI, but now that we have all the Python types we need, we can just get rid of them. The Python bindings can call the functions they pass through to directly from a module with the same path.

PR #255, opened 2023-03-01
feat!: remove intermediate `api` module

jselig-rigetti:
Closes https://github.com/rigetti/qcs-sdk-rust/issues/237

The intermediary `api` modules in both `lib` and `python` crates are no longer necessary thanks to [rigetti-pyo3](https://github.com/rigetti/rigetti-pyo3), so here they are removed.

The new structure, focusing on python-importable modules:
- `qcs_sdk` batteries-included executable + related classes, also reexports auth client
- `qcs_sdk::compiler::quilc` compile_program, get_version_info
- `qcs_sdk::qpu` list_qpus, QPUResultData
- `qcs_sdk::qpu::api` submit, retrieve jobs
- `qcs_sdk::qpu::client` auth client
- `qcs_sdk::qpu::isa` get_isa and ISA class
- `qcs_sdk::qpu::rewrite_arithmetic` rewrite_arithmethic, build_patch_values
- `qcs_sdk::qpu::translation` get_quilt_calibrations, translate
- `qcs_sdk::qvm` QVMResultData

Additional notes:
- Types in `crate/lib` were left mostly as-is, intermediate types are defined in `crates/python` as necessary.
- Unnecessarty parts of `rpcq` were removed
- Documentation styling was updated to be more consistent
- Error naming should be more consistent

Still TODO: more test coverage.


crates/lib/src/qpu/api.rs
```diff
@@ -0,0 +1,106 @@
+//! runner
```
MarquessV, line 1: Should we update this now?

crates/python/qcs_sdk/qpu/translation.pyi
```diff
@@ -0,0 +1,122 @@
...
+    def settings_timestamp(self) -> Optional[str]:
+        """ISO8601 timestamp of the settings used to generate these calibrations."""
+        ...
+    @settings_timestamp.setter
+    def settings_timestamp(self, value: Optional[str]): ...
+
+
+class TranslationResult:
+    """
+    The result of a call to [`translate`] which provides information about the translated program.
```
MarquessV, line 25:
```suggestion
    The result of a call to `translate` which provides information about the translated program.
```

crates/python/tests/compiler/test_quilc.py
```diff
@@ -0,0 +1,70 @@
...
+)
+
+@pytest.fixture
+def target_device(aspen_m_3_isa: InstructionSetArchitecture) -> TargetDevice:
+    return TargetDevice.from_isa(aspen_m_3_isa)
+
+
+def test_default_compiler_timeout():
+    """Ensure this exists and is the correct type."""
+    assert type(DEFAULT_COMPILER_TIMEOUT) is float
```
MarquessV, line 22: Not for this PR, but I wonder if we should consider something like [stubtest](https://mypy.readthedocs.io/en/stable/stubtest.html) to help validate that our stubs match what our package actually exports.

MarquessV (commented):
Looking good! Some minor comments/observations. Will make another pass when the tests are ready.

crates/python/tests/compiler/test_quilc.py
MarquessV: There is also [stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html), which would be a huge boon. Unclear if it will work with the bindings though.

MarquessV: Awesome! I'm going to use stubtest for the quil-py stuff too. Will be good to have some real validation against the stubs.

I think we probably went down a similar rabbit hole with `stubgen`. I tried that and a few other tools that do something similar with no luck. It might just not be possible for bindings.

crates/python/qcs_sdk/_executable.pyi
```diff
@@ -2,81 +2,105 @@
...
 
-class QCSExecutionError(RuntimeError):
-    """Error encounteted when executing programs."""
 
+class ExecutionError(RuntimeError):
+    """Error encountered when executing a program."""
     ...
 
+
+@final
```
MarquessV, line 18: Nice! Never thought to put `@final` on these but they totally should be.

crates/python/qcs_sdk/qpu/rewrite_arithmetic.pyi
```diff
@@ -0,0 +1,69 @@
+
```
MarquessV, line 1: the tiniest of nits: extra newline at the top here.

MarquessV (approved):
Looking good! I like the improvements to the Python tests as well as the added ergonomics from the custom options and tasks.

crates/python/qcs_sdk/_executable.pyi
jselig-rigetti: `stubtest` insisted I add it 😂

crates/python/qcs_sdk/qpu/rewrite_arithmetic.pyi
jselig-rigetti:
```suggestion
```

Issue #258, opened 2023-03-13
Bug: release-python GitHub Action fails from maturin crash


jselig-rigetti:
The output:

```
maturin build --target x86_64-apple-darwin -i python --release --manifest-path crates/python/Cargo.toml --out dist --sdist
🍹 Building a mixed python/rust project
🔗 Found pyo3 bindings
🐍 Found CPython 3.7m at /Users/runner/hostedtoolcache/Python/3.7.15/x64/bin/python
📡 Using build options bindings, compatibility from pyproject.toml
Well, this is embarrassing.

maturin had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "...". Submit an issue or email with the subject of "maturin Crash Report" and include the report as an attachment.

- Homepage: https://github.com/pyo3/maturin
- Authors: konstin <konstin@mailbox.org>, messense <messense@icloud.com>

We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!
Error: The process 'maturin' failed with exit code 101
    at ExecState._setResult (/Users/runner/work/_actions/messense/maturin-action/v1/dist/index.js:1702:25)

PR #259, opened 2023-03-13
fix!: remove python 3.7 support

jselig-rigetti:
From https://github.com/rigetti/qcs-sdk-rust/issues/258

Seems the issue may be specifically the 3.7 build path.
Since `pyproject` specifies that the required python version be `^3.8`, seems 3.7 support could be removed anyway.

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -19,7 +19,7 @@ jobs:
     needs: is-python-release
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


notmgsk:
Probably fails because `poetry install` doesn't like the 3.7. Though not sure why the error report doesn't include that info.

notmgsk (approved):

MarquessV (approved):

PR #261, opened 2023-03-17
Specify `quil` as a dependency

MarquessV:
This PR catches `qcs-sdk-rust` up with the latest release of `quil-rs`, and specifies `quil` as a dependency of `qcs-sdk-python` so that a consistent version of the `quil-rs` is used throughout.


Cargo.toml
```diff
@@ -6,7 +6,7 @@ qcs-api = "0.2.1"
 qcs-api-client-common = "0.4.2"
 qcs-api-client-grpc = "0.4.2"
 qcs-api-client-openapi = "0.5.2"
-quil-rs = "0.15"
+quil-rs = { git = "https://github.com/rigetti/quil-rs", branch = "1455-python-support-for-program-and-beyond" }
```
MarquessV, line 9: Here is the crux of my proposal: When `1455-python-support-for-program-and-beyond` is merged and we have a released version of `quil-py`, we update this spec to:

```toml
# The value for `tag` should correspond with the `quil` dependency specified 
# for `qcs-sdk-python` in `crates/python/pyproject.toml`.
# This ensures that `qcs-sdk`, `qcs-sdk-python`, and the version of `quil-py` it uses
# all share the same version of `quil-rs`
quil-rs = { git = "https://github.com/rigetti/quil-rs", tag = "quil-py/0.1.0" }
```

A tag gets published per crate (see this repo as an example), but since that tag is at the repo level, we can use it to specify a version of `quil-rs` to use. All we have to do is make sure that `qcs-sdk-python`s `pyproject.toml` is kept in sync. Depending on how fancy we want to get, we can check this during CI with a script. We can also have some `qcs-sdk-python` <-> `quil-py` integration tests in the test suite.

MarquessV: I see two primary advantages to this:

1) Maintainability

The re-exporting route requires some sort of build script to make sure that the type hints get copied over and packaged with `qcs-sdk-python` correctly. We would still have to have some way of keeping the versions of quil-rs in sync across each of the packages. 

2) Prevents users from using mismatched versions of `quil-py`

This could be a slight annoyance since it makes our dependency requirements a bit more strict, but it gives us the power and responsibility to say which versions of these two packages are compatible with each other, rather than putting it on the user. This would only really come up if you were already using `quil-py` before adding `qcs-sdk-python`, and in that case, a dependency mismatch error lets them know that they should update `quil-py` so they can properly use it in combination with `qcs-sdk-python`.

In the re-exporting case, it would be totally possible to have a version of `qcs-sdk-python` that re-exports, say `quil-py v0.5.0` and a separate `quil-py v0.2.0`, which could lead to unexpected behavior if they were to assume they could be used interchangeably. This proposal eliminates that possibility all together.

Shadow53: The downside I see with this is that crates.io does not allow uploads with `git` dependencies, so if we want the Rust part of the SDK on crates.io, we'll need to specify a version normally there

MarquessV: Ah yeah, that makes this less of an elegant solution then. If we have to manually specify version numbers, maybe `cargo-deny` can be our safeguard there, with the integration tests as backup.

Cargo.toml
```diff
@@ -6,10 +6,15 @@ qcs-api = "0.2.1"
...
-quil-rs = "0.16.0"
 serde_json = "1.0.86"
 thiserror = "1.0.37"
 tokio = "1.24.2"
+# We specify quil-rs as a git and versioned dependency so that we can keep the version of
+# quil-rs used in both the Rust and Python packages in sync. The tag used should always
+# be a `quil-py` tag and should match the version used in `crates/python/pyproject.toml`
+# The version must also be specified in order to publish to crates.io. Cargo enforces
+# that the specified version is the same as the version in the git repository.
+quil-rs = { version = "0.17.0", git = "https://github.com/rigetti/quil-rs", tag = "quil-py/v0.1.1" }
```
jselig-rigetti, line 17: Hmm, what's the issue with just specifying `0.17.0` if it's enforced that the tag's version has to match anyway?

crates/python/pyproject.toml
```diff
@@ -29,14 +29,21 @@ version = "0.8.0"
...
 
+[tool.maturin]
+features = ["pyo3/extension-module"]
+bindings = "pyo3"
+compatibility = "linux"
+sdist-include = ["README.md"]
+
 [tool.poetry.dependencies]
 python = "^3.8"
+quil = "0.1.0-rc.0"
```
jselig-rigetti, line 46: Technically the same code as is in [`0.1.1`](https://pypi.org/project/quil/0.1.1/) right?

jselig-rigetti (approved):

crates/python/pyproject.toml
MarquessV: yeah, but should've been 0.1.1

Cargo.toml
MarquessV: I might be misunderstanding the question, but cargo only checks if they match if both are specified. If we specified quil-rs by version only. It would be easier to update it to a version that didn't match the version of quil being used.

For example, In the future, if we release quil-rs `0.18.0` and quil-py `0.2.1`, then when I update the quil-rs version here, cargo will fail until I update the tag to match. If we only specified by version, cargo wouldn't complain at all.

Shadow53 (approved):
LGTM

PR #262, opened 2023-03-21
feat: support endpoint_id job target

jselig-rigetti:
This is *one* solution to allowing `endpoint_id` to be provided when submitting a job, presented as a draft for discussion.

To consider, there is benefit in separating translation from submission, but if we keep all that functionality on `Execution` it's not straightforward to me how `Execution` would cache translation output between `submit`, or how it would be made clear that `translate` should be called before `submit`.

### DIFF
--- a/crates/lib/src/qpu/client.rs
+++ b/crates/lib/src/qpu/client.rs
@@ -14,7 +14,9 @@ use qcs_api_client_grpc::{
 };
 use qcs_api_client_openapi::apis::{
     configuration::Configuration as OpenApiConfiguration,
-    endpoints_api::{get_default_endpoint, GetDefaultEndpointError},
+    endpoints_api::{
+        get_default_endpoint, get_endpoint, GetDefaultEndpointError, GetEndpointError,
+    },
     quantum_processors_api::{
         list_quantum_processor_accessors, ListQuantumProcessorAccessorsError,
     },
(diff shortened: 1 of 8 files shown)


MarquessV:
I think it makes sense to have each step of the process available as an individual step. What if `Execution` had a `pub(crate)` method for translation. From there it could manage a cache, fetching the result if it exists already, or requesting it from translation service if not. The `submit` method would just use the `translation` method so it would still be possible to `submit` without translating first, but would also leverage the cached result if it was available. We could then expose `translation` as an optional step to users via the `Executable`.

crates/lib/src/qpu/client.rs
```diff
@@ -122,6 +136,19 @@ impl Qcs {
             .await
     }
 
+    /// Get address for direction connection to Controller, when explicitly targeting a endpoint by ID.
```
Shadow53, line 139:
```suggestion
    /// Get address for direct connection to Controller, when explicitly targeting a endpoint by ID.
```

Shadow53 (commented):
I noticed `JobHandle` was changed from using `Cow<str>` to `String` (through `JobTarget`). Was that intentional?

Otherwise, the changes look good to me.

jselig-rigetti:
I think the suggestions here are good, but extend beyond the scope of this issue/MR. 

This is mainly about updating the "low-level" `qcs::qpu::api` methods -- the ones used by pyquil -- to support `endpoint_id`. Here `Executable` is minimally updated to support those changes.


---

## Reflecting on a redesign of `Executable`

> `translate_for_qpu(qpu_id) -> QPUExecutable`

@kalzoo Do you mean any of the following?
1. That should return (and we should make public) `pub(crate) qcs::qpu::Execution`
2. There should be a new `pub QPUExecutable` struct - I think this would end up looking a lot like (1) anyway, with only `submit(params, endpoint) -> JobHandle` and `retrieve(job_handle) -> Results` methods.
3. There should actually be no return, `translate_for_qpu(qpu_id) -> ()` should modify the internal state of `Executable` which should be updated to cache translated programs per QPU.

> I do think that the term "run" should be used for "translate + execute" rather than "execute", and "submit" should become "execute", to more closely track our API and convention in pyQuil.

Fair enough, but I don't think this MR strays from this being a future possibility.

> What if Execution had a pub(crate) method for translation

@MarquessV I actually added this as part of this MR, but regarding caching I see two paths: either return `EncryptedTranslationResult` to the user and require they pass it back with `submit`, or maintain an internal cache and add cache management methods and heuristics.

I think the benefits of lazy translation means we avoid freshness concerns and don't add mental overhead of managing the translation result to the user, who can't do much with the encrypted data anyway.

kalzoo:
> I think the suggestions here are good, but extend beyond the scope of this issue/MR.

Agreed - I meant to imply that with the comment but should have said that explicitly. This looks good and worthwhile as-is.

kalzoo (approved):

PR #264, opened 2023-03-22
feat: add tracing support

### LINKED ISSUE #263: Add tracing logs
Would be useful to have optional `tracing` log statements in key parts of the SDK, like whenever non-trivial work is being done.

Shadow53:
Resolves #263.

This adds tracing log statements in places where it seemed particularly useful to get a peek at what the SDK is doing.

This PR does not, at this time, include `TRACE` level statements or spans for every function/entrypoint. I wanted to open up discussion on whether we want that at this time. I see at least using `tracing::instrument` at the `TRACE` level on each function as potentially useful.

### DIFF
--- a/crates/lib/src/compiler/quilc.rs
+++ b/crates/lib/src/compiler/quilc.rs
@@ -33,12 +33,19 @@ pub const DEFAULT_COMPILER_TIMEOUT: f64 = 30.0;
 /// recoverable at runtime. This function can fail generally if the provided ISA cannot be converted
 /// into a form that `quilc` recognizes, if `quilc` cannot be contacted, or if the program cannot
 /// be converted by `quilc`.
+#[cfg_attr(
+    feature = "tracing",
+    tracing::instrument(skip(client), level = "trace")
+)]
 pub fn compile_program(
     quil: &str,
     isa: TargetDevice,
     client: &Qcs,
     options: CompilerOpts,
 ) -> Result<Program, Error> {
+    #[cfg(feature = "tracing")]
+    tracing::debug!(%quil, compiler_options=?options, "compiling quil program with quilc",);
+
     let config = client.get_config();
     let endpoint = config.quilc_url();
     let params = QuilcParams::new(quil, isa).with_protoquil(options.protoquil);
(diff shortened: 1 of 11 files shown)


crates/lib/src/compiler/quilc.rs
```diff
@@ -33,12 +33,19 @@ pub const DEFAULT_COMPILER_TIMEOUT: f64 = 30.0;
...
+    tracing::instrument(skip(client), level = "trace")
+)]
 pub fn compile_program(
     quil: &str,
     isa: TargetDevice,
     client: &Qcs,
     options: CompilerOpts,
 ) -> Result<Program, Error> {
+    #[cfg(feature = "tracing")]
+    tracing::debug!(%quil, compiler_options=?options, "compiling quil program with quilc",);
```
kalzoo, line 47: These programs can be massive - tens of thousands of lines long. Probably best not to log

crates/lib/src/qpu/rewrite_arithmetic.rs
```diff
@@ -50,6 +50,9 @@ use crate::executable::Parameters;
 /// distinct parameter from theta. Note that the value of `__SUBST[0]` will actually be
 /// `theta * 1.5 / 2π`.
 pub fn rewrite_arithmetic(program: Program) -> Result<(Program, Substitutions), Error> {
+    #[cfg(feature = "tracing")]
+    tracing::debug!("rewriting arithmetic");
```
kalzoo, line 54: Seems like `tracing::instrument` might be the right way to go for some of these, too? I'm not sure if I have a good rule of thumb for when we'd want: 

`tracing::instrument`

vs

`tracing::instrument(ret, ...)`

vs

event with `tracing::debug` etc

but a rough guess would be "call to external library or service, or prior to trying something likely to fail". However, even "likely to fail" could be `tracing::instrument(err, ...)`, right?

`tracing-error` has been _fantastic_ on other projects to surface spans only in the case of errors, but that might be a bit much to add here.

kalzoo (approved):
Seems great, and this is something we can easily experiment with and won't get locked into, so I'm happy to 🚢 and iterate as needed

MarquessV (commented):
I would echo the request to not log quil programs, but looking good otherwise.

crates/lib/src/qpu/rewrite_arithmetic.rs
Shadow53: Haven't used `tracing-error` before, so I'll at least look at it.

I don't really have a good rule of thumb on these either. In this case, rewriting arithmetic seemed like something worth surfacing as a logged event.

crates/lib/src/compiler/quilc.rs
Shadow53: Good point.

PR #266, opened 2023-03-23
feat!: support HTTPS_PROXY and HTTP_PROXY client proxies

### LINKED ISSUE #238: Ensure that system proxy use is supported
Some users of this library rely on proxies for access to QCS APIs. We should ensure that the following envvars are supported for all network calls made directly or indirectly by this library:

> Curl reads and understands the following environment variables:
> 
> http_proxy, HTTPS_PROXY, FTP_PROXY
> They should be set for protocol-specific proxies. General proxy should be set with
> 
> ALL_PROXY
> A comma-separated list of host names that should not go through any proxy is set in (only an asterisk, * matches all hosts)
> 
> NO_PROXY

(source: [curl docs](https://curl.se/docs/manual.html), see `Environment Variables` section)

jselig-rigetti:
Closes #238 

Route requests through client proxies via `HTTPS_PROXY` and `HTTP_PROXY` environment variables.

Supported protocols are `http`, `https`, and `socks5`.

### DIFF
--- a/crates/lib/src/qpu/client.rs
+++ b/crates/lib/src/qpu/client.rs
@@ -74,11 +74,10 @@ impl Qcs {
         quantum_processor_id: &str,
     ) -> Result<ControllerClient<RefreshService<Channel, ClientConfiguration>>, GrpcEndpointError>
     {
-        self.get_controller_endpoint(quantum_processor_id)
-            .await
-            .map(get_channel)
-            .map(|channel| wrap_channel_with(channel, self.get_config()))
-            .map(ControllerClient::new)
+        let uri = self.get_controller_endpoint(quantum_processor_id).await?;
+        let channel = get_channel(uri).map_err(|err| GrpcEndpointError::GrpcError(err.into()))?;
+        let service = wrap_channel_with(channel, self.get_config());
+        Ok(ControllerClient::new(service))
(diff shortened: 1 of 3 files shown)


crates/lib/src/qpu/client.rs
```diff
@@ -165,16 +164,16 @@ impl Qcs {
...
 
 /// Errors that may occur while trying to resolve a `gRPC` endpoint
 #[derive(Debug, thiserror::Error)]
 pub enum GrpcEndpointError {
-    /// Error due to a malformed URI
-    #[error("Malformed URI for endpoint: {0}")]
-    BadUri(#[from] GrpcError<RefreshError>),
+    /// Error due to a bad gRPC configuration
+    #[error("Error configuring gRPC request: {0}")]
+    GrpcError(#[from] GrpcError<RefreshError>),
```
kalzoo, line 176: FYI while this rename looks good, it is also a breaking change, make sure the commit message reflects that

kalzoo (approved):

crates/lib/src/qpu/client.rs
jselig-rigetti: Sure thing, changed commit to `feat!:`

PR #276, opened 2023-03-31
chore: update dependencies

Shadow53:
Updates dependencies to their newest versions.

A Rust test failed once but it wasn't clear why, so I added a message to print when the assertion fails, but the test succeeded when I ran it again 🤷🏻 

I was also getting warnings about `poetry.lock` being outdated, so I ran `poetry lock --no-update`, which looks like it just updated the hash of the Python library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,9 +15,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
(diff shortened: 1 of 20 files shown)


kalzoo (approved):
Looks good once CI is green

kalzoo (approved):
A little more than you bargained for at first! nicely done.

Issue #277, opened 2023-04-03
Tracing feature causes compilation failure


Shadow53:
A combination of merging #262 and #264 led to a state where enabling the `tracing` feature causes compilation to fail due to unknown variable names.

PR #278, opened 2023-04-03
fix: fix broken tracing after merge conflicts

Shadow53:
Quick fix for #277.

### DIFF
--- a/crates/lib/src/qpu/api.rs
+++ b/crates/lib/src/qpu/api.rs
@@ -48,6 +48,14 @@ pub enum JobTarget {
     EndpointId(String),
 }
 
+impl fmt::Display for JobTarget {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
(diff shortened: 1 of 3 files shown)


crates/lib/src/qpu/api.rs
```diff
@@ -129,11 +137,7 @@ pub async fn retrieve_results(
     client: &Qcs,
 ) -> Result<ControllerJobExecutionResult, GrpcClientError> {
     #[cfg(feature = "tracing")]
-    tracing::debug!(
-        "retrieving job results for {} on {}",
-        job_id,
-        quantum_processor_id
-    );
+    tracing::debug!("retrieving job results for {} on {}", job_id, job_target,);
```
kalzoo, line 140: Should we be using the fields here (`%job_id`) rather than formatting?

kalzoo (approved):

crates/lib/src/qpu/api.rs
Shadow53: Depends on what we want, really. Fields allow us to filter traces by those fields if we, say, output traces as JSON. Otherwise, I think this is more readable for humans.

PR #286, opened 2023-04-28
CI: Fix Release Pipelines

MarquessV:
There were two main issues at play here.

For the library crate, the `katyo/publish-crates` action was outdated and didn't support workspace dependencies. ~Fixed by updating the action, I also had to explicitly set the version of `qcs` in the Python crate to get past a cargo error.~ Turned out to be easier to replace the action with a simple call to `cargo publish`.

For both the library crate and Python package, `zmq` was failing to build on linux without the `-std=c++11` flag. Fixed by setting the `CXXFLAGS` env variable for those build steps.

Here is an example of both pipelines passing with said fixes:

* [Release Rust](https://github.com/rigetti/qcs-sdk-rust/actions/runs/4835045064)
* [Release Python](https://github.com/rigetti/qcs-sdk-rust/actions/runs/4834900147)

### DIFF
--- a/crates/python/Cargo.toml
+++ b/crates/python/Cargo.toml
@@ -16,7 +16,7 @@ name = "qcs_sdk"
 crate-type = ["cdylib"]
 
 [dependencies]
-qcs = { path = "../lib" }
+qcs = { path = "../lib", version = "0.10.0-rc.21" }
 qcs-api.workspace = true
 qcs-api-client-common.workspace = true
 qcs-api-client-grpc.workspace = true
--- a/crates/python/Cargo.toml
+++ b/crates/python/Cargo.toml
@@ -16,7 +16,7 @@ name = "qcs_sdk"
 crate-type = ["cdylib"]
 
 [dependencies]
-qcs = { path = "../lib", version = "*" }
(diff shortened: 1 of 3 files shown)


crates/python/Cargo.toml
```diff
@@ -16,7 +16,7 @@ name = "qcs_sdk"
 crate-type = ["cdylib"]
 
 [dependencies]
-qcs = { path = "../lib" }
+qcs = { path = "../lib", version = "0.10.0-rc.21" }
```
Shadow53, line 19: Can we make sure this version is updated when the library is updated? Alternatively, `version = "*"` work?

Shadow53 (approved):
Thanks for taking care of this!

crates/python/Cargo.toml
MarquessV: Was able to remove the version spec entirely after replacing the publish action with a simple call to `cargo publish`: https://github.com/rigetti/qcs-sdk-rust/actions/runs/4835196172/jobs/8617271158

Shadow53 (approved):
Looks much simpler now

PR #289, opened 2023-05-02
create qvm::api submodule

MarquessV:


crates/lib/src/qvm/api.rs
```diff
@@ -0,0 +1,141 @@
...
+    readouts: &[Cow<'_, str>],
+    params: &Parameters,
+    config: &ClientConfiguration,
+) -> Result<QvmResultData, Error> {
+    if shots == 0 {
+        return Err(Error::ShotsMustBePositive);
+    }
+
+    // Create a clone of the program so MOVE statements can be prepended to it
+    let mut program = program.clone();
```
MarquessV, line 50: This is a change in logic from before. This function took a `&mut Program`, prepended the move instructions then removed them again before returning.

My thinking is, for QVM runs at least, cloning the program makes the function signature more clear (in that the function wont change the program). It's also safer, as we can't accidentally modify the users Program without undoing the changes. Benchmark needed, but I would suspect for up to a certain size of program, working off the clone is faster, as you avoid resizing the vector for each instruction you need to peel off at the end (though that could be optimized with changes to quil-rs).

Thoughts?

crates/python/src/qvm/api.rs
```diff
@@ -0,0 +1,49 @@
...
+        let readouts = readouts.into_iter().map(|value| Cow::Owned(value)).collect::<Vec<Cow<'_, str>>>();
+        Ok(PyQvmResultData(qcs::qvm::api::run(&quil, shots, &readouts, &params, &config)
+            .await
+            .map_err(RustQvmError::from)
+            .map_err(RustQvmError::to_py_err)?))
+    }
+}
+
+py_function_sync_async! {
+    #[pyfunction(config = "None")]
```
jselig-rigetti, line 43:
```suggestion
    #[pyfunction(client = "None")]
```

crates/python/src/qvm/api.rs
```diff
@@ -0,0 +1,49 @@
...
+    funcs: [
+        py_get_version_info,
+        py_get_version_info_async,
+        py_run,
+        py_run_async
+    ],
+}
+
+py_function_sync_async! {
+    #[pyfunction(config = "None")]
```
jselig-rigetti, line 23: My guess is that this is automatic for `Optional<T>` types, but seems like a typo
```suggestion
    #[pyfunction(client = "None")]
```

crates/lib/src/qvm/api.rs
jselig-rigetti: Good move IMO.

I would be surprised if the change in execution time would be noticeable in context of the rest of QVM duration

crates/lib/src/qvm/api.rs
```diff
@@ -0,0 +1,141 @@
...
+                    });
+                }
+            }
+            None => {
+                return Err(Error::RegionNotFound { name: name.clone() });
+            }
+        }
+        for (index, value) in values.iter().enumerate() {
+            program.instructions.insert(
+                0,
```
jselig-rigetti, line 69: I guess it doesn't matter what order the `MOVE` instructions are in, as long as they are all prepended right?

jselig-rigetti (approved):
Overall LGTM 👍

crates/lib/src/qvm/api.rs
MarquessV: Yup, good case to call out, but order doesn't matter for this.

crates/lib/src/qvm/api.rs
```diff
@@ -0,0 +1,357 @@
...
+/// An enum encapsulating the different ways to request data back from the QVM for an address.
+#[derive(Serialize, Debug, Clone, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum AddressRequest {
+    /// Set to `true` to request all values at this address, or `false` to explicitly exclude
+    /// the address.
+    All(bool),
+    /// A list of specific indices to get back for the address.
+    Indices(Vec<usize>),
+}
```
MarquessV, line 108: Still workshopping the name of this enum and it's variants, suggestions welcome.

MarquessV: For more context, the QVM allows you to specify what memory regions you want back in the response with a map like:

```
{
    "ro": true,       # returns all values for "ro" 
    "foo": false,     # omit "foo" entirely
    "bar": [3, 5]     # return the values in bar[3] and bar[5]
}
```

crates/lib/src/qvm/mod.rs
```diff
@@ -32,73 +41,100 @@ impl QvmResultData {
...
-#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
-pub(super) struct Failure {
-    /// The message from QVM describing what went wrong.
-    pub(super) status: String,
+/// Run a Quil program on the QVM. The given [`Parameters`] are used to parameterize the value of
+/// memory locations across shots.
+pub async fn run(
+    quil: &str,
+    shots: u16,
+    addresses: HashMap<String, AddressRequest>,
```
MarquessV, line 49: I like the idea of keeping the `api` module consistent with the QVM API, and in these higher level functions, I'm abstracting away stuff if it isn't useful. Is this `addresses` parameter an example of that? pyQuil's `qc.run` method doesn't provide this option, it just assumes you want all data for all readout registers, which we can just as easily do here using `Program`. Probably going to opt for that here too, unless someone has a good reason to do it differently.

crates/lib/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![deny(clippy::cargo)]
 #![allow(clippy::multiple_crate_versions)] // This should be enforced by cargo-deny
 #![allow(clippy::missing_errors_doc)]
+#![allow(clippy::implicit_hasher)]
```
MarquessV, line 6: I needed to allow this for my changes, and I noticed we were already allowing elsewhere, so I allowed it globally. By default, [it is an allowed pedantic lint](https://rust-lang.github.io/rust-clippy/master/index.html), and the changes needed to satisfy it aren't worth the squeeze IMO, so I don't feel bad about it, but let me know if you feel otherwise.

crates/lib/src/qvm/api.rs
```diff
@@ -0,0 +1,376 @@
...
+
+/// A QVM response that can be deserialized into some successful response type `T`, or into a
+/// [`Failure`] containing an error message.
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+#[serde(untagged)]
+pub(super) enum QvmResponse<T>
+where
+    T: DeserializeOwned,
+{
+    #[serde(bound = "")]
```
MarquessV, line 37: This is needed for generic types that are only bounded by `DeserializeOwned` to be recognized as having `Deserialize`. See [serde#1296](https://github.com/serde-rs/serde/issues/1296).

crates/lib/src/qvm/mod.rs
```diff
@@ -32,73 +41,125 @@ impl QvmResultData {
...
+                Instruction::Move(Move {
+                    destination: ArithmeticOperand::MemoryReference(MemoryReference {
+                        name: name.to_string(),
+                        index: index as u64,
+                    }),
+                    source: ArithmeticOperand::LiteralReal(*value),
+                }),
+            );
         }
     }
```
Shadow53, line 127: Might be missing context here, but this looks like prepending instructions into a `Vec`. That's going to involve shifting all of the existing instructions every time.

The outer loop could probably also be rewritten in terms of iterators. The benefit would be checking all regions _before_ modifying the cloned `Program` and essentially short-circuiting. Something like:

(This might break upon applying, because I originally wrote the suggestion based on just the added lines, which had mismatched braces)

```suggestion
    program.instructions = params.into_iter().map(|(name, values)| {
        match program.memory_regions.get(name.as_ref()) {
            Some(region) => {
                if region.size.length == values.len() as u64 {
                    Ok((name, values))
                } else {
                    Err(Error::RegionSizeMismatch {
                        name,
                        declared: region.size.length,
                        parameters: values.len(),
                    })
                }
            }
            None => {
                Err(Error::RegionNotFound { name: name.clone() });
            }
        }
    }).collect::<Result<Vec<_>, _>>()?
        .into_iter()
        .rev()
        .flat_map(|(name, values)| {
            values.iter().enumerate().rev().map(|(index, value)| {
                Instruction::Move(Move {
                    destination: ArithmeticOperand::MemoryReference(MemoryReference {
                        name: name.to_string(),
                        index: index as u64,
                    }),
                    source: ArithmeticOperand::LiteralReal(*value),
                })
            })
        }).chain(program.instructions).collect();
```

- `rev()` because, with the current logic, the last item ends up first in the list. To keep the same ordering, we need to reverse both the outer and inner iterators
- We should only need to allocate memory once (per `collect`), since iterators allow providing size hints, and iterators over slices should provide those hints, so internally it should be able to use `Vec::with_capacity`.

crates/lib/src/qvm/mod.rs
```diff
@@ -32,73 +41,125 @@ impl QvmResultData {
...
+    #[cfg(feature = "tracing")]
+    tracing::debug!(
+        %shots,
+        ?addresses,
+        ?params,
+        "executing program on QVM"
+    );
+    if shots == 0 {
+        return Err(Error::ShotsMustBePositive);
+    }
```
Shadow53, line 95: https://doc.rust-lang.org/stable/std/num/struct.NonZeroU16.html

crates/lib/src/qvm/mod.rs
```diff
@@ -32,73 +41,125 @@ impl QvmResultData {
...
+#[allow(missing_docs)]
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Error parsing Quil program: {0}")]
+    Parsing(#[from] ProgramError<Program>),
+    #[error("Shots must be a positive integer.")]
+    ShotsMustBePositive,
+    #[error("Declared region {name} has size {declared} but parameters have size {parameters}.")]
+    RegionSizeMismatch {
+        name: Box<str>,
```
Shadow53, line 152: Why a `Box<str>`? Does that give us any benefit over a `String`?

crates/python/pyproject.toml
```diff
@@ -35,27 +35,33 @@ authors = [
...
 
+[tool.maturin]
+features = ["pyo3/extension-module"]
+bindings = "pyo3"
+compatibility = "linux"
+sdist-include = ["README.md"]
+
+[build-system]
+requires = ["maturin>=0.13,<0.14"]
+build-backend = "maturin"
```
Shadow53, line 57:
```suggestion
[build-system]
requires = ["maturin>=0.14,<0.15"]
build-backend = "maturin"
```

Shadow53 (dismissed):
Left some notes. A lot of comments were saved from when I started to review last week and might be outdated, though I tried to look for those.

Biggest thing is a spot where we are shifting elements in a `Vec` _a lot_, with a suggestion on how to rewrite it.

I did not go through the Python/PyO3 stuff as carefully as I would have liked, but nothing stood out there.

crates/lib/src/qvm/api.rs
```diff
@@ -0,0 +1,357 @@
...
+    Success(T),
+    Failure(Failure),
+}
+
+/// Fetch the version information from the running QVM server.
+pub async fn get_version_info(config: &ClientConfiguration) -> Result<String, Error> {
+    #[cfg(feature = "tracing")]
+    tracing::debug!("requesting qvm version information");
+    let client = reqwest::Client::new();
+    let params = vec![("type", "version")];
```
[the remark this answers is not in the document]
MarquessV: Had to switch to a HashMap to get it to serialize as key/value pairs. Turns out an array of tuples just serializes to a nested list.

crates/lib/src/qvm/mod.rs
MarquessV: None of this is new code, should have created an issue, but I went ahead and refactored the type used for shots to be `NonZeroU16`. Ended up impacting a lot of the crate.

crates/lib/src/qvm/mod.rs
MarquessV: I dunno, I didn't write it, just moved it out of the API module. If I had to guess, it's because the only place this error gets initialized, name comes from a map using `Box<str>`. I can change it if you'd like?

crates/lib/src/qvm/mod.rs
MarquessV: Okay, the refactor is done. Ended up with a similar approach, but opted to use a `try_for_each` to check the precondition since it felt more clear and was one less `collect`. Also, didn't use `rev()` since the ordering of the `MOVE` statements doesn't matter.

crates/python/qcs_sdk/qvm/api.pyi
```diff
@@ -36,17 +36,6 @@ class AddressRequest:
...
-    @staticmethod
-    def from_all(all: bool) -> AddressRequest: ...
-    @staticmethod
-    def from_indices(indices: Sequence[int]) -> AddressRequest: ...
-    def as_all(self) -> Optional[bool]: ...
-    def as_indices(self) -> Optional[Sequence[int]]: ...
-    def to_all(self) -> bool: ...
-    def to_indices(self) -> Sequence[int]: ...
-    def is_all(self) -> bool: ...
-    def is_indices(self) -> bool: ...
```
Shadow53: I would personally prefer having static methods `include_all`, `exclude_all`, and `from_indices` as constructors, rather than a `__new__` that takes a boolean or a list of integers. I'm fine with the rest of the methods not existing, though, if we don't expect users to inspect the `AddressRequest` after creating it.

crates/python/src/qvm/api.rs
```diff
@@ -157,6 +178,19 @@ impl PyMultishotMeasureRequest {
...
+
+    #[getter]
+    pub fn trials(&self) -> u16 {
+        self.as_inner().trials.get()
+    }
+
+    #[setter]
+    pub fn set_trials(&mut self, trials: u16) -> PyResult<()> {
+        // `NonZeroU16` doesn't implement `PyClass`, so `pyo3` doesn't allow it to be used
+        // as a method argument, even when combined with a `from_py_with` attribute.
```
Shadow53, line 190: Out of curiosity, do you feel using `NonZeroU16` has made things easier or harder in this PR?

Shadow53 (commented):
A couple of suggestions and a question.

crates/lib/src/qvm/api.rs
```diff
@@ -34,11 +34,26 @@ pub(super) enum QvmResponse<T>
...
+    // https://github.com/serde-rs/serde/issues/1296
     #[serde(bound = "")]
     Success(T),
     Failure(Failure),
 }
 
+impl<T: DeserializeOwned> QvmResponse<T> {
+    /// Converts a [`QvmResponse<T>`] into a [`Result<T, qvm::Error>`] containing the successful response,
+    /// as the Ok value or an [`Error`] containing the error message from the [`Failure`] response.
+    pub(super) fn ok(self) -> Result<T, Error> {
```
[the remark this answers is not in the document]
MarquessV: I was thinking you also use the word `ok` to go from Option -> Result, but I'm fine with `into_result` too, so I changed it.

crates/python/qcs_sdk/qvm/api.pyi
MarquessV: Okay, changed them to static methods.

crates/python/src/qvm/api.rs
MarquessV: Ha, I've been asking myself the same question. I mean, for this PR it definitely made things harder. Looking back, we should've created an issue for it and pressed on as I was just building on the existing API and this isn't exactly what we needed at the moment. I just thought it'd be simpler when I first got into it and didn't want to throw away the work once I got deep into it.

As far as using `NonZeroU16`, I'm torn. On one hand, it feels "correct" to be able to represent bounds as types. For this use case though, `NonZeroU16` added a lot of boiler plate and I'm not sure we gained a ton from it. We don't need to handle the error anymore, but now users have to import `NonZeroU16`, construct it, and handle the option just to use the function when they probably weren't going to consider passing `0` in to begin with, and the consequences of them doing so were handled with 3 lines of code. The pyo3 gymnastics needed was annoying, but I'm generally fine with that tradeoff if we're doing the right thing in the Rust crate. It's an interesting thing to reflect on though, because it makes you think about where the line should be drawn between types and implementation.

crates/python/src/from_py.rs
jselig-rigetti, line 1: Does it make sense to implement this in https://github.com/rigetti/rigetti-pyo3/blob/main/src/py_try_from.rs ?

crates/python/qcs_sdk/__init__.pyi
```diff
@@ -1,29 +1,357 @@
...
+@final
+class ExeParameter:
+    """
+    Program execution parameters.
+
+    Note: The validity of parameters is not checked until execution.
+    """
+
+    def __new__(
+        cls: type["ExeParameter"],
```
jselig-rigetti, line 146: I think the type annotation here can be omitted, depending on the language server - doesn't hurt I guess though

crates/python/qcs_sdk/__init__.pyi
```diff
@@ -1,29 +1,357 @@
...
+    ...
+
+@final
+class Executable:
+    """
+    The builder interface for executing Quil programs on QVMs and QPUs.
+    """
+
+    def __new__(
+        cls,
```
jselig-rigetti, line 26: Omitting the type here is fine, right? (referencing other comment in this file)

crates/python/qcs_sdk/compiler/quilc.pyi
```diff
@@ -148,10 +148,10 @@ class RandomizedBenchmarkingRequest:
     def qubits(self, value: int): ...
     @property
     def gateset(self) -> List[str]:
-        """List of Quil programs, each describing a Clifford."""
+        """Sequence of Quil programs, each describing a Clifford."""
```
jselig-rigetti, line 151: These are still technically returning a `List` though, right?

What's the motivation behind moving from `List` to `Sequence` in general though - are some callers using tuples/sets?

crates/python/qcs_sdk/qvm/api.pyi
```diff
@@ -0,0 +1,230 @@
+from typing import Dict, List, Sequence, Mapping, Optional, Tuple, final
+from typing_extensions import Self
+
+from qcs_sdk import RegisterData
+from qcs_sdk.client import QCSClient
+
+def get_version_info(client: Optional[QCSClient] = None) -> str:
```
jselig-rigetti, line 7: I'd been writing this as `client: Optional[QCSClient] = ...` trying to indicate that a client is actually constructed, not sure it's actually better but curious if you had a particular opinion on one vs the other

jselig-rigetti (approved):
Overall LGTM, a few nitpicks

crates/python/qcs_sdk/qvm/__init__.pyi
```diff
@@ -10,15 +13,72 @@ class QVMResultData:
...
+    gate_noise: Optional[Tuple[float, float, float]] = None,
+    rng_seed: Optional[int] = None,
+    client: Optional[QCSClient] = None,
+) -> QVMResultData:
+    """
+    Asynchronously runs the given program on the QVM.
+
+    :param quil: A quil program as a string.
+    :param shots: The number of times to run the program. Should be a value greater than zero.
+    :param readouts: A mapping of memory region names to an ``AddressRequest`` describing what data to get back for that memory region from the QVM at the end of execution.
```
jselig-rigetti, line 76:
```suggestion
    :param addresses: A mapping of memory region names to an ``AddressRequest`` describing what data to get back for that memory region from the QVM at the end of execution.
```

crates/python/qcs_sdk/qvm/__init__.pyi
```diff
@@ -10,15 +13,72 @@ class QVMResultData:
...
+    gate_noise: Optional[Tuple[float, float, float]] = None,
+    rng_seed: Optional[int] = None,
+    client: Optional[QCSClient] = None,
+) -> QVMResultData:
+    """
+    Runs the given program on the QVM.
+
+    :param quil: A quil program as a string.
+    :param shots: The number of times to run the program. Should be a value greater than zero.
+    :param readouts: A mapping of memory region names to an ``AddressRequest`` describing what data to get back for that memory region from the QVM at the end of execution.
```
jselig-rigetti, line 51: I think this was just misnamed? Also missing a few parameter docs

```suggestion
    :param addresses: A mapping of memory region names to an ``AddressRequest`` describing what data to get back for that memory region from the QVM at the end of execution.
```

crates/python/qcs_sdk/__init__.pyi
MarquessV: Yeah, omitting the type is fine and what I normally do, not sure why I specified it on the other one. Removed it.

crates/python/qcs_sdk/compiler/quilc.pyi
MarquessV: Oh yeah, that should be `List`, I think a heavy handed find and replace changed the docstring here, switched it back.

My general approach is to accept interfaces/abstract types like `Sequence` or `Mapping` as input arguments and return concrete types like `List` or `Dict`.

As input types, abstract types like `Sequence` and `Mapping`s are more flexible since they define the minimum required logic needed. As you mention, a `Sequence` allows for list, tuples, and sets and per the pyo3 docs, it will convert `Sequence`s to `Vec`s and `Mapping` to `HashMap`s, so we aren't misrepresenting the input types we can accept. The `Sequence` and `Mapping` types are also immutable, so we're expressing that we won't modify the input arguments via the type as well. In practice, I have found this flexibility pretty helpful, since there are places throughout pyQuil where we use list and tuples interchangeably.

The idea behind returning concrete types has an inverse justification, they have at least parity with their corresponding abstract type, but often define a superset of functionality, so when I get one of those types back, I know exactly what I'm getting and what I can do with it.

crates/python/qcs_sdk/qvm/api.pyi
MarquessV: My only qualm with using `...` is that it doesn't faithfully represent the actual default value. `...` is its own type in Python, and if someone were to explicitly call `get_version_info(client=...)`, they would get an error.

MarquessV: I've gone ahead and replaced all the `...` with `None` just so they are consistent, but happy to revisit the topic if needed.

crates/python/src/from_py.rs
MarquessV: Created a follow-up issue [here](https://github.com/rigetti/rigetti-pyo3/issues/26)

PR #291, opened 2023-05-04
Return native quil metadata on compilation.

### LINKED ISSUE #290: Forward `native_quil_metadata` from `quilc`
When compiling a program to native quil with `quilc`, the response can optionally contain metadata with helpful details about the result of the compilation. We only return the compiled program right now, but we should also return this metadata.

MarquessV:
Closes #290


crates/python/qcs_sdk/compiler/quilc.pyi
```diff
@@ -233,6 +211,55 @@ async def compile_program_async(
...
 
+@final
+class CompilationResult:
+    @property
+    def program(self) -> str:
+        """ """
+        ...
+    @property
+    def native_quil_metadata(self) -> Optional[NativeQuilMetadata]:
+        """ """
```
jselig-rigetti, line 222: might as well omit the empty comments?

crates/python/qcs_sdk/compiler/quilc.pyi
```diff
@@ -233,6 +211,55 @@ async def compile_program_async(
...
+        ...
+
+class NativeQuilMetadata:
```
jselig-rigetti, line 225: Not sure if this is needed - if the type checker works then /shrug

```suggestion
@final
class NativeQuilMetadata:
```

jselig-rigetti (approved):
Overall LGMT - this is a breaking change right?

crates/python/qcs_sdk/compiler/quilc.pyi
MarquessV: Oops, yeah, those should be filled in, fixed.

crates/python/qcs_sdk/compiler/quilc.pyi
MarquessV: Ah yeah, it is. Helpful hint to have since the error you get when trying to use it as a base class is a bit obtuse.

PR #293, opened 2023-05-05
fix: Deserialize final_rewiring_values from None as an empty Vec

MarquessV:
The initial implementation of this PR assumed that `final_rewiring_values` would always be a list of values, but quilc can return also `None`. This handles those cases by de-serializing `None` as an empty vec.

This also adds a basic constructor and support for pickling `NativeQuilMetadata` so that it can be used with Python's `copy.deepcopy`.


crates/lib/src/compiler/quilc.rs
```diff
@@ -307,13 +307,26 @@ struct QuilToNativeQuilResponse {
...
+    let opt = Option::deserialize(deserializer)?;
+    Ok(opt.unwrap_or_default())
+}
+
 /// Metadata about a program compiled to native quil.
 #[derive(Clone, Deserialize, Debug, PartialEq, PartialOrd)]
 pub struct NativeQuilMetadata {
     /// Output qubit index relabeling due to SWAP insertion.
+    // #[serde(deserialize_with = "deserialize_optional_unit_or_vec_as_vec")]
+    #[serde(deserialize_with = "deserialize_none_as_default")]
```
MarquessV, line 329: We could also change the type to `Option<Vec<u64>>`, but it would would make the API annoyingly clunky to use. If I understand correctly, `quilc` only returns `None` here if no qubits were used in the program. Most users are going to care about programs that do use qubits, and it would be annoying to have to check for `None` each time. I assume that's why [this logic exists in the rpcq package](https://github.com/rigetti/rpcq/blob/master/rpcq/messages.py#L237) as well.

crates/lib/src/compiler/quilc.rs
```diff
@@ -307,13 +307,25 @@ struct QuilToNativeQuilResponse {
...
+{
+    let opt = Option::deserialize(deserializer)?;
+    Ok(opt.unwrap_or_default())
+}
+
 /// Metadata about a program compiled to native quil.
 #[derive(Clone, Deserialize, Debug, PartialEq, PartialOrd)]
 pub struct NativeQuilMetadata {
     /// Output qubit index relabeling due to SWAP insertion.
+    #[serde(deserialize_with = "deserialize_none_as_default")]
```
MarquessV, line 328: We could also change the type to `Option<Vec<u64>>`, but it would would make the API annoyingly clunky to use. If I understand correctly, `quilc` only returns `None` here if no qubits were used in the program. Most users are going to care about programs that do use qubits, and it would be annoying to have to check for `None` each time. I assume that's why [this logic exists in the rpcq package](https://github.com/rigetti/rpcq/blob/master/rpcq/messages.py#L237) as well.

crates/lib/Cargo.toml
```diff
@@ -45,7 +45,6 @@ quil-rs.workspace = true
 reqwest = { version = "0.11.12", default-features = false, features = ["rustls-tls", "json"] }
 rmp-serde = "1.1.1"
 serde = { version = "1.0.145", features = ["derive"] }
-serde_bytes = "0.11.7"
```
MarquessV, line 48: Noticed this wasn't being used anymore, so I removed it.

crates/lib/src/compiler/quilc.rs
```diff
@@ -307,13 +307,25 @@ struct QuilToNativeQuilResponse {
...
+#[allow(unused_qualifications)]
+fn deserialize_none_as_default<'de, D, T>(deserializer: D) -> Result<T, D::Error>
+where
+    D: Deserializer<'de>,
+    T: Deserialize<'de> + std::default::Default,
+{
+    let opt = Option::deserialize(deserializer)?;
+    Ok(opt.unwrap_or_default())
+}
+
```
Shadow53, line 323: Shouldn't need this. I think you just need to do `serde(default)` and it will automatically use the default.

Shadow53 (dismissed):

PR #297, opened 2023-05-11
feat: the default QCS Client configuration now respects environment variable overrides

MarquessV:
As of `qcs-api-client-common` `0.6.6`, `ClientConfiguration` respects the value of environment variable overrides as the default value. Including these changes here allows users to initialize a default `QCSClient` that doesn't require a full blown QCS configuration but still uses overrides from the environment. This is useful for users who just want to use `quilc` and/or the `qvm`. They may not have a QCS config, but they still may set the `QCS_APPLICATIONS_QVM_URL` or `QCS_APPLICATIONS_QUILC_URL` environment variables and expect them to work.

### DIFF
--- a/crates/lib/src/qpu/client.rs
+++ b/crates/lib/src/qpu/client.rs
@@ -25,6 +25,9 @@ use qcs_api_client_openapi::models::QuantumProcessorAccessorType;
 use tonic::transport::{Channel, Uri};
 use tonic::Status;
 
+#[cfg(feature = "tracing")]
+use tracing;
+
 pub use qcs_api_client_common::configuration::LoadError;
 pub use qcs_api_client_grpc::channel::Error as GrpcError;
 pub use qcs_api_client_openapi::apis::Error as OpenApiError;
@@ -46,6 +49,13 @@ impl Qcs {
     pub async fn load() -> Result<Self, LoadError> {
         ClientConfiguration::load_default()
             .await
+            .or_else(|_| {
+                #[cfg(feature = "tracing")]
+                tracing::info!(
+                    "No QCS client configuration found, only generic QVMs will be accessible."
+                );
+                Ok(ClientConfiguration::default())
+            })
             .map(Self::with_config)
     }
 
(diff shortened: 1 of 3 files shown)


crates/lib/src/qpu/client.rs
```diff
@@ -46,6 +49,13 @@ impl Qcs {
     pub async fn load() -> Result<Self, LoadError> {
         ClientConfiguration::load_default()
             .await
+            .or_else(|_| {
+                #[cfg(feature = "tracing")]
+                tracing::info!(
+                    "No QCS client configuration found, only generic QVMs will be accessible."
+                );
+                Ok(ClientConfiguration::default())
+            })
```
MarquessV, line 58: I initially added this to `get_qc` in pyQuil specifically, but it's probably better to do it here, since this library can be used without a full blown QCS config as well.

MarquessV: A separate issue is how this client is technically buried in the `qpu` module, despite being used by the Python bindings for everything. As a follow-up, it probably makes sense to lift this towards the root module and use it as the client for everything in the Rust crate as well.

crates/lib/src/qpu/client.rs
```diff
@@ -25,6 +25,9 @@ use qcs_api_client_openapi::models::QuantumProcessorAccessorType;
 use tonic::transport::{Channel, Uri};
 use tonic::Status;
 
+#[cfg(feature = "tracing")]
+use tracing;
+
```
kalzoo, line 30: This isn't needed

crates/lib/src/qpu/client.rs
kalzoo: Agreed

crates/lib/src/qpu/client.rs
```diff
@@ -46,6 +49,13 @@ impl Qcs {
     pub async fn load() -> Result<Self, LoadError> {
         ClientConfiguration::load_default()
             .await
+            .or_else(|_| {
+                #[cfg(feature = "tracing")]
+                tracing::info!(
+                    "No QCS client configuration found, only generic QVMs will be accessible."
```
kalzoo, line 55:
```suggestion
                    "No QCS client configuration found. QPU data and QCS will be inaccessible and only generic QVMs will be available for execution"
```

kalzoo (approved):

crates/lib/src/qpu/client.rs
MarquessV: Resolving this with a [new issue](https://github.com/rigetti/qcs-sdk-rust/issues/298)

Issue #283, opened 2023-04-19
Use `py.check_signals()` in potentially long running calls


MarquessV:
It's not uncommon for users to want to prematurely terminate a running program by pressing `Ctrl-C`. When this happens, Python sets a flag to act on the `SIGINT` signal the next time it gets a chance. When Python hands off control to these Rust bindings, it can't process these flags, which makes the program unresponsive to keyboard interrupts until control is handed back to Python. Since many of the calls to `qcs-sdk-python` are potentially long running API calls, we should make an effort to check on signals periodically and terminate early when we get one, that way `Ctrl-C` works as expected.

Fortunately, the `pyo3::Py` struct has a `check_signals()` method that will return an error when it receives something like a keyboard interrupt. We should find a way to use this in any function that can potentially run long to break and return early so Python can finish terminating the program.

* [Relevant pyo3 faq](https://pyo3.rs/main/faq#ctrl-c-doesnt-do-anything-while-my-rust-code-is-executing)
* [Check signals api docs](https://docs.rs/pyo3/latest/pyo3/marker/struct.Python.html#method.check_signals)
* This [note](https://docs.rs/pyo3/latest/pyo3/marker/struct.Python.html#note) calls out that user-defined signal handlers can be called inside of `check_signals`, which means that arbitrary python code can be run during this call. I don't think that poses a security risk for our use case, but worth thinking about as we build this feature out.

PR #299, opened 2023-05-19
fix(python): check signals before blocking thread

jselig-rigetti:
Fixes https://github.com/rigetti/qcs-sdk-rust/issues/283

Instead of blocking on a spawned thread, continually check the thread status and `py.check_signals`. Only block the main thread when the spawned thread is ready.

The async counterpart functions already support `Ctrl-C`.

To test this, run the following without having quil-c running (so that requests hang):
```sh
cd crates/python
poetry shell
maturin develop
python -m asyncio
```
and then in the python shell, try both of these
```py
from qcs_sdk.compiler.quilc import get_version_info, get_version_info_async

get_version_info()
await get_version_info_async()
```

Originally this was paired with an update to make the zmq client async, but that isn't actually necessary.

### DIFF
--- a/crates/python/src/py_sync.rs
+++ b/crates/python/src/py_sync.rs
@@ -22,11 +22,15 @@ macro_rules! py_sync {
     ($py: ident, $body: expr) => {{
         let runtime = ::pyo3_asyncio::tokio::get_runtime();
         let handle = runtime.spawn($body);
+
+        // A 100ms loop delay is a bit arbitrary, but seems to
+        // balance CPU usage and SIGINT responsiveness well enough.
         let delay = ::std::time::Duration::from_millis(100);
         while !handle.is_finished() {
             $py.check_signals()?;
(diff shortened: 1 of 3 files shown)


crates/python/src/py_sync.rs
```diff
@@ -19,9 +19,14 @@
 /// assert say_hello("Rigetti") == "hello Rigetti"
 /// ```
 macro_rules! py_sync {
-    ($body: expr) => {{
+    ($py: ident, $body: expr) => {{
         let runtime = ::pyo3_asyncio::tokio::get_runtime();
         let handle = runtime.spawn($body);
+        let delay = ::std::time::Duration::from_millis(100);
```
MarquessV, line 25: Any particular methodology that led to landing on 100ms here? It's fine if not, but I'm curious in what a good value for this is, or if it is really needed.

MarquessV (approved):
Awesome!

PR #302, opened 2023-05-25
feat!: Loading a Qcs Client is now infallible and is the primary client used throughout the library.

MarquessV:
closes #297

### DIFF
--- a/crates/lib/src/client.rs
+++ b/crates/lib/src/client.rs
@@ -79,10 +79,10 @@ impl Qcs {
         self
     }
 
-    /// Return a copy of all settings parsed and resolved from configuration sources.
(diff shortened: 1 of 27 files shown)


Shadow53 (commented):
I'm not a fan of how the client silently (without error, not including debug printing) falls back to creating a _QCS_Client without QCS credentials. If someone is using this non-interactively, the error of invalid QCS config is now moved from loading the client to some issue authenticating with the service. (It looks like this fallback logic was previously in place and needlessly wrapped in `Ok`, so I'm okay making this a separate issue)

At the least, I would prefer the fallback behavior to be opt-in, probably as a separate method. We should arguably have `QCSClient` and `QVMClient` as two separate types, to avoid confusion.

Shadow53 (approved):
I'm happy with having an issue tracking the default behavior. Looks good otherwise.

PR #305, opened 2023-05-31
fix(python): The translate function no longer hangs infinitely

MarquessV:
This fixes an issue were certain synchronous variants of functions exported by `py_function_sync_async!` were hanging indefinitely.

After some experimenting, I narrowed down the issue to `std::thread::sleep`. It doesn't appear to consistently yield CPU time to to other threads.

This naturally led me to `tokio::time::sleep`, which is async and needs to be awaited. I wrapped everything up in a `tokio::select!` statement, which I was worried about as first, since `py.check_signals` needs to run on the main thread to work. However, [the docs state](https://docs.rs/tokio/latest/tokio/macro.select.html#runtime-characteristics) that `select!` runs tasks concurrently on the same thread, and as long as a task isn't `spawn`ed, it will get scheduled on the main thread.

### DIFF
--- a/crates/python/src/py_sync.rs
+++ b/crates/python/src/py_sync.rs
@@ -21,19 +21,21 @@
 macro_rules! py_sync {
     ($py: ident, $body: expr) => {{
         let runtime = ::pyo3_asyncio::tokio::get_runtime();
-        let handle = runtime.spawn($body);
 
-        // A 100ms loop delay is a bit arbitrary, but seems to
-        // balance CPU usage and SIGINT responsiveness well enough.
-        let delay = ::std::time::Duration::from_millis(100);
-        while !handle.is_finished() {
-            $py.check_signals()?;
-            ::std::thread::sleep(delay);
-        }
-
-        runtime
-            .block_on(handle)
-            .map_err(|err| ::pyo3::exceptions::PyRuntimeError::new_err(err.to_string()))?
+        runtime.block_on(async {
+            tokio::select! {
+                result = $body => result,
+                signal_err = async {
+                    // A 100ms loop delay is a bit arbitrary, but seems to
+                    // balance CPU usage and SIGINT responsiveness well enough.
+                    let delay = ::std::time::Duration::from_millis(100);
+                    loop {
+                        $py.check_signals()?;
+                        tokio::time::sleep(delay).await;
+                    }
+                } => signal_err,
+            }
+        })
     }};
 }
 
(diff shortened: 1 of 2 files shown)


jselig-rigetti:
Hmm, `ctrl+c` isn't working for me anymore. The following is hanging again on this branch, but works on `main`:

```sh
cd crates/python
poetry shell
maturin develop
python

>>> from qcs_sdk.compiler.quilc import get_version_info
>>> get_version_info()
^C
```

jselig-rigetti (approved):

crates/python/src/py_sync.rs
```diff
@@ -50,6 +53,12 @@ macro_rules! py_async {
 /// named after it that can be used to invoke either
 /// blocking or async variants of the same function.
 ///
+/// The given function will be spawned on a Rust event loop
+/// this means functions like [`pyo3::Python::with_gil`]
+/// should not be used, as acquiring Python's global
+/// interpreter lock from a Rust runtime
+/// isn't possible.
+///
```
MarquessV, line 61: I don't love that this is only "enforced" by a comment, interested in other ways of handling it.

crates/python/src/qpu/translation.rs
```diff
@@ -115,9 +112,7 @@ py_function_sync_async! {
         translation_options: Option<PyTranslationOptions>,
     ) -> PyResult<PyTranslationResult> {
         let client = PyQcsClient::get_or_create_client(client).await;
-        let translation_options = Python::with_gil(|py| {
-            Option::<TranslationOptions>::py_try_from(py, &translation_options)
-        })?;
+        let translation_options = translation_options.map(|opts| opts.as_inner().clone());
```
MarquessV, line 115: In many cases, including this one, `PyTryFrom` clones the underlying data just like this, so we're not performing any extra work here.

crates/python/src/py_sync.rs
jselig-rigetti: Ideally it'd be passed through to the inner function - I wonder if https://docs.rs/pyo3/latest/pyo3/marker/struct.Python.html#method.assume_gil_acquired can still be used instead, if only in dire circumstances. Since it's not something that can be enforced statically, I think a warning comment is about as much as you can do.

MarquessV: Yeah, I confirmed that the inner function can't take it as an argument because it's async. I'll keep it in mind, but since this side effect isn't new to this PR, I'll go ahead and merge for now.

PR #307, opened 2023-06-06
feat(python): support TranslationOptions

### LINKED ISSUE #281: Support translation backend options

### LINKED ISSUE #306: Missing Python bindings for the `grpc` submodule
`crates/python/grpc` has useful structs and wrappers, but it is not built into the Python library

kalzoo:
Closes #306
Closes #281 

* Adds Python type stubs
* Fixes outdated python type stub for `translate`
* Builds the `grpc` module into Python bindings

TODO:

- [ ] `grpc/controller`. Will add that once I know these changes are on the right track

### DIFF
--- a/crates/python/qcs_sdk/grpc/models/translation.pyi
+++ b/crates/python/qcs_sdk/grpc/models/translation.pyi
@@ -1,0 +1,45 @@
+from typing import Optional, final
+
+class TranslationOptions:
+    translation_backend: Optional[TranslationBackend] = None
+
+    ...
+
+@final
+class TranslationBackend:
+    """
+    Object specifying which translation backend to use to translate a particular program,
+    and for that given backend, what options to apply.
+
+    Variants:
+        ``v1``: Corresponds to the V1 translation backend.
+        ``v2``: Corresponds to the V2 translation backend.
+
+    Methods (each per variant):
+        - ``is_*``: if the underlying values are that type.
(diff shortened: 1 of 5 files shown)


crates/python/src/lib.rs
```diff
@@ -36,6 +36,7 @@ create_init_submodule! {
     submodules: [
         "client": client::init_submodule,
         "compiler": compiler::init_submodule,
+        "grpc": grpc::init_submodule,
```
MarquessV, line 39: I know you are just carrying forward with an existing module and just actually exporting it here, and I know the plan is to add a wrapped type later, but I don't love exporting raw bindings to the API here, and especially exposing them as a "GRPC" module to users.

Other modules get coupled to it because by it's very nature it contains models that requests across the stack need, but the individual models are typically scoped to a single kind of request, so why have them all co-located? It's also ambiguous when you need to pull it in, since our services use a mix of REST and GRPC.

To me, it feels like we're exposing unnecessary implementation details to users and making it more complex than needed to reason about our API and how to use it. `translation` is a clean concept within the scope of our product but `grpc` is ambiguous and doesn't help users discover useful things about our API. 

I could be convinced to re-export the needed GRPC model directly from the translation module, that way it is at least grouped within a concept, reducing the number of imports needed and making it easier to find the needed type with tooling.

PR #308, opened 2023-06-06
feat!: TranslationOptions builder-like struct with Python bindings

kalzoo:
Breaking change: it modifies the signature of `qpu::translate`.

This is an alternative approach to #307 , following recommendations made by @MarquessV . It supersedes that MR if accepted (it's not a merge train). It _partially_ decouples this library from the QCS API's generated gRPC code.

This is partial because it still allows the library consumer to set translation options using the gRPC generated structs. However, it now defines a `TranslationOptions` struct of its own, and a conversion from that struct to the gRPC generated struct. This is a middle ground between the options of (a) entirely decoupling from the gRPC generated code or (b) only accepting the gRPC generated struct, as `main` currently does.

The benefit is flexibility. Consider the alternatives:

* If `translate` in this library were to only accept the `TranslationOptions` generated from the protobuf schema (as it currently does on `main`), then consumers would have to use that struct directly. It's built from a wire format and thus is not the most intuitive to understand and use. As it stands today, the struct is rather simple and so would not cause too much trouble; however, `oneof`s and optional fields can be challenging and verbose to manipulate, especially when nested.
* If `translate` only accepted the new, hand-written struct defined here, then it would be a _frequent_ occurrence for users to be unable to access new features in the API because the struct is out of date.

Instead of those, `translate` now takes `translation_options: Option<impl Into<grpc::TranslationOptions>>` so the user can take the path that works best for them. They can provide the less-ergonomic but possibly more-complete grpc generated struct, or the locally-defined, hand-written options struct, or anything else that implements `Into` for the grpc struct.

Two downsides/considerations:

* we still couple the API to the gRPC generated structs. I think that's alright, given that a breaking change in that protobuf spec would roll down into breaking changes and chaos here anyway
* This generic behavior is not directly extensible through our Python bindings, we'd need to support this polymorphism by hand. Read: python users will _have_ to wait for client support of API features and will be forced to upgrade their client in order to access new fields in this options message.


MarquessV:
One very minor thing - `qcs_sdk/qpu/translation.pyi:4` is still importing `TranslationOptions` from the old path.


Otherwise, this looks good! I like the flexibility of the generic on the Rust side, even if it doesn't realize itself on the Python side as easily. I think it is a win regardless, but just curious about this:

> If translate only accepted the new, hand-written struct defined here, then it would be a frequent occurrence for users to be unable to access new features in the API because the struct is out of date.

Does the generic really solve that problem? The fastest way for a user to get access to new fields in the prost generated `TranslationOptions` would be to update their version of the GRPC client, but then that version of the struct wouldn't be compatible with the QCS SDK until we updated it, right? I could be mistaken about how Rust would handle that though.

Either way, happy with this, and thanks for taking on the side quest!

MarquessV (approved):

kalzoo:
> One very minor thing - `qcs_sdk/qpu/translation.pyi:4` is still importing `TranslationOptions` from the old path.
> 
> Otherwise, this looks good! I like the flexibility of the generic on the Rust side, even if it doesn't realize itself on the Python side as easily. I think it is a win regardless, but just curious about this:
> 
> > If translate only accepted the new, hand-written struct defined here, then it would be a frequent occurrence for users to be unable to access new features in the API because the struct is out of date.
> 
> Does the generic really solve that problem? The fastest way for a user to get access to new fields in the prost generated `TranslationOptions` would be to update their version of the GRPC client, but then that version of the struct wouldn't be compatible with the QCS SDK until we updated it, right? I could be mistaken about how Rust would handle that though.
> 
> Either way, happy with this, and thanks for taking on the side quest!

Good spot, fixed - thank you.

As far as the ability to upgrade, it would still be compatible if it were within the version bounds in Cargo.toml. It may be the case that the grpc package isn't so much out of date and needs the user to upgrade it, but that not all options have been copied over into the hand-written struct/builder yet. From past experience, I'd say that the builder will probably only get updated by request or when those features are recognized as missing, so days might pass between (user wanting to use a new feature) and (that feature is actually available). The `Into` approach lets us offer them an escape hatch, on the Rust side at least.

PR #309, opened 2023-06-08
feat!: QVM timeout can now be configured via a new options parameter

MarquessV:
There wasn't a good way to set a timeout on QVM requests, so I created an options struct, similar to how we handle it for quilc.

### DIFF
--- a/crates/lib/src/qvm/mod.rs
+++ b/crates/lib/src/qvm/mod.rs
@@ -20,7 +20,7 @@ pub mod api;
 mod execution;
 
 /// Number of seconds to wait before timing out.
-const DEFAULT_QVM_TIMEOUT: f32 = 30.0;
+const DEFAULT_QVM_TIMEOUT: Duration = Duration::from_secs(30);
 
 /// Encapsulates data returned after running a program on the QVM
 #[allow(clippy::module_name_repetitions)]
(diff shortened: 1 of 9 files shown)


crates/lib/src/qvm/mod.rs
```diff
@@ -147,6 +153,33 @@ pub fn apply_parameters_to_program(
...
+    #[must_use]
+    pub fn new() -> Self {
+        Self { timeout: None }
+    }
+}
+
+impl Default for QvmOptions {
+    fn default() -> Self {
+        Self {
+            timeout: Some(Duration::from_secs_f32(DEFAULT_QVM_TIMEOUT)),
```
Shadow53, line 178: Goes with the above suggestion

```suggestion
            timeout: Some(DEFAULT_QVM_TIMEOUT),
```

crates/python/qcs_sdk/qvm/__init__.pyi
```diff
@@ -25,6 +25,30 @@ class QVMResultData:
         """
         ...
 
+@final
+class QVMOptions:
```
Shadow53, line 29: Other docs suggest a `.default()` method exists on this class, but does not appear here.

Shadow53 (changes requested):

PR #311, opened 2023-06-21
feat(python): Integrate with logging

MarquessV:
More detailed logs can be retrieved by setting the log level to `INFO` or lower.

This is implemented via `pyo3_log`, which integrates with Rust's `log`. On the Rust side, we enable the `log` feature of `tracing` so it also emits its logs to `log`. 

logging3

### DIFF
--- a/crates/python/README.md
+++ b/crates/python/README.md
@@ -7,3 +7,58 @@ to compile and run Quil programs on Rigetti quantum processors. Internally, it i
 
 While this package can be used directly, [pyQuil](https://pypi.org/project/pyquil/) offers more functionality and a 
 higher-level interface for building and executing Quil programs. This package is still in early development and breaking changes should be expected between minor versions.
+
+## Troubleshooting
+
+### Enabling Debug logging
+
+This package integrates with Python's [logging facility](https://docs.python.org/3/library/logging.html) through a Rust crate called [`pyo3_log`](https://docs.rs/pyo3-log/latest/pyo3_log/). The quickest way to get started is to just enable debug logging:
+
+```python
+import logging
+logging.basicConfig(level=logging.DEBUG)
+```
+
+Because this is implemented with Rust, there are some important differences in regards to log levels and filtering.
+
+#### The `TRACE` log level
+
+Rust has a `TRACE` log level that doesn't exist in Python. It is less severe than `DEBUG` and is set to a value of 5. While the `DEBUG` level is recommended for troubleshooting, you can choose to target `TRACE` level logs and below like so:
+
+```python
+import logging
+logging.basicConfig(level=5)
+```
+
+#### Runtime Configuration and Caching
+ 
+`pyo3_log` caches loggers and their level filters to improve performance. This means that logger re-configuration done at runtime may cause logs to log or not log as expected in certain situations. If this is a concern, [this section of the pyo3_log documentation](https://docs.rs/pyo3-log/latest/pyo3_log/#performance-filtering-and-caching) goes into more detail.
+
+#### Filtering Logs
+
+Because the logs are emitted from a Rust library, the logger names will correspond to the fully qualified path of the Rust module in the library where the log occured. These fully qualified paths all have their own logger, and have to be configured individually.
+
+For example, say you wanted to disable the following log:
+
+```
+DEBUG:hyper.proto.h1.io:flushed 124 bytes
+```
+
+You could get the logger for `hyper.proto.h1.io` and disable it like so:
+
+```python
+logging.getLogger("hyper.proto.h1.io").disabled = True
+```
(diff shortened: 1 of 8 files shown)


crates/python/src/py_sync.rs
```diff
@@ -20,22 +20,26 @@
 /// ```
 macro_rules! py_sync {
     ($py: ident, $body: expr) => {{
-        let runtime = ::pyo3_asyncio::tokio::get_runtime();
-        let handle = runtime.spawn($body);
+        $py.allow_threads(|| {
```
MarquessV, line 23: Per the [`pyo3_log` documentation](https://docs.rs/pyo3-log/latest/pyo3_log/#interaction-with-python-gil), we need to wrap this function with `allow_threads` to avoid deadlocks. This necessitates using `Python::with_gil` to check the signals. In my testing so far, that's worked as expected.

crates/python/src/lib.rs
```diff
@@ -43,5 +43,6 @@ create_init_submodule! {
 
 #[pymodule]
 fn qcs_sdk(py: Python<'_>, m: &PyModule) -> PyResult<()> {
+    pyo3_log::init();
```
kalzoo, line 46: Should this be `try_init` with a warning message on failure? Seems like this is generally going to succeed, but it's be best to not brick the package when it doesn't for whatever reason.

kalzoo (commented):
It's nice when it's easy!

Let's take the time here to add some notes to the docs - the `pyo3_log` docs mention a few sharp edges that could come as a surprise to python users in terms of filtering and configuration. Just a brief summary with a link to the `pyo3_log` docs would help address some of that.

Also, some usage examples/tips, like how to filter out all of the hyper messages and show only qcs messages, and how to change the level, esp to the harder-to-discover `trace`:

>The [Trace](https://docs.rs/log/0.4.18/x86_64-unknown-linux-gnu/log/enum.Level.html#variant.Trace) doesn’t exist on the Python side, but is mapped to a level with value 5

MarquessV:
@kalzoo I added a section to the README with examples on filtering. Let me know what you think.

crates/python/README.md
```diff
@@ -7,3 +7,58 @@ to compile and run Quil programs on Rigetti quantum processors. Internally, it i
...
+logging.basicConfig(level=5)
+```
+
+#### Runtime Configuration and Caching
+ 
+`pyo3_log` caches loggers and their level filters to improve performance. This means that logger re-configuration done at runtime may cause logs to log or not log as expected in certain situations. If this is a concern, [this section of the pyo3_log documentation](https://docs.rs/pyo3-log/latest/pyo3_log/#performance-filtering-and-caching) goes into more detail.
+
+#### Filtering Logs
+
+Because the logs are emitted from a Rust library, the logger names will correspond to the fully qualified path of the Rust module in the library where the log occured. These fully qualified paths all have their own logger, and have to be configured individually.
```
kalzoo, line 39:
```suggestion
Because the logs are emitted from a Rust library, the logger names will correspond to the fully qualified path of the Rust module in the library where the log occurred. These fully qualified paths all have their own logger, and have to be configured individually.
```

crates/python/README.md
```diff
@@ -7,3 +7,58 @@ to compile and run Quil programs on Rigetti quantum processors. Internally, it i
...
+Rust has a `TRACE` log level that doesn't exist in Python. It is less severe than `DEBUG` and is set to a value of 5. While the `DEBUG` level is recommended for troubleshooting, you can choose to target `TRACE` level logs and below like so:
+
+```python
+import logging
+logging.basicConfig(level=5)
+```
+
+#### Runtime Configuration and Caching
+ 
+`pyo3_log` caches loggers and their level filters to improve performance. This means that logger re-configuration done at runtime may cause logs to log or not log as expected in certain situations. If this is a concern, [this section of the pyo3_log documentation](https://docs.rs/pyo3-log/latest/pyo3_log/#performance-filtering-and-caching) goes into more detail.
```
kalzoo, line 35:
```suggestion
`pyo3_log` caches loggers and their level filters to improve performance. This means that logger re-configuration done at runtime may cause unexpected logging behavior in certain situations. If this is a concern, [this section of the pyo3_log documentation](https://docs.rs/pyo3-log/latest/pyo3_log/#performance-filtering-and-caching) goes into more detail.
```

Should we expose a simple binding to `reset_logging` and indicate that here? Logging is simple when it works but can be maddening when it doesn't, and an ability for the user to dig themselves out of a caching problem will likely be appreciated. If that sounds good I can add the commit.

kalzoo:
> @kalzoo I added a section to the README with examples on filtering. Let me know what you think.

looks good!

PR #312, opened 2023-06-23
feat!: Standardize use_gateway

MarquessV:


crates/lib/src/executable.rs
```diff
@@ -478,6 +482,7 @@ impl<'execution> Executable<'_, 'execution> {
         &mut self,
         quantum_processor_id: S,
         translation_options: Option<TranslationOptions>,
+        connection_strategy: ConnectionStrategy,
```
kalzoo, line 485: I would pool these into an `ExecutionOptions` with a builder:

* This'll be the last time we have to make a breaking change in this signature. Future options can just be added to that struct
* Keeps this from becoming a longer and longer list as we add more configuration

crates/lib/src/executable.rs
```diff
@@ -585,9 +604,9 @@ pub enum Error {
     /// QPU.
     #[error("Could not authenticate a request to QCS for the requested QPU.")]
     Authentication,
-    /// The requested QPU was not found. Either the QPU does not exist or you do not have access to it.
-    #[error("The requested QPU was not found.")]
-    QpuNotFound,
+    /// An API error occured while connecting to the QPU.
```
kalzoo, line 607:
```suggestion
    /// An API error occurred while connecting to the QPU.
```

crates/lib/src/qpu/api.rs
```diff
@@ -91,12 +107,21 @@ impl From<String> for JobId {
...
+/// * `client` - The [`Qcs`] client to use.
+/// * `connection_strategy` - The [`ConnectionStrategy`] to use. If `job_target` is an
+///       endpoint ID, then direct access to that endpoint ID overrides this parameter.
 pub async fn submit(
     job_target: &JobTarget,
     program: EncryptedControllerJob,
     patch_values: &Parameters,
     client: &Qcs,
-) -> Result<JobId, GrpcClientError> {
+    connection_strategy: ConnectionStrategy,
```
kalzoo, line 123: likewise - `ExecutionOptions`

crates/lib/src/qpu/api.rs
```diff
@@ -144,21 +171,152 @@ pub async fn retrieve_results(
...
+}
+
+/// The connection strategy to use when submitting and retrieving jobs from a QPU.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum ConnectionStrategy {
+    /// Connect through the publicly accessible gateway.
+    #[default]
+    GatewayAlways,
+    /// Connect directly to the QPU endpoint, bypassing the gateway. Should only be used when you
+    /// have direct access and an active reservation.
```
kalzoo, line 194:
```suggestion
    /// have direct network access and an active reservation.
```

crates/lib/src/qpu/api.rs
```diff
@@ -144,21 +171,152 @@ pub async fn retrieve_results(
...
+            next_page_token = accessors.next_page_token.clone();
+            if next_page_token.is_none() {
+                break;
+            }
+        }
+        gateways.sort_by_key(|acc| acc.rank);
+        gateways
+            .first()
+            .map(|accessor| accessor.url.clone())
+            .ok_or_else(|| QpuApiError::GatewayNotFound(quantum_processor_id.to_string()))
```
kalzoo, line 270: FYI, `min_by_key` is a little more efficient than `sort` -> `first`.

As an optimization, we can avoid allocating the `Vec` if we just accumulate the minimum value as we loop. In practice, not likely to affect anything (this is called rarely and is dominated by IO time), but also an easy fix.

crates/python/qcs_sdk/qpu/api.pyi
```diff
@@ -92,6 +94,7 @@ def submit(
     :param quantum_processor_id: The ID of the quantum processor to run the executable on.
     :param client: The ``QCSClient`` to use. Creates one using environment configuration if unset - see https://docs.rigetti.com/qcs/references/qcs-client-configuration
     :param endpoint_id: submit the program to an explicitly provided endpoint. If `None`, the default endpoint for the given quantum_processor_id is used.
+    :param connection_strategy: The ``ConnectionStrategy`` to use to connect to the QPU. If unset, the default is ConnectionStrategy.GatewayAlways. If `enpodint_id` is set, then direct access to that endpoint overrides this parameter.
```
kalzoo, line 97:
```suggestion
    :param connection_strategy: The ``ConnectionStrategy`` to use to connect to the QPU. If unset, the default is ConnectionStrategy.GatewayAlways. If `endpoint_id` is set, then direct access to that endpoint overrides this parameter.
```

and below

crates/lib/src/qpu/api.rs
```diff
@@ -74,6 +84,96 @@ impl From<&JobTarget> for get_controller_job_results_request::Target {
     }
 }
 
+impl JobTarget {
```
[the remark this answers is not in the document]
kalzoo: You're right, they do seem to serve the same purpose. That being, "how do you want to connect to run your job?" to which our answers are

1. The gateway for QPU Q
2. The direct access URL for QPU Q
3. Ignore the QPU and connect to endpoint E

It does make sense to roll the endpoint approach into `ConnectionStrategy`.

crates/python/qcs_sdk/qpu/api.pyi
Shadow53, line 97:
```suggestion
    :param connection_strategy: The ``ConnectionStrategy`` to use to connect to the QPU. If unset, the default is ``ConnectionStrategy.GatewayAlways``. If `endpoint_id` is set, then direct access to that endpoint overrides this parameter.
```

crates/python/qcs_sdk/qpu/api.pyi
```diff
@@ -116,6 +120,7 @@ async def submit_async(
...
```
Shadow53, line 123: See above

crates/python/qcs_sdk/qpu/api.pyi
```diff
@@ -137,6 +143,7 @@ def retrieve_results(
     :param quantum_processor_id: The ID of the quantum processor the job ran on.
     :param client: The ``QCSClient`` to use. Creates one using environment configuration if unset - see https://docs.rigetti.com/qcs/references/qcs-client-configuration
     :param endpoint_id: retrieve the results of a program submitted to an explicitly provided endpoint. If `None`, the default endpoint for the given quantum_processor_id is used.
+    :param connection_strategy: The ``ConnectionStrategy`` to use to connect to the QPU. If unset, the default is ConnectionStrategy.GatewayAlways. If `enpodint_id` is set, then direct access to that endpoint overrides this parameter.
```
Shadow53, line 146: See above

crates/python/qcs_sdk/qpu/api.pyi
Shadow53: Not sure if the double ticks is correct for `GatewayAlways`

crates/lib/src/qpu/api.rs
```diff
@@ -144,21 +171,152 @@ pub async fn retrieve_results(
...
+    }
+
+    async fn get_gateway_address(
+        &self,
+        quantum_processor_id: &str,
+        client: &Qcs,
+    ) -> Result<String, QpuApiError> {
+        let mut gateways = Vec::new();
+        let mut next_page_token = None;
+        loop {
```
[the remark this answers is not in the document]
kalzoo: In practice there is only ever one or two accessors in the list, so the rest of the mechanism is just forward compatibility. If that premise changes we can add a query parameter to sort by rank.

crates/lib/src/qpu/api.rs
```diff
@@ -144,21 +171,152 @@ pub async fn retrieve_results(
...
+
+/// The connection strategy to use when submitting and retrieving jobs from a QPU.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum ConnectionStrategy {
+    /// Connect through the publicly accessible gateway.
+    #[default]
+    GatewayAlways,
+    /// Connect directly to the QPU endpoint, bypassing the gateway. Should only be used when you
+    /// have direct access and an active reservation.
+    DirectAccessAlways,
```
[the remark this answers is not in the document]
kalzoo: Yeah, that's a good point, `GatewayOnly` or just `Gateway` do read more clearly. To illustrate, the next option is one of them falling back to the other if unavailable.

crates/lib/src/executable.rs
MarquessV: Something I keep tripping over as we add these options structs is that they are all constructed and modified differently. Some use public fields, some mutate self, others return updated copies, none so far use an actual builder type. What do you have in mind when you say builder? I can use that approach for `ExecutionOptions`, then create a ticket to consolidate them. Maybe [`derive_builder`](https://docs.rs/derive_builder/latest/derive_builder/)?

crates/lib/src/qpu/api.rs
MarquessV: This was lifted out of the old client code, but I went ahead and refactored it.

Just to clarify, `accessor.rank` is an `Option`, meaning it could be `None`. How should that value rank against accessors with `Some(i64)`. By default Rust rules it's the minimum value (`None < Some(0)`). I have a feeling we want the opposite though, so I defaulted them to i64::MAX. Is that correct? Or does `None` signify the preferred gateway?

kalzoo: Yep, I agree, good spot - no rank is "lowest rank"

crates/lib/src/executable.rs
kalzoo: Good eye! yeah that stuff matters in usage but can be hard to spot in reviews and focused work. I do like `derive_builder` and that provides a friendly UX.

MarquessV: Okay, switched to using `ExecutionOptions`

crates/lib/src/qpu/api.rs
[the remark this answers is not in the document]
MarquessV: Okay, refactored it. It has it's own set of oddities but it's handled more by the SDK than by the user.

crates/python/src/qpu/api.rs
```diff
@@ -187,17 +187,82 @@ py_function_sync_async! {
...
+    fn connection_strategy(&self, connection_strategy: PyConnectionStrategy) -> Self {
+        // `derive_builder::Builder` doesn't implement AsMut, meaning we can't use `PyWrapperMut`,
+        // which forces us into this awkward clone.
+        Self::from(
+            self.as_inner()
+                .clone()
+                .connection_strategy(connection_strategy.as_inner().clone())
+                .clone(),
+        )
+    }
```
MarquessV, line 243: I don't think there is any way around doing at least two clones here. We can't return `&Self` because of Python/pyo3, and there is no way to get the inner `ConnectionStrategy` directly.

kalzoo: What if the python-side builder were just more limited and didn't allow chaining? That is, `fn set_connection_strategy(...) -> () `?

crates/python/src/qpu/api.rs
```diff
@@ -41,9 +43,9 @@ create_init_submodule! {
 /// Errors that may occur when submitting a program for execution
 #[derive(Debug, thiserror::Error)]
 enum RustSubmissionError {
-    /// Failed a gRPC API call
-    #[error("Failed a gRPC call: {0}")]
-    GrpcError(#[from] GrpcClientError),
+    /// An API error occured
+    #[error("An API error occured: {0}")]
```
kalzoo, line 47:
```suggestion
    /// An API error occurred
    #[error("An API error occurred: {0}")]
```

crates/python/qcs_sdk/qpu/api.pyi
```diff
@@ -159,10 +167,45 @@ async def retrieve_results_async(
...
+
+@final
+class ConnectionStrategy(Enum):
+    """An enum containing variants for each possible quantum processor connection strategy."""
+
+    GatewayOnly = "GatewayOnly"
+    """
+    Always connect through the publicly accessible gateway.
+    """
+    DirectAccess = "DirectAccess"
```
kalzoo, line 208:
```suggestion
    Gateway = "Gateway"
    """
    Always connect through the publicly accessible gateway.
    """
    DirectAccess = "DirectAccess"
```

To extend @Shadow53 's point, seems like these would benefit from consistency - either both `Only` or neither

crates/lib/src/executable.rs
```diff
@@ -726,13 +745,10 @@ impl<'a> JobHandle<'a> {
         self.job_id.clone()
     }
 
-    /// The execution target of the QCS Job, either the quantum processor or an expicit endpoint.
+    /// The quatnum processor ID the job was submitted to.
```
kalzoo, line 748:
```suggestion
    /// The ID of the quantum processor to which the job was submitted.
```

crates/lib/src/executable.rs
```diff
@@ -502,13 +507,23 @@ impl<'execution> Executable<'_, 'execution> {
     /// Call [`Executable::retrieve_results`] to wait for execution to complete and retrieve the
     /// results.
     ///
+    /// # Arguments
+    /// 1. `quantum_processor_id`: The name of the QPU to translate the program for on.
```
kalzoo, line 511:
```suggestion
    /// 1. `quantum_processor_id`: The ID of the QPU for which to translate the program.
```

(recognizing that it was copied from above)

crates/lib/src/qpu/api.rs
```diff
@@ -91,74 +72,287 @@ impl From<String> for JobId {
...
+///
+/// Use [`Default`] to get a reasonable set of defaults, or start with [`ExecutionOptionsBuilder`]
+/// to build a custom set of options.
+#[derive(Builder, Clone, Debug, Default, PartialEq, Eq)]
+pub struct ExecutionOptions {
+    #[doc = "The [`ConnectionStrategy`] to use to establish a connection to the QPU."]
+    connection_strategy: ConnectionStrategy,
+}
+
+impl ExecutionOptions {
```
kalzoo, line 185:
```suggestion
impl ExecutionOptions {
  pub fn builder() -> ExecutionOptionsBuilder {
    ExecutionOptionsBuilder::default()
  }
```

A common pattern I've seen and used to make it more ergonomic to get at the builder

crates/lib/src/qpu/api.rs
```diff
@@ -91,74 +72,287 @@ impl From<String> for JobId {
...
+
+/// The connection strategy to use when submitting and retrieving jobs from a QPU.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub enum ConnectionStrategy {
+    /// Connect through the publicly accessible gateway.
+    #[default]
+    GatewayOnly,
+    /// Connect directly to the default endpoint, bypassing the gateway. Should only be used when you
+    /// have direct network access and an active reservation.
+    DirectAccess,
```
kalzoo, line 201: (this would be a better place for the note about name consistency, `Only` vs ``

Noting that the docstrings are slightly different here from the python bindings ("always"). Those seem a little more specific.

crates/lib/src/qpu/api.rs
```diff
@@ -91,74 +72,287 @@ impl From<String> for JobId {
...
-        .ok_or_else(|| GrpcClientError::ResponseEmpty("Job Execution Results".into()))
+        .ok_or_else(|| GrpcClientError::ResponseEmpty("Job Execution Results".into()))?)
+}
+
+/// Options avaialable when executing a job on a QPU.
+///
+/// Use [`Default`] to get a reasonable set of defaults, or start with [`ExecutionOptionsBuilder`]
+/// to build a custom set of options.
+#[derive(Builder, Clone, Debug, Default, PartialEq, Eq)]
+pub struct ExecutionOptions {
```
kalzoo, line 180: looks great 👍

crates/lib/src/qpu/api.rs
```diff
@@ -91,74 +72,287 @@ impl From<String> for JobId {
...
+
+    fn get_results_target(
+        &self,
+        quantum_processor_id: &str,
+    ) -> get_controller_job_results_request::Target {
+        match self {
+            Self::EndpointId(endpoint_id) => {
+                get_controller_job_results_request::Target::EndpointId(endpoint_id.to_string())
+            }
+            _ => get_controller_job_results_request::Target::QuantumProcessorId(
```
kalzoo, line 229: I'd explicitly name the variants here to keep this up to date as more are added.

crates/lib/src/qpu/api.rs
```diff
@@ -91,74 +72,287 @@ impl From<String> for JobId {
...
+            }
+            _ => execute_controller_job_request::Target::QuantumProcessorId(
+                quantum_processor_id.to_string(),
+            ),
+        }
+    }
+
+    fn get_results_target(
+        &self,
+        quantum_processor_id: &str,
```
kalzoo, line 223: I _think_ this should be `Option<&str>` with a return type of `Option<...>` so that the user isn't forced to provide a QPU ID if it's never used.

crates/lib/src/qpu/api.rs
```diff
@@ -91,74 +72,287 @@ impl From<String> for JobId {
...
+    DirectAccess,
+
+    /// Connect directly to a specific endpoint using its ID.
+    EndpointId(String),
+}
+
+/// Methods that help select the right controller service client given a quantum processor ID and
+/// [`ConnectionStrategy`].
+impl ConnectionStrategy {
+    fn get_job_target(&self, quantum_processor_id: &str) -> execute_controller_job_request::Target {
```
kalzoo, line 210: likewise about `Option<>`

crates/lib/src/qpu/api.rs
```diff
@@ -91,74 +72,287 @@ impl From<String> for JobId {
...
+            _ => get_controller_job_results_request::Target::QuantumProcessorId(
+                quantum_processor_id.to_string(),
+            ),
+        }
+    }
+
+    async fn get_controller_client(
+        &self,
+        client: &Qcs,
+        quantum_processor_id: &str,
```
kalzoo, line 238: Likewise.

For all of these, if it's easier to return a `Result` that complains about needing a QPU ID, that's fair too

kalzoo (dismissed):

crates/python/src/qpu/api.rs
MarquessV: I ended up going with this suggestion because I realized, as written, it had a sharp edge. If you weren't using chaining, the inner builder wasn't actually updated, so if you instead opted to line by line update the builder instead of chaining, and threw away the returned builders along the way, you would effectively be doing nothing.

Restricting the return type lets us implement it as a setter, which forces users into setting it as a property and gives user one clear, correct way of doing it. Unfortunately, it doesn't reduce the amount of clones needed

crates/lib/src/qpu/api.rs
MarquessV: Yeah, that makes sense. A bit less clunky in the EndpointId case.

crates/python/qcs_sdk/qpu/api.pyi
```diff
@@ -103,19 +104,19 @@ def submit(
...
+    connection_strategy: Optional[ConnectionStrategy] = None,
 ) -> str:
     """
     Submits an executable `program` to be run on the specified QPU.
     (async analog of ``submit``)
 
     :param program: An executable program (see `translate`).
     :param patch_values: A mapping of symbols to their desired values (see `build_patch_values`).
-    :param quantum_processor_id: The ID of the quantum processor to run the executable on.
+    :param quantum_processor_id: The ID of the quantum processor the job ran on. This field is required, unless being used with the `ConnectionStrategy.endpoint_id()` execution option.
```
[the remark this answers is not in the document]
MarquessV: You're right, I think I wrote the updated docstring for `retrieve_results` first, stamped it on submit, but didn't fix the tense.

crates/lib/Cargo.toml
```diff
@@ -59,6 +59,11 @@ zmq = { version = "0.10.0" }
 itertools = "0.10.5"
 rstest = "0.17.0"
 insta = "1.29.0"
+derive_builder = "0.12.0"
+
+# Not used directly by the crate, but required to avoid using yanked versions of crates that trigger
+# `cargo-deny` errors
+hermit-abi = "^0.3.2"
```
MarquessV, line 66: 0.3.1 was yanked and made `cargo-deny` throw an error. I assume we don't want to maintain `Cargo.lock` in this repo, since it is library (though it is tracked currently, I'll create a follow-up issue to remove it), so is this how we should handle it? There is no way to declare an exception for specific yanked crates.

I'm not sure if this makes sense, since it doesn't allow users to use crates below `0.3.2` that weren't necessarily yanked. Alternatively, we could reduce the severity of yanked crates to `warn`, which is [the default](https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html#the-yanked-field-optional).

MarquessV: Per a recommendation by Kalan, I tried to see if we could have a more flexible bounds that excludes `0.3.1` by specifying multiple ranges like you can in [npm](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#dependencies). Unfortunately, [the cargo book] suggests that it has no such equivalent.

BatmanAoD: We talked about having a lock-file in libraries used by Python [here](https://rigetti.slack.com/archives/G011GB25W0K/p1687367116016159) a few weeks ago... looks like we chose to ignore the lockfile in `quil-rs`.

Won't this self-resolve any time a downstream Cargo build is done without `--locked`?

MarquessV: Yeah, I tried to be clever and not rely on changing the lockfile since we were going to remove it, but it is the reason the yanked version was being chosen. I reverted this change and just updated the lockfile. I'll leave the removal to #319.

crates/lib/src/qpu/api.rs
```diff
@@ -91,74 +74,316 @@ impl From<String> for JobId {
...
+            )
+            .await?;
+
+            let accessor = accessors
+                .accessors
+                .into_iter()
+                .filter(|acc| {
+                    acc.live
+                    // `as_deref` needed to work around the `Option<Box<_>>` type.
+                    && acc.access_type.as_deref() == Some(&QuantumProcessorAccessorType::GatewayV1)
```
BatmanAoD, line 319: 👍🏻 

I definitely thought there would be a way to accomplish this without an explicit `deref` anywhere... but without deref-patterns, it seems there isn't.

BatmanAoD (approved):

crates/lib/src/qpu/api.rs
```diff
@@ -91,74 +74,316 @@ impl From<String> for JobId {
...
+
+/// Methods that help select the right controller service client given a quantum processor ID and
+/// [`ConnectionStrategy`].
+impl ConnectionStrategy {
+    fn get_job_target(
+        &self,
+        quantum_processor_id: Option<&str>,
+    ) -> Result<execute_controller_job_request::Target, QpuApiError> {
+        match self {
+            Self::EndpointId(endpoint_id) => Ok(
```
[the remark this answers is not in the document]
BatmanAoD: Alternatively, should this function return an error if both `quantum_processor_id` and `endpoint_id` are provided?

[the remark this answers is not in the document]
MarquessV: There is no direct way to validate that as far as I know, though we may be able to make API calls to connect the dots. Either way, I don't think it's worth the squeeze, at least not for this PR. I suspect that if a user is calling an Endpoint ID directly, they know what they are doing. If this turns out to be a foot gun though, I'm happy to adjust. At the very least, the API as written allows us to do that without breaking the signature.

PR #322, opened 2023-07-06
fix: The MissingQubitBenchmark error has been demoted to a warning.

MarquessV:
A default fidelity of 0.0 is used instead.

### DIFF
--- a/crates/lib/src/compiler/isa/qubit.rs
+++ b/crates/lib/src/compiler/isa/qubit.rs
@@ -172,8 +172,9 @@ const DEFAULT_DURATION_RX: f64 = 50.0;
 fn rx_gates(node_id: i64, frb_sim_1q: &FrbSim1q) -> Result<Vec<Operator>, Error> {
     let fidelity = match frb_sim_1q.fidelity_for_qubit(node_id) {
         Ok(fidelity) => Ok(fidelity),
-        Err(Error::MissingBenchmarkForQubit(e)) => {
-            tracing::warn!(e);
+        Err(error @ Error::MissingBenchmarkForQubit(_)) => {
+                    #[cfg(feature = "tracing")]
+                    tracing::warn!(%error);
             Ok(0.0)
         }
         Err(e) => Err(e),
--- a/crates/lib/src/compiler/isa/qubit.rs
+++ b/crates/lib/src/compiler/isa/qubit.rs
@@ -171,6 +171,8 @@ fn rx_gates(node_id: i64, frb_sim_1q: &FrbSim1q) -> Result<Vec<Operator>, Error>
     let fidelity = frb_sim_1q
         .fidelity_for_qubit(node_id)
         .or_else(|e| match e {
+            // Prevents compilation error when building without the `tracing` feature flag
+            #[allow(unused_variables)]
             error @ Error::MissingBenchmarkForQubit(_) => {
                 #[cfg(feature = "tracing")]
                 tracing::warn!(%error);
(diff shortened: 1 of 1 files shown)


crates/lib/src/compiler/isa/qubit.rs
```diff
@@ -267,6 +274,50 @@ mod describe_rx_gates {
...
+
+    #[test]
+    fn it_defaults_to_0_for_missing_benchmark() {
+        let frb_sim_1q = FrbSim1q(Vec::new());
+        let gates = rx_gates(1, &frb_sim_1q).expect("should not error for missing benchmark");
+        let expected = vec![
+            Operator::Gate {
+                arguments: vec![Argument::Int(1)],
+                duration: 50.0,
+                fidelity: 1.0,
```
MarquessV, line 286: `rx_gates` hard codes the fidelity value of 1.0 for `RX(0.0)`. I assume that's because it is a no-op.

crates/lib/src/compiler/isa/qubit.rs
```diff
@@ -170,7 +170,14 @@ impl FrbSim1q {
 const DEFAULT_DURATION_RX: f64 = 50.0;
 
 fn rx_gates(node_id: i64, frb_sim_1q: &FrbSim1q) -> Result<Vec<Operator>, Error> {
-    let fidelity = frb_sim_1q.fidelity_for_qubit(node_id)?;
+    let fidelity = match frb_sim_1q.fidelity_for_qubit(node_id) {
+        Ok(fidelity) => Ok(fidelity),
+        Err(Error::MissingBenchmarkForQubit(e)) => {
+            tracing::warn!(e);
```
kalzoo, line 176: Does this need `#[cfg(...)]`?

BatmanAoD, line 176:
```suggestion
        Err(error @ Error::MissingBenchmarkForQubit(_)) => {
                    #[cfg(feature = "tracing")]
                    tracing::warn!(%error);
```

That resolves the "failed to resolve" check, and also makes the output from this:

```
WARN qcs::compiler::isa::qubit: e=15
```

to this:

```
WARN qcs::compiler::isa::qubit: error=Benchmark missing for qubit 15
```

crates/lib/src/compiler/isa/qubit.rs
```diff
@@ -170,7 +170,14 @@ impl FrbSim1q {
 const DEFAULT_DURATION_RX: f64 = 50.0;
 
 fn rx_gates(node_id: i64, frb_sim_1q: &FrbSim1q) -> Result<Vec<Operator>, Error> {
-    let fidelity = frb_sim_1q.fidelity_for_qubit(node_id)?;
+    let fidelity = match frb_sim_1q.fidelity_for_qubit(node_id) {
```
kalzoo, line 173: style: `.or_else` may be more succinct

BatmanAoD:
With this change, I am able to submit & execute on Ankaa without pyquil!

kalzoo (approved):
Looks like the `add_operation` docstring may also need an update under `Errors`. 

Small comments though - good to go 👍

crates/lib/src/compiler/isa/qubit.rs
kalzoo: yep, @BatmanAoD 's suggestion above lgtm

crates/lib/src/compiler/isa/qubit.rs
MarquessV: Unfortunately, `or_else` doesn't let us distinguish between error types, so it's not possible to say "fallback to 0.0 for `Error::MissingQubitBenchmark` but not for any other kind of error".

crates/lib/src/compiler/isa/qubit.rs
MarquessV: Great catch. This is the downside of using rust-analyzer with all features.

Fun little edge case - `error` is an unused binding if the `tracing` feature isn't enabled, which causes compilation errors. That is the first time I've had to legitimately use `#[allow(unused_variables)]`.

PR #323, opened 2023-07-07
feat: Add timeout to `ExecutionOptions` builder

MarquessV:


crates/lib/src/qpu/api.rs
```diff
@@ -155,15 +150,10 @@ pub async fn retrieve_results(
 
     let request = GetControllerJobResultsRequest {
         job_execution_id: job_id.0,
-        target: Some(
-            execution_options
-                .connection_strategy
-                .get_results_target(quantum_processor_id)?,
-        ),
+        target: Some(execution_options.get_results_target(quantum_processor_id)?),
```
kalzoo, line 153: Noting this because it's in the diff, even though it's preexisting; this is optional in the request, and so it should be optional from `get_results_target` as well.

crates/lib/src/qpu/api.rs
```diff
@@ -238,11 +237,11 @@ impl ConnectionStrategy {
...
-            Self::EndpointId(endpoint_id) => Ok(
+        match self.connection_strategy() {
+            ConnectionStrategy::EndpointId(endpoint_id) => Ok(
                 get_controller_job_results_request::Target::EndpointId(endpoint_id.to_string()),
             ),
-            Self::Gateway | Self::DirectAccess => Ok(
+            ConnectionStrategy::Gateway | ConnectionStrategy::DirectAccess => Ok(
                 get_controller_job_results_request::Target::QuantumProcessorId(
                     quantum_processor_id
                         .ok_or(QpuApiError::MissingQpuId)?
```
kalzoo, line 247: Pairing with my other comment - as long as we're here, this should return an `Option` rather than a `Result`, and `MissingQpuId` isn't needed here. One should be able to submit the job without a `target`, because the backend does accept that in direct-access mode. Only the gateway insists on a target, and we should allow the service to return the error stating as much rather than getting the client involved.

crates/lib/src/qpu/api.rs
```diff
@@ -257,38 +256,40 @@ impl ConnectionStrategy {
...
     }
 
     fn grpc_address_to_client(
+        &self,
         address: &str,
         client: &Qcs,
     ) -> Result<ControllerClient<RefreshService<Channel, ClientConfiguration>>, QpuApiError> {
         let uri = parse_uri(address).map_err(QpuApiError::GrpcError)?;
-        let channel = get_channel(uri).map_err(|err| QpuApiError::GrpcError(err.into()))?;
+        let channel = get_channel_with_timeout(uri, self.timeout())
```
kalzoo, line 291: As an earmark for a discussion elsewhere: seems like we should use a channel wrapper builder to support all the different options that might be applied to a channel and to prevent having to support every combination of options we might want

MarquessV: Yeah, I agree. I added `get_channel_with_timeout` since it seemed like it would be one of our most common use cases and would be likely to save us some boiler plate. However, for future-proofing, I also added a `get_channel_with_endpoint` function that takes a tonic [`Endpoint`](https://docs.rs/tonic/latest/tonic/transport/channel/struct.Endpoint.html) that can be configured however a user likes.

Shadow53 (approved):
LGTM

PR #327, opened 2023-07-13
ci: Define a ci-flow task for Python that includes unit tests and stubtest

MarquessV:


Shadow53 (approved):
LGTM.

Side note: looks like your editor auto-formatted the `Makefile.toml`, which isn't a huge deal, but makes the diff noisier than necessary. Might want to revert those changes and/or change editor settings to avoid this happening again?

MarquessV:
@Shadow53 fair, that is `taplo`, which I like for consistency, but since the whole team doesn't use it, I've reverted it.

PR #334, opened 2023-07-26
feat: A new `diagnostics` module has been added. The `get_report` function will print a summary of key diagnostic data to aid in debugging.

kalzoo:
```
package version: 0.15.1
rustc version: rustc 1.71.0 (8ede3aae2 2023-07-12)
features: 
qcs:
  address: https://api.qcs.rigetti.com
  network reachable: true
  authentication: <id> / email <email>
  ^^^ for your security, remove this line before posting publicly! ^^^
quilc:
  address: tcp://127.0.0.1:5555
  version: 1.26.0
  available: true
qvm:
  address: http://127.0.0.1:5000
  version: -
  available: false

```

TODO:

- [x] Fix python build failures in use of `py_function_sync_async`
- [x] Add Python-specific diagnostics within the python bindings
- [x] Rename `gather` to `get_report` to make it more obvious that it's a string.


Cargo.lock
Shadow53, line 1: Curious why this was deleted.

crates/lib/src/diagnostics.rs
```diff
@@ -0,0 +1,212 @@
...
+            QvmDiagnostics::gather(&client),
+        )
+        .await;
+
+        Self {
+            version: build_info::PKG_VERSION.to_owned(),
+            rust_version: build_info::RUSTC_VERSION.to_owned(),
+            features: build_info::FEATURES
+                .iter()
+                .map(std::borrow::ToOwned::to_owned)
```
Shadow53, line 48: Could you use `copied()` or `cloned()` instead of this `map()`?

Cargo.lock
kalzoo: Ah - not intentionally, there was a merge conflict and I deleted it so it would be rebuilt

crates/lib/Cargo.toml
```diff
@@ -73,3 +59,7 @@ tokio = { version = "1.21.2", features = ["macros", "rt-multi-thread"] }
 warp = { version = "0.3.3", default-features = false }
 regex = "1.7.0"
 test-case = "3.1.0"
+tracing-subscriber = "0.3.17"
```
kalzoo, line 62:
```suggestion
```

This will be useful to have, but it's unrelated to this diff

crates/lib/src/diagnostics.rs
kalzoo: yeah, not sure what I was thinking here

crates/lib/src/lib.rs
```diff
@@ -8,7 +8,7 @@
 #![forbid(unsafe_code)]
 #![warn(future_incompatible)]
 #![warn(rust_2018_compatibility, rust_2018_idioms)]
-#![deny(
+#![warn(
```
kalzoo, line 11: Across the board, we should get out of the habit of causing compiler errors for style issues, they just add noise while development is in full swing. They can be raised as errors during linting.

crates/lib/src/qpu/execution.rs
```diff
@@ -159,8 +159,12 @@ impl<'a> Execution<'a> {
             quil.parse().map_err(Error::Quil)?
         };
 
+        println!("program: {program}");
+        let rewritten = RewrittenProgram::try_from(program).map_err(Error::RewriteArithmetic)?;
+        println!("rewritten: {}", rewritten.to_string().0);
```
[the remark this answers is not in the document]
kalzoo: indeed, whoops. This was mistakenly committed from some other unstaged changes I have

crates/python/src/lib.rs
```diff
@@ -68,3 +68,10 @@ fn reset_logging() {
         }
     }
 }
+
+py_sync::py_function_sync_async! {
+    #[cfg()]
```
kalzoo, line 73: This was to satisfy the necessary `meta` in the macro pattern, which - if possible - we may want to make optional

MarquessV: The build failures were due to the missing `#[pyfunction]` attribute, so adding it fixes those and satisfies the macro pattern. I believe the reason it's not included in the macro by default, is because it makes it hard to customize the attribute, which we commonly do (e.g. `#[pyfunction(client = "None", options = "None")]).

crates/lib/src/diagnostics.rs
MarquessV: `to_vec()` works in this case, so I went with that.

crates/python/src/lib.rs
kalzoo: aha - thanks. Worth adding to the macro docstring, or is it omitted for a reason?

MarquessV: I can't think of a good reason, probably an over sight, added it!

crates/python/src/lib.rs
```diff
@@ -68,3 +69,9 @@ fn reset_logging() {
         }
     }
 }
+
+#[pyfunction]
+#[pyo3(name = "_gather_diagnostics")]
+fn gather_diagnostics(py: Python<'_>) -> PyResult<String> {
+    py_sync::py_sync!(py, async { Ok(qcs::diagnostics::get_report().await) })
+}
```
MarquessV, line 77: Decided to only export a synchronous version of this function to keep the API simpler.

crates/python/qcs_sdk/__init__.pyi
```diff
@@ -11,6 +11,11 @@ from qcs_sdk.compiler.quilc import CompilerOpts
 
 from qcs_sdk.client import QCSClient as QCSClient
 
+from . import qpu as qpu
+from . import qvm as qvm
+from . import compiler as compiler
+from . import client as client
```
MarquessV, line 17: This was the result of `stubtest` complaining about missing modules for these names after setting `__all__` in the root of `__init__.py`. It turns out the way we defined them were only valid as packages and re-exporting them here makes them proper sub-modules. That makes sense, but `stubtest` never complained about it before. Either way, happy that we stumbled into this, as I think it'll resolve some inconsistent behavior I've seen with different tools and import styles.

crates/python/tests/diagnostics.py
```diff
@@ -0,0 +1,8 @@
+from qcs_sdk import gather_diagnostics
+
+
+def test_gather_diagnostics():
```
kalzoo, line 4: 👍

crates/python/qcs_sdk/__init__.py
```diff
@@ -1,4 +1,24 @@
+# type: ignore
 # See the following documentation for why this file is necessary:
 # https://pyo3.rs/v0.18.0/python_typing_hints#__init__py-content
 
+import sys
+
 from .qcs_sdk import *
+
+
+def gather_diagnostics() -> str:
```
kalzoo, line 10:
```suggestion
def get_diagnostics_report() -> str:
    """
    Return a string describing the package and its environment for use in bug reporting and diagnosis.
    
    Note: this format is not stable and its content may change between versions.
    """
```

crates/python/qcs_sdk/__init__.py
```diff
@@ -1,4 +1,24 @@
...
+Python C api version: {sys.api_version}
+Python executable: {sys.executable}
+venv prefix: {sys.prefix}
+platform: {sys.platform}
+{_gather_diagnostics()}
+   """
+
+
+__doc__ = qcs_sdk.__doc__
+__all__ = ["gather_diagnostics"] + getattr(qcs_sdk, "__all__", [])
```
kalzoo, line 24:
```suggestion
__all__ = ["get_diagnostics_report"] + getattr(qcs_sdk, "__all__", [])
```

crates/python/qcs_sdk/__init__.pyi
```diff
@@ -356,11 +353,16 @@ class RegisterData:
...
     See <https://docs.rs/pyo3-log/latest/pyo3_log/> for more information.
-    """
\ No newline at end of file
+    """
+
+def gather_diagnostics() -> str:
+    """
+    Gather and return a summary of diagnostics that describe how the package was built and the environment
+    it's running in. Helpful for diagnosing unexpected behavior.
+    """
```
kalzoo, line 368: Is this needed for direct python exports? Is it that this file clobbers the inferred type hints?

crates/python/tests/diagnostics.py
```diff
@@ -0,0 +1,8 @@
...
+    """
+    Assert that gathering diagnostics doesn't panic.
+    """
+    gather_diagnostics()
```
kalzoo, line 8:
```suggestion
    get_diagnostics_report()
```

crates/python/tests/diagnostics.py
```diff
@@ -0,0 +1,8 @@
+from qcs_sdk import gather_diagnostics
```
kalzoo, line 1:
```suggestion
from qcs_sdk import get_diagnostics_report
```

crates/python/src/lib.rs
```diff
@@ -57,7 +57,8 @@ fn qcs_sdk(py: Python<'_>, m: &PyModule) -> PyResult<()> {
         }
         Err(e) => eprintln!("Failed to initialize the qcs_sdk logger: {e}"),
     }
-    init_submodule("qcs_sdk", py, m)
+    init_submodule("qcs_sdk", py, m)?;
+    m.add("__version__", env!("CARGO_PKG_VERSION"))
```
kalzoo, line 61: 👍

kalzoo (commented):
Great! I can't approve since I was the first author, but lgtm.

My thought on the renames from gather_diagnostics -> get_report is that we may want to vend an actual "diagnostics" type at some point (like the private type in the Rust crate), and "report" is instead the string derived from that object rather than the diagnostics itself. Minor point, not important.

PR #337, opened 2023-07-28
feat: Default endpoints and gateway accessors are cached between requests

### LINKED ISSUE #336: The gateway accessor and default-endpoint for a QPU ID should be cached
Instead of getting the default endpoint or gateway accessor every time a request is made, we should cache the result for some period of time since it is unlikely to change.

MarquessV:
Resolves #336.

Two open questions:

- [ ] Is one hour a reasonable TTL? (Since this is client-side and the addresses should be very static, do we even need a TTL?)
- [ ] Is it important to split the test in `mocked_qpu` into two tests, one for `default_endpoint` and one for `accessors`?

### DIFF
--- a/crates/lib/Cargo.toml
+++ b/crates/lib/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "qcs"
 description = "High level interface for running Quil on a QPU"
-version = "0.15.3-rc.0"
+version = "0.15.3-rc.1"
 edition = "2018"
 license = "Apache-2.0"
 repository = "https://github.com/rigetti/qcs-sdk-rust"
--- a/crates/lib/Cargo.toml
+++ b/crates/lib/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "qcs"
 description = "High level interface for running Quil on a QPU"
-version = "0.15.3-rc.1"
+version = "0.15.3"
 edition = "2018"
 license = "Apache-2.0"
 repository = "https://github.com/rigetti/qcs-sdk-rust"
--- a/crates/lib/Cargo.toml
+++ b/crates/lib/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "qcs"
 description = "High level interface for running Quil on a QPU"
-version = "0.15.3"
+version = "0.15.4-rc.0"
 edition = "2018"
 license = "Apache-2.0"
 repository = "https://github.com/rigetti/qcs-sdk-rust"
--- a/crates/lib/Cargo.toml
+++ b/crates/lib/Cargo.toml
@@ -16,6 +16,7 @@ tracing-config = ["tracing", "qcs-api-client-common/tracing-config", "qcs-api-cl
 otel-tracing = ["tracing-config", "qcs-api-client-grpc/otel-tracing", "qcs-api-client-openapi/otel-tracing"]
 
 [dependencies]
+cached = "0.44.0"
 dirs = "5.0.0"
 enum-as-inner = "0.5.1"
 futures = "0.3.24"
@@ -43,9 +44,8 @@ uuid = { version = "1.2.1", features = ["v4"] }
 tonic = { version = "0.9.2", features = ["tls", "tls-roots"] }
 zmq = { version = "0.10.0" }
 itertools = "0.11.0"
-rstest = "0.17.0"
-insta = "1.29.0"
 derive_builder = "0.12.0"
+once = "0.3.4"
 
 [dev-dependencies]
 erased-serde = "0.3.23"
@@ -60,6 +60,8 @@ warp = { version = "0.3.3", default-features = false }
 regex = "1.7.0"
 test-case = "3.1.0"
 tracing-subscriber = "0.3.17"
+rstest = "0.17.0"
+insta = "1.29.0"
 
 [build-dependencies]
 built = "0.6.1"
--- a/crates/lib/Cargo.toml
+++ b/crates/lib/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "qcs"
 description = "High level interface for running Quil on a QPU"
-version = "0.15.4-rc.0"
+version = "0.15.4"
 edition = "2018"
 license = "Apache-2.0"
 repository = "https://github.com/rigetti/qcs-sdk-rust"
--- a/crates/lib/Cargo.toml
+++ b/crates/lib/Cargo.toml
@@ -45,7 +45,6 @@ tonic = { version = "0.9.2", features = ["tls", "tls-roots"] }
 zmq = { version = "0.10.0" }
 itertools = "0.11.0"
 derive_builder = "0.12.0"
-once = "0.3.4"
 
 [dev-dependencies]
 erased-serde = "0.3.23"
--- a/crates/lib/src/qpu/api.rs
+++ b/crates/lib/src/qpu/api.rs
@@ -3,6 +3,7 @@
 
 use std::{fmt, time::Duration};
 
+use cached::proc_macro::cached;
 use derive_builder::Builder;
 use qcs_api_client_common::{configuration::RefreshError, ClientConfiguration};
 pub use qcs_api_client_grpc::channel::Error as GrpcError;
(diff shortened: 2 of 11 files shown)


kalzoo:
This looks like the right approach. I'd further split the new function into `get_accessor` and `get_accessor_[from|with]_cache` to allow for uncached access if/when needed, but given that it's a private function that's only stylistic at this point

crates/lib/src/qpu/api.rs
```diff
@@ -356,6 +326,55 @@ impl ExecutionOptions {
...
+
+        let accessor = accessors
+            .accessors
+            .into_iter()
+            .filter(|acc| {
+                acc.live
+                // `as_deref` needed to work around the `Option<Box<_>>` type.
+                && acc.access_type.as_deref() == Some(&QuantumProcessorAccessorType::GatewayV1)
+            })
+            .min_by_key(|acc| acc.rank.unwrap_or(i64::MAX));
```
BatmanAoD, line 363: Question for my understanding: how are these `rank` values set, and do we actually have situations where multiple accessors are returned from the API?

crates/lib/src/qpu/api.rs
```diff
@@ -356,6 +326,55 @@ impl ExecutionOptions {
     }
 }
 
+#[cached(
+    result = true,
+    time = 60,
```
BatmanAoD, line 331: Why a TTL? Do we expect the accessor to change?

MarquessV: I threw that out arbitrarily when I put together this draft, I think accessors would very rarely change, if at all. I suppose a reasonable TTL would be a minimum duration we would want to maintain an old accessor after putting up a new one, that way users don't experience a disruption when we swap them out.

crates/lib/src/qpu/api.rs
kalzoo: They are set arbitrarily and that does not need to be known to the user. We do anticipate multiple values in a response

https://docs.api.qcs.rigetti.com/#operation/ListQuantumProcessorAccessors

crates/lib/Cargo.toml
```diff
@@ -43,9 +44,8 @@ uuid = { version = "1.2.1", features = ["v4"] }
 tonic = { version = "0.9.2", features = ["tls", "tls-roots"] }
 zmq = { version = "0.10.0" }
 itertools = "0.11.0"
-rstest = "0.17.0"
-insta = "1.29.0"
 derive_builder = "0.12.0"
+once = "0.3.4"
```
BatmanAoD, line 48: Is this actually used?

BatmanAoD: Confirmed it is not used

crates/lib/src/qpu/api.rs
```diff
@@ -356,6 +326,56 @@ impl ExecutionOptions {
     }
 }
 
+#[cached(
+    result = true,
+    time = 3600,
```
BatmanAoD, line 331: @kalzoo How often would an accessor change?

I think ideally we'd just keep the accessor cached indefinitely but force a cache update any time we can't connect to a cached accessor. I don't think that can be easily done without losing the benefit of this convenient proc-macro, though.

crates/lib/tests/mocked_qpu.rs
```diff
@@ -167,8 +208,27 @@ mod mock_qcs {
...
+            .and(warp::get())
+            .map(|| {
+                ACCESSORS_CALL_COUNT.fetch_add(1, SeqCst);
+                let rsp = ListQuantumProcessorAccessorsResponse {
+                    accessors: vec![QuantumProcessorAccessor {
+                        access_type: Some(Box::new(QuantumProcessorAccessorType::GatewayV1)),
+                        live: true,
+                        rank: Some(0),
+                        id: Some(QPU_ID.to_string()),
+                        // TODO: does the content of this string matter?
```
notmgsk, line 222: Not unless the tests are going to use it (I don't think they do).

BatmanAoD: Oops, I forgot to double-check this; thanks for the comment. Apparently it's actually being used to ensure the test hits this mock endpoint; if I pick a nonexistent URL, I get this:

```
message: "service call failed with error: transport error"
```

crates/lib/src/qpu/api.rs
MarquessV: It looks like the proc macro generates a function that can be used to refresh the cache. We might be able to utilize it to recover from a connection error? That might be overly complicated to try and do at this point, but it might be the most robust way of doing it. I can't link to the specific example, but it's [one of the last ones in these docs](https://docs.rs/cached/0.44.0/cached/proc_macro/index.html).

As I write this, I wonder that, even if we could transparently switch accessors in the client, I wonder if it will still be disruptive to the user if they have jobs queued on the old accessor that haven't been processed yet. So maybe we shouldn't even bother trying to switch accessors in a given session?

crates/lib/src/qpu/api.rs
```diff
@@ -307,55 +308,96 @@ impl ExecutionOptions {
...
+        .ok_or_else(|| QpuApiError::GatewayNotFound(quantum_processor_id.to_string()))
+}
+
+#[cached(
+    result = true,
+    time = 3600,
+    key = "String",
+    convert = r"{ String::from(quantum_processor_id)}"
+)]
+async fn get_default_endpoint_with_cache(
```
MarquessV, line 379: Thanks for adding this one too! 🔥

MarquessV (commented):
I can't approve since I'm the original author, and still interested in @kalzoo's take on the TTL for the cache, but looks good to me!

crates/lib/src/qpu/api.rs
BatmanAoD: Oh, interesting, it just generates a function ending with `_prime_cache` in the same namespace? _Surely_ that's documented somewhere, not just implicitly in those examples... but yes, that seems like a decent approach in case we end up needing manual control.

I think I need to better understand, conceptually, what an "accessor" is here. Kalan confirmed above that we do indeed expect to have multiple accessors for one QPU ID, but what does that _mean_? Are they all "pointing" to the same ConServ instance, or not?

If not, then you're right that changing accessors would be disruptive; my testing in this branch showed that, without caching, the `accessor` endpoint is hit twice for each run, presumably once to submit the job and once to get the results.

BatmanAoD: (I've checked Confluence a few times to be sure I'm not missing something obvious, and...well, if I'm missing something, it's definitely not _obvious_. And the API documentation itself doesn't really indicate to me what "accessors" are conceptually.)

kalzoo: You're right that this is lightly documented, and that stands to be improved in our public docs.

This is about service discovery, and accessors are the advertised means of accessing the particular QPU. They're not pointing to the same service, but to the same QPU (architecture). This is one example of how there may be many:

* one direct access, gRPC API (what we have today)
* an alternative future ™️ API if and when we roll that out
* one or more gateways, where each offers lowest-latency access from different geographical areas (where we can accelerate access via some private backbone as an alternative to public internet - imagine AUS or UK gateways for a US QPU)
* simulated endpoints where that architecture is available but backed by something that isn't a QPU (and of course, marked as such)

Today, accessors are simple: gateway and direct access, that's it (in practice). But the mechanism is meant to be general enough to serve its purpose well into the future, and let us evolve access to our QPUs without breaking changes or new API routes.

kalzoo: As far as the TTL length, that's a good point that changing the accessor, even from one valid-and-working accessor to another, may break a client's workflow.

We can't cache indefinitely because there may be long-running processes which rely on this code, and those would never pick up changes without a bounce (not an ideal requirement).

I'd err on the side of shorter right now, imagining the scenario where we publish data that results in a broken client and then need to roll back or fix. How long of downtime would we want to accept in that case? Some clients may fail and be restarted (thus flushing the cache), but any that have retries built in would not.

If we can apply Kyle's suggestion to flush the cache in case of errors, then it becomes easier to accept an hour's wait. However, that does add complexity, either in that we force a request retry (which may not be idempotent) or we require the user to do so.

BatmanAoD: I've gone back to 1 minute. I'll open a separate ticket for ensuring that clients are always able to get results even if accessors change.

BatmanAoD: https://github.com/rigetti/qcs-sdk-rust/issues/341

BatmanAoD (approved):
Approving because Marquess commented approval and Kalan provided feedback on TTL.

PR #342, opened 2023-08-10
feat(python): Constructors, iterators, and other utility methods have been added to make working with `ResultData`, `RegisterMap`, and others easier.

MarquessV:
This supports pyQuil#1630 by making it easier to construct and iterate over the result data types.

This does not change the return shape of `qpu::api::retrieve_results`. The target would be to construct a `QpuResultData`, but the data we need simply isn't available in that function to build it.

We could alternatively use the `Executable` struct, which has it's own workflow that already returns `QpuResultData`, but pyQuil's `QAM` is already workflow based and it's quite clunky to try and use one in the other.

I think it would be good to reconcile this before a v1.0 release, but this gets us passed the issue for now. In the end, this actually doesn't shove off _too_ much to pyQuil.

## TODO:

- [x] Add stubs after API approval
- [x] ~Figure out if we can better expose types for `QPU/QVMResultData.asdict()`~ Took a different approach.


crates/lib/src/execution_data.rs
```diff
@@ -83,7 +83,7 @@ pub enum RegisterMatrix {
 /// register.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[repr(transparent)]
-pub struct RegisterMap(HashMap<String, RegisterMatrix>);
+pub struct RegisterMap(pub HashMap<String, RegisterMatrix>);
```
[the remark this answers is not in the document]
Shadow53: This seems fine as long as it's not possible to create invalid state with direct access to the inner map

crates/python/src/execution_data.rs
```diff
@@ -58,14 +62,8 @@ impl PyExecutionData {
...
 
-// From gRPC
-py_wrap_data_struct! {
-    PyReadoutValues(ReadoutValues) as "ReadoutValues" {
-        values: Option<Values> => Option<PyReadoutValuesValues>
-    }
-}
-
 py_wrap_type! {
+    #[pyo3(mapping)]
```
Shadow53, line 66: Was not aware this was a thing!

crates/python/src/execution_data.rs
```diff
@@ -75,6 +73,14 @@ py_wrap_type! {
 
 #[pymethods]
 impl PyRegisterMatrix {
+    fn to_ndarray(&self, py: Python<'_>) -> Py<PyAny> {
+        self.as_integer(py)
+            .map(|array| array.into_py(py))
+            .or(self.as_real(py).map(|array| array.into_py(py)))
+            .or(self.as_complex(py).map(|array| array.into_py(py)))
+            .expect("A RegisterMatrix can't be any other type.")
+    }
```
Shadow53, line 82: Is there a way to write this that does not involve possible panics?

crates/python/src/execution_data.rs
```diff
@@ -153,9 +159,109 @@ impl PyRegisterMatrix {
...
+            .ok_or(PyKeyError::new_err(format!(
+                "Key {item} not found in RegisterMap"
+            )))
+    }
+
+    pub fn __iter__(&self, py: Python<'_>) -> PyResult<Py<PyRegisterMapKeysIter>> {
+        Py::new(
+            py,
+            PyRegisterMapKeysIter {
+                inner: self.as_inner().0.clone().into_iter(),
```
Shadow53, line 187: Curious if it would be any different to do this:

```suggestion
                inner: self.as_inner().0.iter().cloned(),
```

Shadow53: Also, shouldn't this be just the keys?

```suggestion
                inner: self.as_inner().0.keys().cloned(),
```

crates/python/src/execution_data.rs
```diff
@@ -153,9 +159,109 @@ impl PyRegisterMatrix {
...
+
+    pub fn keys(&self, py: Python<'_>) -> PyResult<Py<PyRegisterMapKeysIter>> {
+        self.__iter__(py)
+    }
+
+    pub fn values(&self, py: Python<'_>) -> PyResult<Py<PyRegisterMapValuesIter>> {
+        Py::new(
+            py,
+            PyRegisterMapValuesIter {
+                inner: self.as_inner().0.clone().into_iter(),
```
Shadow53, line 200: Similarly, shouldn't this just be values?

```suggestion
                inner: self.as_inner().0.values().cloned(),
```

crates/python/src/execution_data.rs
```diff
@@ -153,9 +159,109 @@ impl PyRegisterMatrix {
...
+            py,
+            PyRegisterMapItemsIter {
+                inner: self.as_inner().0.clone().into_iter(),
+            },
+        )
+    }
+
+    pub fn get(&self, key: &str, default: Option<PyRegisterMatrix>) -> Option<PyRegisterMatrix> {
+        self.__getitem__(key).ok().or(default)
+    }
```
Shadow53, line 216: Does Python not provide this?

crates/python/src/execution_data.rs
Shadow53: Alternatively, why not just use the [pyo3 provided](https://docs.rs/pyo3/latest/pyo3/types/struct.PyIterator.html) `PyIterator`? Then you wouldn't need to define the various iterators below.

crates/python/src/execution_data.rs
MarquessV: We could return a `PyResult` to raise a different exception. pyo3 already raises a [PanicException](https://pyo3.rs/main/doc/pyo3/panic/struct.panicexception) when Rust code called from Python panic. It will usually unwind the whole stack, which I think is what we want if this block somehow fails the expectation.

crates/python/src/execution_data.rs
MarquessV: Maybe I'm missing something, but I think `PyIterator` is just the Rust native type for interacting with Python iterators. That is, if you call some Python code from Rust that returns an iterator, a `PyIterator` is what you would want to put it in to work with it. I see no way to build or return one from a Rust iterator. This setup is based on  the [docs on iterable objects in pyo3](https://pyo3.rs/v0.19.2/class/protocols.html?highlight=iter#iterable-objects).

The problem with using `cloned()`, `keys()`, and `values()` is that they are all an iterator type based on a reference to the `HashMap`, which means the iterator struct would require lifetimes, which isn't compatible with `#[pyclass]`.

I'll add a comment to the iterator structs to help clarify.

crates/lib/src/execution_data.rs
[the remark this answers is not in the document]
MarquessV: Yeah, technically you can't since the RegisterMatrix is built on `ndarray`, so they can't be modified to be non-rectangular, which is what would be invalid state.

crates/python/src/execution_data.rs
MarquessV: Unfortunately, no. Unlike inheriting `abc.Mapping` in Python, [you need to implement a lot more](https://pyo3.rs/v0.19.2/class/protocols#mapping--sequence-types) to get the C API to recognize a type as a container type and get these methods. The full list unfortunately includes mutable methods, which I don't think we want to support here. I believe pyo3 is planning on supporting read-only container types in the future, but not yet.

The `#[pyo3(mapping)]` is really only insurance against pyo3 implementing default behavior for sequences that would be incompatible with a mapping type.

crates/python/src/qpu/result_data.rs
```diff
@@ -43,4 +43,27 @@ impl PyQpuResultData {
...
+                .readout_values()
+                .iter()
+                .map(|(register, values)| {
+                    (
+                        register.to_string(),
+                        match values {
+                            ReadoutValues::Integer(values) => values.to_object(py),
+                            ReadoutValues::Real(values) => values.to_object(py),
+                            ReadoutValues::Complex(values) => values.to_object(py),
+                        },
```
MarquessV, line 62: Is there a more concise way of doing this? Only thing I could think of is some kind of trait on `ReadoutValues`, but then we'd bleed python functionality into the Rust crate.

crates/python/src/qpu/result_data.rs
```diff
@@ -43,4 +43,27 @@ impl PyQpuResultData {
     fn readout_values(&self, py: Python<'_>) -> PyResult<HashMap<String, PyReadoutValues>> {
         self.as_inner().readout_values().to_python(py)
     }
+
+    fn asdict(&self, py: Python<'_>) -> PyResult<Py<PyDict>> {
```
MarquessV, line 47: This makes working with the type on Python much simpler. It is a little bit fragile in that we have to statically name the keys, but the ergonomics are nice and let the user get at an object that works well as a Python type when you don't need/want the extra type safety from the wrapping type.

crates/python/src/qpu/result_data.rs
```diff
@@ -43,4 +43,40 @@ impl PyQpuResultData {
...
+}
+
+#[derive(Debug)]
+#[pyclass(name = "RawQPUReadoutData")]
+pub struct RawQpuReadoutData {
+    #[pyo3(get)]
+    mappings: HashMap<String, String>,
+    #[pyo3(get)]
+    readout_values: HashMap<String, Py<PyList>>,
+}
```
MarquessV, line 76: The idea here is to make it easier to get at what Python users likely want from `QPUResultData`. The type safety provided by `QpuResultData` and `ReadoutValues` is nice, but only if you care about it. If you don't, it's cumbersome to have to call `inner()` on every instance of `ReadoutValues`.

crates/python/src/execution_data.rs
```diff
@@ -46,35 +51,53 @@ py_wrap_data_struct! {
...
+            duration: match duration {
+                None => None,
+                Some(delta) => Some(
+                    delta
+                        .as_ref(py)
+                        .call_method0("total_seconds")
+                        .map(|result| result.extract::<f64>())?
+                        .map(Duration::from_secs_f64)?,
+                ),
+            },
```
Shadow53, line 75: Suggesting mostly as something that would maybe trip up clippy.

```suggestion
            duration: duration.map(|delta| {
                delta
                    .as_ref(py)
                    .call_method0("total_seconds")
                    .map(|result| result.extract::<f64>())?
                    .map(Duration::from_secs_f64)
            }).transpose()?,
```

crates/python/src/execution_data.rs
```diff
@@ -153,9 +176,114 @@ impl PyRegisterMatrix {
...
             .map(PyRegisterMatrix::from)
     }
+
+    pub fn __len__(&self) -> usize {
+        self.as_inner().0.len()
+    }
+
+    pub fn __contains__(&self, key: String) -> bool {
+        self.as_inner().0.contains_key(&key)
+    }
```
Shadow53, line 191:
```suggestion
    pub fn __contains__(&self, key: &str) -> bool {
        self.as_inner().0.contains_key(key)
    }
```

crates/python/src/qpu/result_data.rs
Shadow53: This is honestly pretty concise, imo.

crates/python/src/qpu/result_data.rs
Shadow53: Maybe put this in a doc comment?

Shadow53 (approved):
A few minor suggestions, but looks good overall

crates/python/src/execution_data.rs
BatmanAoD: Why is the `transpose()` preferable to having the `?` after the last line of the closure?

I agree with keeping the `duration.map(|delta|...)` rather than the explicit `match`.

crates/python/src/execution_data.rs
```diff
@@ -46,35 +51,53 @@ py_wrap_data_struct! {
...
 }
+impl_repr!(PyRegisterMatrix);
 
 #[pymethods]
 impl PyRegisterMatrix {
+    fn to_ndarray(&self, py: Python<'_>) -> Py<PyAny> {
+        self.as_integer(py)
+            .map(|array| array.into_py(py))
+            .or(self.as_real(py).map(|array| array.into_py(py)))
+            .or(self.as_complex(py).map(|array| array.into_py(py)))
```
BatmanAoD, line 97: Is it possible to use `match` to determine whether this is integer/real/complex, rather than calling `as_...()` for each type?

If not, it still might be nicer to use a sequence of `if let`s.

crates/python/src/qpu/result_data.rs
BatmanAoD: Do the Python values lose the numeric-type information?

crates/python/qcs_sdk/qpu/__init__.pyi
```diff
@@ -69,6 +75,27 @@ class QPUResultData:
...
+    @property
+    def mapping(self) -> Dict[str, str]:
+        """
+        Get the mappings of a memory region (ie. "ro[0]") to it's key name in readout_values
+        """
+        ...
+    @property
+    def readout_values(self) -> Dict[str, Union[List[int], List[float], List[complex]]]:
+        """
+        Get the mappings of a readout values identifier (ie. "q0") to a list of those readout values
```
BatmanAoD, line 96: I'm guessing this is something I could learn by reading more about quil, but what are "readout values identifiers"? Are they something the quil program defines, or something internal?

Also, what are the "key names"? Is that internal, or something quil defines?

Are these two maps related/overlapping? I noticed in the tests you have `"a"` as a key for both maps, but based on the comments, it looks like the keys have different meanings.

crates/python/src/grpc/models/controller.rs
```diff
@@ -14,6 +15,27 @@ py_wrap_data_struct! {
...
 
+#[pymethods]
+impl PyReadoutValues {
+    pub fn as_ndarray(&self, py: Python<'_>) -> PyResult<Option<PyObject>> {
+        match &self.as_inner().values {
+            None => Ok(None),
+            Some(Values::IntegerValues(ints)) => {
+                Ok(Some(PyArray::from_slice(py, &ints.values).to_object(py)))
+            }
+            Some(Values::ComplexValues(complex)) => Ok(Some(
```
BatmanAoD, line 26: Why does `PyReadoutValues` only support integer or complex numbers, while the raw version supports reals as well? Do we just return reals as complex numbers with a real-part of 0?

crates/python/src/qpu/result_data.rs
BatmanAoD: ...oh, I see, `values` has a different `Item` type depending on the numeric type it represents.

crates/python/src/qvm/mod.rs
```diff
@@ -16,24 +17,33 @@ mod api;
...
-    classes: [PyQvmResultData, PyQvmOptions],
+    classes: [PyQvmResultData, PyQvmOptions, RawQvmReadoutData],
     errors: [QVMError],
     funcs: [py_run, py_run_async],
     submodules: [
         "api": api::init_submodule
     ],
 }
 
+wrap_error!(RustQvmError(qcs::qvm::Error));
```
BatmanAoD, line 29: Does moving these after the `create_init_submodule` change anything?

BatmanAoD (approved):
This looks reasonable to me. I've added lots of questions for my own edification, but I'm not sure any of them indicate a lack of sufficient comments or documentation so much as my own lack of context.

crates/python/src/qpu/result_data.rs
MarquessV: I removed this in favor of a different approach that flattens the object into a struct that can be typed.

crates/python/src/execution_data.rs
MarquessV: Yeah, coming back to this fresh, I re-wrote it to match on the inner type.

crates/python/src/execution_data.rs
MarquessV: I rewrote it as suggested, but left off the `transpose`

crates/python/qcs_sdk/qpu/__init__.pyi
MarquessV: The Quil spec wouldn't answer this question exactly, but as background it is important to understand [Quil-T calibrations](https://quil-lang.github.io/#12-6Defining-Calibrations). They essentially let you map a gate, parameters, and qubits to a sequence of analog control instructions. These instructions offer lower level control than a typical circuit-based program. The default calibrations for a QPU can be retrieved by running `qc.compiler.get_calibration_program()` in pyQuil.

Among other things, these are used to calibrate `MEASURE` instructions on a QPU. So when a user writes something like `MEASURE 0 ro[0]`, the QPU will run a calibration that can actually perform and process the measure. That calibration will write the value to a buffer named something like `q0`. The service maintains a mapping of what readout buffers map to which memory regions in the program, and that mapping gets returned. The key names themselves are defined in the calibrations, so they are defined in Quil, but we control what they are.

The two maps are related. In the above example, a user might want to see the readout values written to `ro[0]` in their program. They would get `mappings["ro[0]"]` as the key in `readout_values` to get the actual values.

crates/python/src/grpc/models/controller.rs
MarquessV: Ah, there are two `PyReadoutValues`. This one is a raw mapping to the controller service gRPC model of readout values, which only supports `int` and `complex`. I actually don't think this method is even needed any more, as I'm using the higer level `PyReadoutValues` that supports `real` as well. The only reason for the difference between the model and the SDK version is for future proofing. If we ever return real numbered data from the QPU, we don't have to make a breaking change.

crates/python/src/qvm/mod.rs
MarquessV: I don't think so. Perhaps @Shadow53 moved them for stylistic purposes?

crates/python/qcs_sdk/qpu/__init__.pyi
kalzoo: > The service maintains a mapping of what readout buffers map to which memory regions in the program, and that mapping gets returned

all true but I'd add "...from compilation" to clarify. These mappings are set with the PRAGMA LOAD-MEMORY instruction.

A little more context on this - the data that most users care about (classified qubit values) is actually generated in a post-processing step after execution, and the stream processor is constructed like a tree of processor nodes. Each node has a key, and the "sink" nodes in the graph have the IDs seen as the keys of this readout map as returned from execution. In fact the entire graph is constructed by the user within the Quil program using PRAGMA FILTER-NODE instructions (which themselves come as default in the DEFCAL blocks).

These are great questions from @BatmanAoD and some that many users will ask. This information should get captured in our documentation, likely on gitbook since it's specific to our backend and not to any one client.

kalzoo (approved):

PR #346, opened 2023-08-21
new: allow for additional "clients" for quilc and QVM

notmgsk:
# Quilc

The problem I am trying to solve here is we would like to introduce support for `libquil-sys` as a way of using `quilc`. Currently we can only use `quilc` via RPCQ. This MR reworks the `qcs::compiler::quilc::compile_program()` compilation entry point such that it takes in a type that implements the (new) `qcs::compiler::quilc::Client` trait. Then, to add new clients, we just implement this trait for the type.

On the Python side, I have introduced a new type `RPCQClient` which holds a `qcs::compiler::quilc::rpcq::Client`. Calls to `compile_program()` will use this client rather than creating a new one per request. To allow for additional clients to be passed from Python to Rust, there is a enum `QuilcClient` which implements `pyo3::FromPyObject`. When we want to add another client, we will simply extend that enum. Adding new clients is not a breaking change.

# QVM

There are similar changes to provide a QVM client. 

## TODO
- [x] `rpcq::Client` has a custom implementation for `Clone` which creates a new client (and ZMQ socket). This is inefficient -- we should reuse the client.
- [x] Python documentation and type stubs


crates/lib/src/compiler/rpcq.rs
```diff
@@ -76,27 +106,66 @@ impl Client {
...
     Deserialization(#[from] rmp_serde::decode::Error),
+    /// Response ID did not match request ID
     #[error("Response ID did not match request ID")]
     ResponseIdMismatch,
+    /// Server responded with an error message
     #[error("Received error message from server: {0}")]
     Response(String),
+    /// Error occurred when trying to lock the ZMQ socket
+    #[error("Could not lock RPCQ client: {0}")]
+    ZmqSocketLock(String),
```
MarquessV, line 168: Could this be `#[from] zmq::WhateverTheZmqSocketLockErrorIs`?

crates/lib/src/compiler/quilc.rs
```diff
@@ -17,6 +16,18 @@ use crate::client::Qcs;
 /// Number of seconds to wait before timing out.
 pub const DEFAULT_COMPILER_TIMEOUT: f64 = 30.0;
 
+/// The Quilc compiler methods
+pub trait Client {
+    /// Compile the program `quil` for the given target device `isa`
+    /// with the compilation options `options`.
+    fn compile_program(
```
MarquessV, line 23: Whats the plan for the other quilc requests? Will those be supported by `libquil`?

crates/lib/src/compiler/rpcq.rs
```diff
@@ -1,26 +1,52 @@
...
+    socket: Arc<Mutex<Socket>>,
+}
+
+impl Clone for Client {
+    fn clone(&self) -> Self {
+        Self {
+            endpoint: self.endpoint.clone(),
+            socket: self.socket.clone(),
+        }
+    }
```
MarquessV, line 30: Is this implementation different from `#[derive(Clone)]`? What exactly is the ineficciency you mention in the PR description? Since `socket` is now wrapped in an `Arc`, we are just creating a new reference to it, not initializing a new one, right? That seems like what we would want, but I might be missing something.

notmgsk: Yeah. Previously a ZMQ client was created for every request to Quilc. This way we can have a single client.

crates/lib/src/compiler/rpcq.rs
notmgsk: It is possible. The error type is `PoisonError<Guard<'a, Socket>` and thus requires introducing lifetimes wherever the `Error` enum is used. Not fundamentally opposed to that... it just seemed heavy-handed.

crates/lib/src/compiler/quilc.rs
notmgsk: Yes! Working on it.

crates/lib/src/compiler/rpcq.rs
MarquessV: Ah yeah, that doesn't seem worth it to preserve the error type.

crates/lib/src/executable.rs
```diff
@@ -119,9 +121,10 @@ impl<'executable> Executable<'executable, '_> {
             params: Parameters::new(),
             compile_with_quilc: true,
             compiler_options: CompilerOpts::default(),
-            client: None,
             qpu: None,
             qvm: None,
+            qcs_client: None,
+            quilc_client: None,
```
MarquessV, line 127: If we use the existence of the client to determine whether or not to compile with quilc, should we remove the `compile_with_quilc` property from the `Executable` struct? Also, compiling with quilc was the default, so should we instantiate a `quilc_client` by default?

crates/lib/src/executable.rs
```diff
@@ -301,11 +315,16 @@ impl Executable<'_, '_> {
...
-    /// If set, the Executable will be compiled using `quilc` prior to compilation on QCS. If not set, the program
-    /// is treated as native quil and will not be sent to `quilc`.
+    /// Set the client used for compilation.
+    ///
+    /// To disable compilation, set this to `None`.
     #[must_use]
-    pub fn compile_with_quilc(mut self, compile: bool) -> Self {
-        self.compile_with_quilc = compile;
+    #[allow(trivial_casts)]
+    pub fn quilc_client<C: quilc::Client + Send + Sync + 'static>(
```
jselig-rigetti, line 323: Maybe `with_quilc_client`?

crates/python/src/qvm/mod.rs
```diff
@@ -24,14 +24,32 @@ py_wrap_type! {
...
     funcs: [py_run, py_run_async],
     submodules: [
         "api": api::init_submodule
     ],
 }
 
+#[pyo3::pyclass]
+#[pyo3(name = "QVMHTTPClient")]
+#[derive(Debug, Clone)]
+pub struct PyQvmHttpClient(pub qvm::http::HttpClient);
```
Shadow53, line 38: Same thing here with `py_wrap_struct`

Shadow53 (commented):
Some suggestions, but overall looks good.

crates/lib/src/executable.rs
```diff
@@ -42,7 +43,9 @@ use quil_rs::program::ProgramError;
 /// #[tokio::main]
 /// async fn main() {
 ///     use std::num::NonZeroU16;
-///     let mut result = Executable::from_quil(PROGRAM).with_client(Qcs::default()).with_shots(NonZeroU16::new(4).unwrap()).execute_on_qvm().await.unwrap();
+///     use qcs::qvm;
+///     let qvm_client = qvm::http::HttpClient::new(Qcs::load().await.get_config().qvm_url().to_string());
```
MarquessV, line 47: If this is the way we want people to initialize the client, should we offer a `from_qcs` constructor?

crates/python/tests/conftest.py
```diff
@@ -89,3 +91,13 @@ def bell_program() -> str:
...
+
+
+@pytest.fixture
+def qvm_http_client() -> QVMHTTPClient:
+    return QVMHTTPClient("http://localhost:5000")
+
+
+@pytest.fixture
+def quilc_rpcq_client() -> RPCQClient:
+    return RPCQClient("tcp://localhost:5555")
```
MarquessV, line 103: For configurability/consistency, might be better to initialize these fixtures with a QCS client, that way they aren't hardcoded.

MarquessV (commented):
A few comments, but this is looking good

kalzoo (commented):
This looks good, and an improvement. A couple small comments but otherwise l'd be happy to merge it. Since it's a breaking change, it should be prefixed to match.

We may want to consider if we want to take it one or two steps further, to break the API big and once:

1. Should the `Client` trait itself actually be defined in libquilc-sys? Then it could `impl Client for InProcessQuilcClient` and maybe even `RpcqClient` (or Http for QVM) itself, and this SDK could just use those. Defining the interface for a `Client` is closer in principle to libquil than it is to QCS.
2. Not here, but down the road - we may want to generalize even further with a "quil compiler" trait which only has a couple of these methods and allows users to define their own preprocessor passes which don't use `quilc` at all, although `quilc` could offer one implementation of it.

Putting those together:

* If a user wanted to get at the particular tricks of `quilc` - especially the calls beyond `get_version_info` and `native_quil_to_quil` - then they would go right to the source, `libquil-sys`. That is sort of the defacto libquil client, via both bindings and `rpcq`.
* If they just wanted quil to be optimized, then they would build an object of higher abstraction here in the QCS SDK around `quilc`, itself a thin wrapper around those methods. They provide one or more quil processing passes which are called in sequence; if none are provided, that's just like disabling `quilc` today.

Describing this, it doesn't seem like much more code in net, just moving around what's here. What do you think?

crates/python/src/compiler/quilc.rs
```diff
@@ -99,17 +102,42 @@ impl PyTargetDevice {
     }
 }
 
+wrap_error!(RustRpcqError(rpcq::Error));
+py_wrap_error!(quilc, RustRpcqError, RpcqError, PyRuntimeError);
+
+#[pyclass]
+#[pyo3(name = "RPCQClient")]
+#[derive(Clone)]
+pub struct PyRpcqClient(pub rpcq::Client);
```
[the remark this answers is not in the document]
notmgsk: `py_wrap_type` does allow visibility specifiers for the wrapped type

crates/lib/src/executable.rs
```diff
@@ -301,11 +320,16 @@ impl Executable<'_, '_> {
...
+    /// Set the client used for compilation.
+    ///
+    /// To disable compilation, set this to `None`.
     #[must_use]
-    pub fn compile_with_quilc(mut self, compile: bool) -> Self {
-        self.compile_with_quilc = compile;
+    #[allow(trivial_casts)]
+    pub fn with_quilc_client<C: quilc::Client + Send + Sync + 'static>(
+        mut self,
+        client: Option<C>,
```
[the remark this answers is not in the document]
notmgsk: That's how you disable the use of quilc as part of the execution flow. I'm not in love with it (not in love with the general design of `Executable` or `Execution`) but it felt like the best compromise without drastically rewriting Exec*.

crates/lib/src/executable.rs
```diff
@@ -301,11 +320,16 @@ impl Executable<'_, '_> {
...
+    ) -> Self {
+        self.quilc_client = client.map(|c| Arc::new(c) as _);
```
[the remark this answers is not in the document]
notmgsk: I thought so too; alas
```
error[E0308]: mismatched types
   --> crates/lib/src/executable.rs:332:29
    |
328 |     pub fn with_quilc_client<C: quilc::Client + Send + Sync + 'static>(
    |                              - this type parameter
...
332 |         self.quilc_client = client.map(Arc::new);
    |         -----------------   ^^^^^^^^^^^^^^^^^^^^ expected `Option<Arc<dyn Client + Send + Sync>>`, found `Option<Arc<C>>`
    |         |
    |         expected due to the type of this binding
    |
    = note: expected enum `std::option::Option<Arc<(dyn quilc::Client + std::marker::Send + Sync + 'static)>>`
               found enum `std::option::Option<Arc<C>>`
    = help: type parameters must be constrained to match other types
    = note: for more information, visit https://doc.rust-lang.org/book/ch10-02-traits.html#traits-as-parameters
```

I don't fully understand _why_, but it is the case that the explicit cast `as _` needs to be there. I did ask the Rust discord but they were only able to tell me to use `as _`, not why. Stackoverflow might be better.

crates/lib/src/compiler/quilc.rs
```diff
@@ -302,12 +277,12 @@ impl Error {
...
 #[derive(Clone, Deserialize, Debug, PartialEq, PartialOrd)]
-struct QuilToNativeQuilResponse {
+pub(crate) struct QuilToNativeQuilResponse {
     /// The compiled program
-    quil: String,
+    pub(crate) quil: String,
     /// Metadata about the compiled program
     #[serde(default)]
-    metadata: Option<NativeQuilMetadata>,
+    pub(crate) metadata: Option<NativeQuilMetadata>,
```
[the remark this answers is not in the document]
notmgsk: It did occur to me that those types (and the QVM variants) might be better suited elsewhere, but hadn't thought about incorporating them into `libquil-sys`. That does make sense to me, though it does add another MR in this train. I can do it quickly and see if it feels like a benefit.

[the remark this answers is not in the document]
notmgsk: Actually, I think depending on libquil-sys at this point is probably more work than it is worth since it is very much still in flux and there are certainly going to be breaking changes there that might bubble up here.

crates/lib/src/qpu/execution.rs
```diff
@@ -123,36 +120,28 @@ impl<'a> Execution<'a> {
         shots: NonZeroU16,
         quantum_processor_id: Cow<'a, str>,
         client: Arc<Qcs>,
-        compile_with_quilc: bool,
+        quilc_client: Option<Arc<dyn quilc::Client + Send + Sync>>,
```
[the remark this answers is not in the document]
notmgsk: It's only used the once, though? I would agree if this type were used many times.

MarquessV:
Some CI issues, but pending fixes for those I'm good to merge this. The more we have to update the `Executable` API the more I want to re-design it, something to think about before the 1.0 release.

MarquessV (approved):

PR #348, opened 2023-08-22
Remove unused qcs-api dependency and remove Cargo.lock

### LINKED ISSUE #319: Don't track `Cargo.lock`
Since `qcs-sdk-rust` is primarily a library, [we shouldn't be tracking `Cargo.lock`](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries)

### LINKED ISSUE #347: Remove unused qcs-api dependency

notmgsk:
Closes #347 and closes #319


notmgsk:
There are more issues arising from `cargo-deny`:

* `Advisory: https://rustsec.org/advisories/RUSTSEC-2023-0052`. Not really sure what to do about that. `webpki` doesn't have a fix, and for some packages `webpki` is required.
* `found 2 duplicate entries for crate 'windows-sys'`. I would just change these to warnings.

@Shadow53 @MarquessV thoughts?

PR #351, opened 2023-09-05
feat(python): Build & publish wheels for Windows

MarquessV:
Builds and publishes wheels for x86 and x64 based windows systems. Created #352 to track creating Windows for ARM wheels when it's better supported.


MarquessV:
Successful run here: https://github.com/rigetti/qcs-sdk-rust/actions/runs/6102826202/job/16562116923?pr=351

crates/python/qcs_sdk/__init__.pyi
```diff
@@ -200,11 +226,11 @@ class RegisterMatrix:
     def to_real(self) -> NDArray[np.float64]: ...
     def to_complex(self) -> NDArray[np.complex128]: ...
     @staticmethod
-    def from_integer(inner: NDArray[np.int64]) -> "RegisterMatrix": ...
+    def from_integer(matrix: NDArray[np.int64]) -> "RegisterMatrix": ...
```
jselig-rigetti, line 229: I see per https://github.com/rigetti/qcs-sdk-rust/blob/55f28f7de80cfa18074429720d3b788d6bf91ebd/crates/python/src/execution_data.rs#L136 that these are implemented manually instead of with `pyo3` macros - is there a reason to use the keyword `matrix` here instead of aligning with the macro's use of `inner` ?

.github/workflows/release-python.yml
```diff
@@ -90,10 +89,45 @@ jobs:
...
+          architecture: ${{ matrix.target }} 
+      - name: Update rustup 
+        run: rustup self update 
+      - name: Install Rust toolchain 
+        uses: actions-rs/toolchain@v1 
+        with: 
+          toolchain: stable 
+          target: x86_64-pc-windows-msvc 
+          profile: minimal 
+          default: true
```
Shadow53, line 112: This is inconsistent with https://github.com/rigetti/quil-rs/pull/280#pullrequestreview-1614323964, which uses a different action to install a different toolchain.

Maybe it would be worth creating our own [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) to standardize how we do this?

crates/python/qcs_sdk/__init__.pyi
MarquessV: No good reason, just that `stubtest` flagged it now that `__text_signature__` is defined by default. Will change it to inner.

.github/workflows/release-python.yml
MarquessV: To close the loop here, I've reverted the action change in the quil-rs PR, it was only meant to be a temporary change for testing purposes.

notmgsk (approved):

PR #362, opened 2023-09-15
Add a public method to get a GRPC connection without specifying the API

BatmanAoD:
This enables using different protobuf types rather than the ones the SDK makes available.

### DIFF
--- a/crates/lib/src/qpu/api.rs
+++ b/crates/lib/src/qpu/api.rs
@@ -182,11 +182,11 @@ pub async fn retrieve_results(
         )
 }
 
-// The ExecutionOptions are really generic over all QPU operations.
 /// Options available when connecting to a QPU.
 ///
 /// Use [`Default`] to get a reasonable set of defaults, or start with [`QpuConnectionOptionsBuilder`]
 /// to build a custom set of options.
+// These are aliases because the ExecutionOptions are actually generic over all QPU operations.
 pub type QpuConnectionOptions = ExecutionOptions;
 /// Builder for setting up [`QpuConnectionOptions`].
 pub type QpuConnectionOptionsBuilder = ExecutionOptionsBuilder;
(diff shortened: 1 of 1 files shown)


crates/lib/src/qpu/api.rs
```diff
@@ -260,11 +269,25 @@ impl ExecutionOptions {
         }
     }
 
-    async fn get_controller_client(
+    /// Get a controller client for the given QPU ID.
+    pub async fn get_controller_client(
```
BatmanAoD, line 273: This doesn't necessarily need to be public, but I added `pub` just because it seems like a pretty reasonable thing to want, and with `get_qpu_grpc_connection` being public, there's no longer any private code being invoked internally, so this is basically just some boilerplate that the client could have written.

MarquessV: Yeah, agreed that it's reasonable.

crates/lib/src/qpu/api.rs
```diff
@@ -260,11 +269,25 @@ impl ExecutionOptions {
...
+        Ok(ControllerClient::new(service)
+            .max_decoding_message_size(MAX_DECODING_MESSAGE_SIZE_BYTES))
+    }
+
+    /// Get a GRPC connection to a QPU, without specifying the API to use.
+    pub async fn get_qpu_grpc_connection(
+        &self,
+        client: &Qcs,
+        quantum_processor_id: Option<&str>,
+    ) -> Result<RefreshService<Channel, ClientConfiguration>, QpuApiError> {
```
MarquessV, line 290: This is the heart of the change right, to add a public function that returns the wrapped `Channel`? Seems fine to me!

MarquessV (approved):
Some CI stuff to check, but this looks good to me!

crates/lib/src/qpu/api.rs
BatmanAoD: Yep!

PR #370, opened 2023-09-28
ci(python): Don't publish wheels automatically for pre-releases

MarquessV:
This stops wheels from publishing on pre-releases automatically. A pre-release will still always be created on merge to main, but only a sdist will be published to PyPI. The Publish workflow can be triggered manually with the `publishWheels` option if we decide we want to create wheels for a particular pre-release.

I validated these changes as best as I could with [nektos/act](https://github.com/nektos/act):

A new pre-release being published only results in a sdist being published
```
act release -n -e .github/test_events/python_prerelease.json -W .github/workflows/release-python.yml
```

A manual run by default only publishes a sdist
```
act workflow_dispatch -n -W .github/workflows/release-python.yml
```

A full release being published results in a sdist and wheels being published
```
act release -n -e .github/test_events/python_release.json -W .github/workflows/release-python.yml
```

A manual run with `publishWheels` set to true publishes a sdist and wheels
```
act workflow_dispatch -n --input publishWheels=true -W .github/workflows/release-python.yml
```


.github/test_events/README.md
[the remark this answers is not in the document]
Shadow53: I think it's fine as long as `test_events` isn't a special folder name

.github/workflows/release-python.yml
```diff
@@ -5,17 +5,27 @@ on:
...
-    if: ${{ github.event_name == 'workflow_dispatch' || startsWith(github.event.release.tag_name, 'python/v') }}
+    if: github.event_name == 'workflow_dispatch' || startsWith(github.event.release.tag_name, 'python/v')
     runs-on: ubuntu-latest
     steps:
       - run: echo "Release tag starts with python/v, proceeding with release"
+  should-publish-wheels:
+    if: (github.event_name == 'workflow_dispatch' && inputs.publishWheels) || (github.event_name == 'release' && !github.event.release.prerelease)
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Publishing wheels (${{ github.event_name == 'release' }} && ${{ !github.event.release.prerelease }}) ${{ (github.event_name == 'workflow_dispatch' && inputs.publishWheels) || (github.event_name == 'release' && !github.event.release.prerelease) }})"
```
Shadow53, line 24: Is including the condition in the `echo` necessary? I would expect that the `echo` only happens if the condition is true.

.github/workflows/release-python.yml
```diff
@@ -155,13 +166,20 @@ jobs:
...
     runs-on: ubuntu-latest
+    # `needs` forces this job to wait until all specified jobs
+    # are finished to run. Typically, those jobs would all have
+    # to be successful, but when combined with `if: always()`,
+    # this job is allowed to run after all the needed jobs
+    # finish, regardless of their outcome. In this case, we
+    # still make sure that at least a source distribution
+    # can be published.
     needs: [ macos, linux, windows, sdist ]
+    if: always() && needs.sdist.result == 'success'
```
Shadow53, line 179: Is the `always()` necessary if you have the `needs.sdist.result` bit? I can see it having a side effect that is needed, but from a boolean expression standpoint it seems pointless to have what is effectively `true &&`

Shadow53 (approved):
A few questions, but nothing that should block merge.

.github/workflows/release-python.yml
MarquessV: Oops, it definitely is not. That was me trying to debug the condition. I was having an issue where the condition was `false`, but the job was still running. Printing it helped me learn the condition was indeed false, and the formatting in the condition was a problem. Removed it.

.github/test_events/README.md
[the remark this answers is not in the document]
BatmanAoD: The only special folder seems to be `workflows`, based on this: https://stackoverflow.com/a/61301254/1858225

.github/workflows/release-python.yml
MarquessV: Yeah it's weird, but [`always()`](https://docs.github.com/en/actions/learn-github-actions/expressions#always) is special, it doesn't just mean `true` here. It means always run the job. On its own, `needs` requires all specified jobs to finish successfully. That would be a problem, because we're making some jobs optional and "skipped" doesn't count as a success. We fix this by combining `needs` with `if: always()`, which makes it so the jobs specified in needs just need to be finished before the job can run, they don't need to be successful. We can add `needs.sdist.result == 'success'` to ensure there is always at least an sdist available to publish, but it would effectively do nothing on it's own.

BatmanAoD (approved):

PR #371, opened 2023-10-06
Add libquil support

notmgsk:
This MR adds new libquil implementations of the `quilc::Client` and `qvm::Client` traits.

Things to review:
* All libquil-related code is behind the `libquil` feature. This is specifically an opt-in feature and thus should have no effect on standard installations.

To do
- [ ] Testing
- [ ] Add `libquil` feature to CI testing


crates/lib/Cargo.toml
```diff
@@ -14,6 +14,7 @@ manual-tests = []
 tracing = ["dep:tracing", "qcs-api-client-common/tracing", "qcs-api-client-grpc/tracing", "qcs-api-client-openapi/tracing"]
 tracing-config = ["tracing", "qcs-api-client-common/tracing-config", "qcs-api-client-grpc/tracing-config", "qcs-api-client-openapi/tracing-config"]
 otel-tracing = ["tracing-config", "qcs-api-client-grpc/otel-tracing", "qcs-api-client-openapi/otel-tracing"]
+libquil = ["libquil-sys"]
```
BatmanAoD, line 17: Should we make this a default feature?

crates/lib/src/compiler/libquil.rs
```diff
@@ -0,0 +1,325 @@
...
+    fn quilc_client() -> Client {
+        Client {}
+    }
+
+    #[tokio::test]
+    async fn compare_native_quil_to_expected_output() {
+        let output = quilc_client()
+            .compile_program(
+                "MEASURE 0",
+                TargetDevice::try_from(qvm_isa()).expect("Couldn't build target device from ISA"),
```
Shadow53, line 175: `expect` messages should indicate [what is expected](https://doc.rust-lang.org/stable/std/result/enum.Result.html#recommended-message-style) to happen (also below)

crates/python/src/executable.rs
```diff
@@ -112,6 +111,8 @@ impl PyExecutable {
     ) -> Self {
         let quilc_client = quilc_client.map(|c| match c.inner {
             QuilcClient::Rpcq(c) => c,
+            #[cfg(feature = "libquil")]
+            QuilcClient::LibquilSys(_) => todo!(),
         });
```
Shadow53, line 116: `todo`

Shadow53 (changes requested):

crates/lib/Cargo.toml
notmgsk: Oh, interesting. I didn't realise there were implicit features. Thanks.

crates/lib/src/compiler/libquil.rs
notmgsk: Agreed. This was copy/pasted from a different file with the same issue.

notmgsk: I'll fix this when I improve the testing for libquil.

crates/python/src/qvm/mod.rs
```diff
@@ -52,15 +57,99 @@ impl PyQvmClient {
...
 
+    #[cfg(feature = "libquil")]
+    #[staticmethod]
+    fn new_libquil() -> PyResult<Self> {
+        Ok(Self {
+            inner: QvmClient::Libquil(qvm::libquil::Client {}),
+        })
+    }
+
+    #[cfg(not(feature = "libquil"))]
```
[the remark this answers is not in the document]
notmgsk: The Python stubs will have to contain this function regardless of whether the feature is enabled, so I think it makes sense to always have this function available and error if the feature isn't enabled.

crates/python/src/compiler/quilc.rs
```diff
@@ -106,9 +107,82 @@ impl PyTargetDevice {
...
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("error when compiling with RPCQ client: {0}")]
+    Rpcq(#[from] qcs::compiler::quilc::Error),
+    #[cfg(feature = "libquil")]
+    #[error("error when compiling with libquil client: {0}")]
+    Libquil(#[from] qcs::compiler::libquil::Error),
+}
+
+impl qcs::compiler::quilc::Client for QuilcClient {
```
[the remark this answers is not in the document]
notmgsk: I did try this but came up against a problem: the `Client` trait as an associated type `Client::Error`, which is different for both branches in that match. I originally had something like this
```rust
impl QuilcClient {
    pub(crate) fn as_client<E>(&self) -> &dyn qcs::compiler::quilc::Client<Error = E> {
        match self {
            QuilcClient::Rpcq(client) => client,
            QuilcClient::LibquilSys(client) => client,
        }
    }
}
```
but then I wasn't sure how to provide `E`.

crates/lib/src/executable.rs
```diff
@@ -329,7 +329,7 @@ impl Executable<'_, '_> {
     /// To disable compilation, set this to `None`.
     #[must_use]
     #[allow(trivial_casts)]
-    pub fn with_quilc_client<C: quilc::Client + Send + Sync + 'static>(
+    pub fn with_quilc_client<C: quilc::Client<Error = quilc::Error> + Send + Sync + 'static>(
```
jselig-rigetti, line 332: Is it possible/preferrable to make this constraint a `quilc::Client<Error = Into<quilc::Error>>`? I get an error trying to use the libquil client, see https://github.com/rigetti/qcs-sdk-rust/pull/387

PR #373, opened 2023-10-12
feat(python): provide access to the translation backend option

### LINKED ISSUE #372: Provide access to the configured `TranslationBackend` variant
See https://github.com/rigetti/pyquil/issues/1678

notmgsk:
Closes #372

### DIFF
--- a/crates/lib/src/qpu/translation.rs
+++ b/crates/lib/src/qpu/translation.rs
@@ -122,6 +122,12 @@ pub struct TranslationOptions {
 }
 
 impl TranslationOptions {
+    /// Get the backend used for translation
+    #[must_use]
+    pub fn get_backend(&self) -> Option<TranslationBackend> {
+        self.inner.translation_backend.clone()
+    }
+
     /// Use the first-generation translation backend available on QCS since 2018.
     pub fn use_backend_v1(&mut self) {
         self.inner.translation_backend =
--- a/crates/lib/src/qpu/translation.rs
+++ b/crates/lib/src/qpu/translation.rs
@@ -124,8 +124,8 @@ pub struct TranslationOptions {
 impl TranslationOptions {
     /// Get the backend used for translation
     #[must_use]
-    pub fn get_backend(&self) -> Option<TranslationBackend> {
-        self.inner.translation_backend.clone()
+    pub fn backend(&self) -> Option<&TranslationBackend> {
+        self.inner.translation_backend.as_ref()
     }
 
     /// Use the first-generation translation backend available on QCS since 2018.
--- a/crates/python/src/qpu/translation.rs
+++ b/crates/python/src/qpu/translation.rs
@@ -21,7 +21,7 @@ create_init_submodule! {
         PyQuiltCalibrations,
         PyTranslationOptions,
         PyTranslationResult,
-        PyBackend
+        PyTranslationBackend
     ],
     errors: [
         GetQuiltCalibrationsError,
@@ -91,13 +91,13 @@ py_wrap_error!(
 );
 
 #[derive(Copy, Clone)]
-pub enum Backend {
+pub enum TranslationBackend {
     V1,
     V2,
 }
 
 py_wrap_simple_enum! {
-    PyBackend(Backend) as "Backend" {
+    PyTranslationBackend(TranslationBackend) as "TranslationBackend" {
         V1,
         V2
     }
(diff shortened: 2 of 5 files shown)


crates/python/src/qpu/translation.rs
```diff
@@ -88,33 +90,59 @@ py_wrap_error!(
...
+        })
+    }
+
+    #[getter]
+    fn backend(&self) -> Option<PyTranslationBackend> {
+        self.backend.map(Into::into)
     }
 
     fn use_backend_v1(&mut self) {
-        self.0.use_backend_v1()
```
BatmanAoD, line 109: Isn't this method already available in Python?

crates/python/src/qpu/translation.rs
```diff
@@ -88,33 +90,59 @@ py_wrap_error!(
...
+    }
+}
+
 #[derive(Clone, Default)]
 #[pyclass(name = "TranslationOptions")]
-pub struct PyTranslationOptions(TranslationOptions);
+pub struct PyTranslationOptions {
+    inner: TranslationOptions,
+    backend: Option<TranslationBackend>,
+}
```
BatmanAoD, line 111: Since `TranslationOptions` already includes the backend (...in fact, I think that's currently the _only_ option), why do we need to duplicate that?

Shadow53: I'm also wondering the same thing

crates/python/src/qpu/translation.rs
notmgsk: Not sure what you mean?

BatmanAoD (approved):

BatmanAoD:
Any idea why the `Cargo.lock` diff is so big?

notmgsk:
> Any idea why the `Cargo.lock` diff is so big?

No. I had to do a `cargo update` to update a library that was causing CI to fail (I think it was `gix`). I'm guessing it hadn't been updated in a while.

Shadow53:
> > Any idea why the `Cargo.lock` diff is so big?
> 
> No. I had to do a `cargo update` to update a library that was causing CI to fail (I think it was `gix`). I'm guessing it hadn't been updated in a while.

For the future, you should be able to do, e.g., `cargo update -p gix` to only update one package.

crates/lib/src/qpu/translation.rs
```diff
@@ -122,6 +122,12 @@ pub struct TranslationOptions {
 }
 
 impl TranslationOptions {
+    /// Get the backend used for translation
+    #[must_use]
+    pub fn get_backend(&self) -> Option<TranslationBackend> {
+        self.inner.translation_backend.clone()
+    }
```
Shadow53, line 129: This name/signature is [more conventional](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).

```suggestion
    #[must_use]
    pub fn backend(&self) -> Option<&TranslationBackend> {
        self.inner.translation_backend.as_ref()
    }
```

Shadow53 (commented):
One nit to make a new function better match the Rust API Guidelines, otherwise looks good

crates/lib/src/qpu/translation.rs
BatmanAoD: Hm, I glossed over the `.clone()` here. `TranslationBackend` is just an enum with unit members, i.e. a tiny numeric type. So it should be `Copy`.

Given this, I think it should probably return a value.

BatmanAoD: ....actually, `TranslationBackend` _is_ `Copy`, and `Option<T>` implements `Copy` IFF `T` does... so why do we need `clone()` here?

notmgsk: `TranslationBackend` is _not_ copy.

notmgsk: > This name/signature is [more conventional](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).

Kalan has often asked that getters be prefixed with `get_`, but happy to follow the Rust style guide.

BatmanAoD: Is it a different `TranslationBackend` than this one? https://github.com/rigetti/qcs-sdk-rust/pull/373/files#diff-e83cc7a96cad7becb1c3df31f77635e277497be983a6dd80c3546ca22fade877R94

BatmanAoD: ...oh, I see, it's the GRPC type.

PR #380, opened 2023-10-19
feat: make Executable.get_qcs_client public

jselig-rigetti:
Up for debate, but in cases where you need to provide some execution configuration, it would help to use a single source of truth for the qcs client

For example, to use https://github.com/rigetti/qcs-sdk-rust/blob/a4c832ca7360d546a0bd0c71230353f6a4b28eaa/crates/lib/src/executable.rs#L367 

Here's what you have to do:

```rust
// ... in some other function ...

let exe = qcs::executable::Executable::from_quil("...");

// ... later on ...

// > before this change: who knows if `exe` is configured with
// the same client or if we're doing a redundant load
let qcs_client = qcs::client::Qcs::load().await;

// > after this change: same client, only one load
let qcs_client = exe.get_qcs_client().await;

let qvm_client = qcs::qvm::http::HttpClient::from(&qcs_client);
exe.execute_on_qvm(&qvm_client)
```

### DIFF
--- a/crates/lib/src/executable.rs
+++ b/crates/lib/src/executable.rs
@@ -299,8 +299,9 @@ impl<'executable> Executable<'executable, '_> {
         self
     }
 
-    /// Get a reference to the [`Qcs`] client used by the executable,
-    /// which loads and stores a default client if one has not been set.
+    /// Get a reference to the [`Qcs`] client used by the executable.
+    ///
+    /// If one has not been set, a default client is loaded, set, and returned.
     pub async fn get_qcs_client(&mut self) -> Arc<Qcs> {
         if let Some(client) = &self.qcs_client {
             client.clone()
--- a/crates/lib/src/executable.rs
+++ b/crates/lib/src/executable.rs
@@ -302,7 +302,7 @@ impl<'executable> Executable<'executable, '_> {
     /// Get a reference to the [`Qcs`] client used by the executable.
     ///
     /// If one has not been set, a default client is loaded, set, and returned.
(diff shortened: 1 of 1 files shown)


crates/lib/src/executable.rs
```diff
@@ -299,14 +299,15 @@ impl<'executable> Executable<'executable, '_> {
         self
     }
 
-    /// Load `self.client` if not yet loaded, then return a reference to it.
-    async fn get_qcs_client(&mut self) -> Result<Arc<Qcs>, Error> {
+    /// Get a reference to the [`Qcs`] client used by the executable,
+    /// which loads and stores a default client if one has not been set.
+    pub async fn get_qcs_client(&mut self) -> Arc<Qcs> {
```
Shadow53, line 304: Since we are making this into a public method, we may want to consider more closely following the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) for naming such things. In this case, I think the advice would be to just call this `qcs_client()`.

Shadow53 (commented):
I don't see any issue with making this public. Just a couple of comments on tidying things up before merge.

BatmanAoD (approved):
Seems reasonable to me.

The most obvious alternative, to me, is to have something like a thread-static or other "global" value. I'm not clear on why you'd need multiple QCS clients, or at least not multiple per profile. But the fact that you might indeed need one for each profile means you'd need to have a global map or something, which is...icky. So I think this is better.

PR #390, opened 2023-11-09
Add libquil support

notmgsk:
This MR adds new libquil implementations of the quilc::Client and qvm::Client traits.

Things to review:
* All libquil-related code is behind the libquil feature. This is specifically an opt-in feature and thus should have no effect on standard installations.

To do
- [x] Testing
- [x] Add libquil feature to CI testing


crates/python/qcs_sdk/qvm/__init__.pyi
```diff
@@ -13,6 +13,10 @@ class QVMClient:
     def new_http(endpoint: str) -> QVMClient:
         """Construct a new client which uses HTTP to communicate with QVM"""
         ...
+    @staticmethod
+    def new_libquil() -> QVMClient:
+        """Construct a new client which uses libquil to communicate with QVM"""
```
BatmanAoD, line 18: To "communicate with" QVM, or as the QVM implementation itself?

BatmanAoD (commented):
Partial review; mostly stylistic comments

crates/lib/Makefile.toml
```diff
@@ -21,6 +21,10 @@ run_task = [{name = "test-flow"}]
 [tasks.lint]
 dependencies = ["clippy-flow"]
 
+[tasks.examples]
+command = "cargo"
+args = ["build", "--examples", "--features", "libquil"]
```
Shadow53, line 26: Is `--features libquil` required given the change on line 3?

crates/lib/src/compiler/libquil.rs
```diff
@@ -0,0 +1,349 @@
...
+impl quilc::Client for Client {
+    fn compile_program(
+        &self,
+        quil: &str,
+        isa: quilc::TargetDevice,
+        options: quilc::CompilerOpts,
+    ) -> Result<quilc::CompilationResult, quilc::Error> {
+        let program = libquil_sys::quilc::Program::from_str(quil).map_err(Error::from)?;
+        let isa = serde_json::to_string(&isa).map_err(Error::from)?;
+        let chip = libquil_sys::quilc::Chip::from_str(&isa).map_err(Error::from)?;
```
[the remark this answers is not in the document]
Shadow53: I might be wrong, but this looks necessary, since it's going from a libquil error -> `Error` -> `super::quilc::Error`

crates/lib/src/compiler/libquil.rs
```diff
@@ -0,0 +1,349 @@
...
+
+        let compilation_result = if options.protoquil.unwrap_or_default() {
+            libquil_sys::quilc::compile_protoquil(&program, &chip)
+        } else {
+            libquil_sys::quilc::compile_program(&program, &chip)
+        }
+        .map_err(Error::from)?;
```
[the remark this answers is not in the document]
Shadow53: See my note on previous comment

crates/lib/src/compiler/libquil.rs
```diff
@@ -0,0 +1,349 @@
...
+        Ok(quilc::CompilationResult {
+            program,
+            native_quil_metadata: compilation_result.metadata.map(Into::into),
+        })
+    }
+
+    fn get_version_info(&self) -> Result<String, quilc::Error> {
+        Ok(libquil_sys::quilc::get_version_info()
+            .map_err(Error::from)?
+            .version)
```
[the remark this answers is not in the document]
Shadow53: I think that would be stylistically preferable, though we'd lose the implicit error conversion. `get_version_info().map(|info| info.version).map_err(|error| Error::from(error).into)` is a little more readable, IMO.

crates/lib/src/compiler/libquil.rs
```diff
@@ -0,0 +1,349 @@
...
+    fn conjugate_pauli_by_clifford(
+        &self,
+        request: quilc::ConjugateByCliffordRequest,
+    ) -> Result<quilc::ConjugatePauliByCliffordResponse, quilc::Error> {
+        let pauli_terms = request
+            .pauli
+            .symbols
+            .into_iter()
+            .map(CString::new)
+            .collect::<Result<Vec<_>, _>>()
```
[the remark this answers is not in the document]
Shadow53: I'd only recommend using `itertools` if it is already present as a dependency -- it's not worth adding just for this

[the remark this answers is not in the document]
Shadow53: Also the above example @BatmanAoD gave actually won't work because of the `?` -- setting the type outside of the turbofish is stating what is expected _after_ the `?`. This should work, though:

```rust
let pauli_terms: Vec<_> = request
    ...
    .collect::<Result<_, _>>()?;
```

crates/lib/src/compiler/libquil.rs
```diff
@@ -0,0 +1,349 @@
...
+            .map_err(Error::from)?;
+        let interleaver = request
+            .interleaver
+            .map(|s| s.parse::<libquil_sys::quilc::Program>())
+            .transpose()
+            .map_err(Error::from)?;
+        let result = libquil_sys::quilc::generate_rb_sequence(
+            i32::try_from(request.depth).map_err(Error::from)?,
+            i32::try_from(request.qubits).map_err(Error::from)?,
```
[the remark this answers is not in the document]
Shadow53: Per my notes above, I think you'll still need the error mapping:

```suggestion
            request.depth.try_into().map_err(Error::from)?,
            request.qubits.try_into().map_err(Error::from)?,
```

crates/lib/src/compiler/libquil.rs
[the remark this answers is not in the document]
Shadow53: This could be made unnecessary by implementing `From<libquil::Error>` for `super::quilc::Error`, which would remove the need for this intermediate transformation by moving it into the `From` impl.

crates/lib/src/executable.rs
```diff
@@ -841,6 +841,7 @@ mod describe_qpu_for_id {
             panic!("Expected an error!");
         };
         let result_string = format!("{err:?}");
+        println!("{result_string}");
```
Shadow53, line 844: Debug print

crates/lib/src/qvm/libquil.rs
```diff
@@ -0,0 +1,184 @@
...
+            .compiled_quil
+            .parse()
+            .map_err(Error::LibquilSysQuilc)?;
+        let addresses = request
+            .addresses
+            .iter()
+            .map(|(address, indices)| match indices {
+                AddressRequest::Indices(indices) => Ok((
+                    address.clone(),
+                    libquil_sys::qvm::MultishotAddressRequest::Indices(indices.clone()),
```
Shadow53, line 65: This applies throughout the PR, but why not import these types?

crates/lib/src/qvm/libquil.rs
```diff
@@ -0,0 +1,184 @@
...
+                    libquil_sys::qvm::MultishotAddressRequest::All,
+                )),
+                AddressRequest::ExcludeAll => Err(Error::UnsupportedIndicesRequestType),
+            })
+            .collect::<Result<_, _>>()?;
+        let result =
+            libquil_sys::qvm::multishot(&program, addresses, i32::from(request.trials.get()))
+                .map_err(Error::LibquilSysQvm)?;
+        let mut registers = HashMap::new();
+        for (address, values) in result {
```
Shadow53, line 78: On mobile, so multiline suggestions are difficult, but you should be able to replace this loop with an iterator that collects into a `HashMap`. It may even be slightly more optimized since iterators provide size hints that collections can use to reserve the right capacity all at once.

This is mostly a style suggestion though, and is technically fine as-is. For the optimization part, you may want to use `HashMap::with_capacity(result.len())` above

crates/lib/src/qvm/libquil.rs
```diff
@@ -0,0 +1,184 @@
...
+    async fn run_and_measure(
+        &self,
+        request: &http::MultishotMeasureRequest,
+        _options: &QvmOptions,
+    ) -> Result<Vec<Vec<i64>>, super::Error> {
+        let program = request
+            .compiled_quil
+            .parse()
+            .map_err(Error::LibquilSysQuilc)?;
+        let qubits = request.qubits.clone();
```
Shadow53, line 128: Should be able to do

```rust
request.qubits.iter()
    .copied()
    .map(...)
    ...
```

Right? Or are qubits not `Copy`?

crates/lib/tests/qvm_api.rs
```diff
@@ -6,28 +6,34 @@ use std::{collections::HashMap, num::NonZeroU16};
...
+fn libquil_qvm_client() -> qvm::libquil::Client {
+    qvm::libquil::Client {}
+}
+
+#[test_case::test_case(http_qvm_client().await ; "with HTTP client")]
+#[test_case::test_case(libquil_qvm_client() ; "with libquil client")]
 #[tokio::test]
-async fn test_get_version_info() {
-    let client = qvm_client().await;
+async fn test_get_version_info<C: qvm::Client>(client: C) {
```
Shadow53, line 36: Can also do this, it amounts to the same thing:

```suggestion
async fn test_get_version_info(client: impl qvm::Client) {
```

crates/python/src/executable.rs
```diff
@@ -112,6 +111,8 @@ impl PyExecutable {
     ) -> Self {
         let quilc_client = quilc_client.map(|c| match c.inner {
             QuilcClient::Rpcq(c) => c,
+            #[cfg(feature = "libquil")]
+            QuilcClient::LibquilSys(_) => todo!(),
```
Shadow53, line 115: TODO

Cargo.toml
```diff
@@ -24,7 +24,7 @@ quil-rs = { version = "0.22.0", git = "https://github.com/rigetti/quil-rs", tag
 # and need to be updated together.
 ndarray = { version = "0.15.6", features = ["serde"] }
 numpy = "0.19"
-pyo3 = "0.19.2"
+pyo3 = { version = "0.19.2", features = ["extension-module"] }
```
MarquessV, line 27: Is the `extension-module` feature needed here? IIRC we omit it here and instead include it in `build.rs` because having the feature here causes problems when building the Python library as a Rust target.

crates/lib/src/compiler/libquil.rs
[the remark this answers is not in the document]
BatmanAoD: I think you can still have the `?` in there?

.github/workflows/checks.yml
```diff
@@ -30,7 +36,6 @@ jobs:
         env:
           CARGO_MAKE_RUN_CHECK_FORMAT: true
           CARGO_MAKE_RUN_CLIPPY: true
-          CARGO_MAKE_BUILD_EXAMPLES: true
```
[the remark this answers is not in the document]
notmgsk: I removed this because it was trying to build examples without the `libquil` feature and thus failing on the libquil example. I added a `cargo build --examples --features libquil` elsewhere so they are still checked.

crates/lib/src/compiler/libquil.rs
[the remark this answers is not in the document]
notmgsk: It would actually have to be `From<libquil_sys::quilc::Error for super::quilc::Error`, and I don't think it makes sense to have `libquil_sys`-specific error handling in the parent module (`super::quilc`).

crates/lib/src/compiler/libquil.rs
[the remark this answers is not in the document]
notmgsk: I don't really mind the turbofish syntax. Kinda like it, and not only because it's called "turbofish" 🤷🏽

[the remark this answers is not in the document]
notmgsk: That said, the `Vec<_>` _is_ unnecessary in the referenced code.

crates/lib/src/compiler/libquil.rs
```diff
@@ -0,0 +1,349 @@
...
+            gateset.iter().collect(),
+            request
+                .seed
+                .map(i32::try_from)
+                .transpose()
+                .map_err(Error::from)?,
+            interleaver.as_ref(),
```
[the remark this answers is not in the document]
notmgsk: Temporary values, man... temporary values. It is a little bit cleaner this way however
```rust
        let gateset = request
            .gateset
            .iter()
            .map(String::as_str)
            .map(str::parse)
            .collect::<Result<Vec<_>, _>>()
            .map_err(Error::from)?;
        let gateset = gateset.iter().collect();
```

crates/python/qcs_sdk/qvm/__init__.pyi
notmgsk: `libquil` is not the implementation, but the bridge _to_ the implementation (QVM)

crates/lib/Makefile.toml
notmgsk: I think so. Those are the features passed to `cargo test`.

.github/workflows/checks.yml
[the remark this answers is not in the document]
BatmanAoD: Can the libquil example be marked `[cfg(feature="libquil")]`?

Cargo.toml
notmgsk: Hm. No, not needed. I was having trouble getting tests to run, and the pyo3 docs suggested this change.

deny.toml
```diff
@@ -92,7 +92,7 @@ ignore = false
 # More documentation about the 'bans' section can be found here:
 # https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
 [bans]
-multiple-versions = "deny"
+multiple-versions = "warn"
```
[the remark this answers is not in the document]
notmgsk: Unfortunately https://github.com/rigetti/qcs-sdk-rust/actions/runs/6897076783

.github/workflows/checks.yml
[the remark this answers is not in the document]
Shadow53: I think the suggested thing is to mark it as requiring the `libquil` feature in `Cargo.toml` (see [here](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples) and [here](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#configuring-a-target)):

```toml
[[example]]
name = "libquil"
required-features = ["libquil"]
```

If I remember correctly how this works, it will just be skipped if the feature is not enabled.

Cargo.toml
```diff
@@ -24,7 +24,7 @@ quil-rs = { version = "0.22.0", git = "https://github.com/rigetti/quil-rs", tag
 # and need to be updated together.
 ndarray = { version = "0.15.6", features = ["serde"] }
 numpy = "0.19"
-pyo3 = "0.19.2"
+pyo3 = { version = "0.19.2" }
```
Shadow53, line 27: This change can be undone

.github/workflows/checks.yml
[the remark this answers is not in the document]
Shadow53: Just saw that you did this below. You should be able to re-enable this and not get failures.

crates/lib/Cargo.toml
```diff
@@ -64,3 +66,28 @@ insta = "1.29.0"
...
+path = "examples/parametric_compilation.rs"
+
+[[example]]
+name = "quil-t"
+path = "examples/quil_t.rs"
+
+[[example]]
+name = "compilation-and-simulation-with-libquil"
+path = "examples/libquil.rs"
+required-features = ["libquil"]
```
Shadow53, line 93: I don't think you should need to specify every example like this if you are just specifying the required features for one of them. I think you can get the same effect (with different names) with just

```suggestion

[[example]]
name = "libquil"
required-features = ["libquil"]
```

crates/lib/src/compiler/libquil.rs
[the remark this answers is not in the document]
Shadow53: Is this thread resolved? Looks like the `From<Error> for quilc::Error` above addresses this

crates/lib/src/compiler/quilc.rs
```diff
@@ -210,19 +210,22 @@ pub enum Error {
...
-impl Error {
-    pub(crate) fn from_quilc_error(quilc_uri: String, source: rpcq::Error) -> Self {
-        match source {
-            rpcq::Error::Response(message) => Error::QuilcCompilation(message),
-            source => Error::QuilcConnection(quilc_uri, source),
-        }
-    }
+/// Errors during compilation with one of the supported clients
+#[derive(Debug, thiserror::Error)]
+pub enum CompilationError {
```
Shadow53, line 221: Looking at this again, we may want to add `#[non_exhaustive]` ([link](https://doc.rust-lang.org/reference/attributes/type_system.html)) to all of these enums that have a conditional `Libquil` item.

crates/python/qcs_sdk/qvm/__init__.pyi
Shadow53: Doesn't the `libquil` library contain the QVM code, though? Reading this, I would understand the comment to mean that I still need a QVM service running, but it uses a different protocol to communicate with the service, whereas my understanding is that `libquil` _contains_ (i.e. links to) the QVM code itself.

Maybe this would be better?

 ```suggestion
        """Construct a new client which uses libquil to provide the QVM"""
```

File: crates/python/src/compiler/quilc.rs [hunk 21]
```diff
@@ -103,12 +100,60 @@ impl PyTargetDevice {
     }
 }
 
-wrap_error!(RustRpcqError(rpcq::Error));
-py_wrap_error!(quilc, RustRpcqError, RpcqError, PyRuntimeError);
-
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub enum QuilcClient {
```
Shadow53, line 104: For reference, this would be another enum that would want `#[non_exhaustive]`

Shadow53: OTOH, we don't intend for this to be used in _Rust_ code outside of this crate, so maybe not.

crates/lib/src/compiler/libquil.rs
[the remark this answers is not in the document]
notmgsk: yes

crates/python/src/compiler/quilc.rs
notmgsk: `#[non_exhaustive]` is new to me -- thanks.

notmgsk: But yeah, probably not necessary here.

.github/workflows/checks.yml
```diff
@@ -17,6 +17,12 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           version: '3.20.1'
 
+      - name: Install libquil dependencies
+        run: sudo apt install -y libffi\* libblas\* liblapack\*
```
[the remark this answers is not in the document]
notmgsk: The `libquil` README already has [installation instructions](https://github.com/rigetti/libquil#automated-installation).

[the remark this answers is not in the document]
notmgsk: Added this to the qcs-sdk-rust readme
```
### `libquil`

[`libquil`](https://github.com/rigetti/libquil) provides [quilc](https://github.com/quil-lang/quilc) and [QVM](https://github.com/quil-lang/qvm) as a shared library, which can be used by `qcs-sdk-rust` as an alternative client for those tools.

To use `libquil`:
* install the library (see [installation instructions](https://github.com/rigetti/libquil#automated-installation))
* enable the feature with `--features libquil`
```

crates/python/src/qvm/mod.rs
```diff
@@ -52,15 +67,79 @@ impl PyQvmClient {
...
+            #[cfg(feature = "libquil")]
+            QvmClient::Libquil(_) => Ok("".into()),
+        }
     }
 }
 
-wrap_error!(RustQvmError(qcs::qvm::Error));
-py_wrap_error!(api, RustQvmError, QVMError, PyRuntimeError);
+#[async_trait::async_trait]
+impl qvm::Client for QvmClient {
```
BatmanAoD, line 97: It looks like you're now using `.as_client()` wherever there used to be a direct conversion to `qvm::http::HttpClient`, so why is this new implementation necessary? Or, conversely, why not get rid of some of the `as_client()` calls in other modules, and rely on this implementation?

BatmanAoD (approved):
One question about the new `impl qvm::Client for QvmClient`, but this looks good!

crates/python/src/qvm/mod.rs
Shadow53: FWIW, that had been my suggestion in the beginning: use `as_client()` in this implementation and then rely on this implementation everywhere else.

crates/lib/src/executable.rs
```diff
@@ -899,7 +899,7 @@ mod describe_qpu_for_id {
         let mut exe = exe.with_qcs_client(Qcs::default());
         let result = exe.qpu_for_id("Aspen-8").await;
 
-        assert!(matches!(result, Err(_)));
-        assert!(matches!(exe.qpu, None));
+        assert!(result.is_err());
```
Shadow53, line 902: I prefer to have assertions that the error is the one we expect, not just any error

Shadow53 (approved):
One note about making a test more robust, otherwise looks good!

PR #391, opened 2023-11-14
feat: instrument pyo3-opentelemetry

erichulburd:
### Goals

The goal of this MR is to support collection of Rust traces by Python end users. Specifically this entails:

1. Connecting traces from Python to Rust.
2. Ensuring all network API calls are instrumented with tracing (QCS API client already has support for this with the tracing-opentelemetry feature).
3. Supporting the ability of Python users to configure and initialize Rust tracing subscribers.

### Design & Implementation

* The `pyo3_opentelemetry::pypropagate` macro will get the OpenTelemetry context from Python, deserialize it, and set it as the current context on the Rust side. This ensures traces initialized on the Python side are connected to traces within Rust.
* You will see many calls to [FutureExt::with_current_context](https://docs.rs/opentelemetry/latest/opentelemetry/trace/trait.FutureExt.html#method.with_current_context) on the Rust side. This ensures that any Rust futures propagate their traces across async / await. This is especially relevant for network calls.
* In order to configure and initialize a tracing subscriber _from Python_, we add the `tracing_subscriber` module from `pyo3-tracing-subscriber` (see crates/python/src/lib.rs::qcs_sdk). We also autogenerate stubs for this Python module in crates/python/build.rs (stub files are formatted with `cargo make format-tracing-subscriber`).


crates/python/Cargo.toml
```diff
@@ -33,6 +33,11 @@ rigetti-pyo3.workspace = true
 paste = "1.0.11"
 pyo3-log = "0.8.2"
 once_cell = "1.18.0"
+pyo3-opentelemetry = { version = "0.2.0-rc.0" } 
+opentelemetry = "0.20.0"
+tracing = "0.1.37"
+pyo3-tracing-subscriber = { version = "0.2.0-rc.0", features = ["layer-otel-otlp-file", "layer-otel-otlp"] }
```
MarquessV, line 39: I'm noticing we missed this with `pyo3-log`, but can you specify these in the root `Cargo.toml` with the other `pyo3` dependencies? They often need to be updated together, so grouping them there makes that process a bit easier.

crates/python/qcs_sdk/tracing_subscriber/__init__.pyi
```diff
@@ -0,0 +1,119 @@
...
+    ...
+
+class TracingShutdownError(RuntimeError):
+    """
+    Raised if the tracing subscriber fails to shutdown cleanly.
+    """
+
+    ...
+
+class BatchConfig:
```
MarquessV, line 41: These type hints do not pass [stubtest](https://mypy.readthedocs.io/en/stable/stubtest.html). There are 65 in total, but for an example, this one fails with:

```
error: qcs_sdk.BatchConfig is not present in stub
Stub: in file .../qcs-sdk-rust/crates/python/qcs_sdk/__init__.pyi
```

When you are in `crates/python`: `cargo make stubtest-flow` will install the module and run stubtest and `cargo make stubtest` will just run `stubtest`. The latter is helpful when you are only iterating in the `pyi` files.

crates/python/src/compiler/quilc.rs
```diff
@@ -136,8 +136,10 @@ impl PyQuilcClient {
 }
 
 py_function_sync_async! {
+    #[pyo3_opentelemetry::pypropagate(on_context_extraction_failure = "print")]
```
MarquessV, line 139: Why do we print the failures on these instead of the default?

crates/python/src/executable.rs
```diff
@@ -250,6 +266,7 @@ impl PyExecutable {
                     translation_options,
                     execution_options.unwrap_or_default().as_inner(),
                 )
+                .with_current_context()
```
MarquessV, line 269: I'm a little bit worried about these being easy to forget. It seems like there isn't any kind of warning or error raised if it's not there. I guess it just requires us to have a diligent eye when testing our traces.

crates/python/src/lib.rs
```diff
@@ -58,6 +58,7 @@ fn qcs_sdk(py: Python<'_>, m: &PyModule) -> PyResult<()> {
         Err(e) => eprintln!("Failed to initialize the qcs_sdk logger: {e}"),
     }
     init_submodule("qcs_sdk", py, m)?;
+    pyo3_tracing_subscriber::add_submodule("qcs_sdk.tracing_subscriber", py, m)?;
```
MarquessV, line 61: I'm not noticing any problems with imports as written, but just in case something unexpected comes up with them, we may need to update `sys.modules` and the module's `__name__` manually. Similar to what [rigetti-pyo3 does](https://github.com/rigetti/rigetti-pyo3/pull/35/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R234).

crates/python/src/py_sync.rs
```diff
@@ -113,14 +113,16 @@ macro_rules! py_function_sync_async {
...
+            $crate::py_sync::py_sync!(py, $name($($arg),*).with_current_context())
         }
 
         $(#[$meta])+
         #[pyo3(name = $name "_async")]
         #[allow(clippy::too_many_arguments)]
         pub fn [< py_ $name _async >](py: ::pyo3::Python<'_> $(, $(#[$arg_meta])*$arg: $kind)*) -> ::pyo3::PyResult<&::pyo3::PyAny> {
-            $crate::py_sync::py_async!(py, $name($($arg),*))
+            use opentelemetry::trace::FutureExt;
+            $crate::py_sync::py_async!(py, $name($($arg),*).with_current_context())
```
MarquessV, line 125: I don't think this should cause issues, but these macros are sensitive to deadlocking and/or blocking signals, so that functionality is worth some extra care when testing these changes.

MarquessV (commented):
Thanks for doing this! A couple of comments, but mostly looks like issues with the type hints.

crates/python/pyproject.toml
```diff
@@ -44,6 +44,7 @@ sdist-include = ["README.md"]
 
 [tool.poetry.dependencies]
 python = "^3.8"
+opentelemetry-api = "^1.20.0"
```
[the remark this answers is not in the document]
MarquessV: After becoming more familiar with the [pyo3_opentelemetry docs](https://docs.rs/pyo3-opentelemetry/latest/pyo3_opentelemetry/), it looks like it is required for Python consumers.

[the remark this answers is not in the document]
erichulburd: Good point. Yes it would definitely be required, but only when using the tracing subscriber. I'm thinking we should treat that as an extra, so we don't burden anyone with unwanted dependencies.

I suppose that means including it as an [optional-dependency](https://peps.python.org/pep-0631/#optional-dependencies) and that I should define `tool.poetry.extras.tracing-opentelemetry`.

crates/python/src/compiler/quilc.rs
erichulburd: I actually changed this upstream to make "print" the default behavior, so once I update the version here, I can drop these.

crates/python/src/executable.rs
erichulburd: Yup, I think that's just the nature of the instrumentation library. To that end, I've implemented integration tests to cover the most important spans. If we refactor and drop one of these in a sensitive place, it will cause a test failure.

crates/python/src/lib.rs
erichulburd: Actually, `pyo3_tracing_subscriber::add_submodule` copies what's done in rigetti-pyo3 so we should be good there.

crates/python/src/py_sync.rs
erichulburd: Is there an example of this behavior I can reference?

`FutureExt::with_context`'s main job is to attach the current context: https://docs.rs/opentelemetry/latest/src/opentelemetry/trace/context.rs.html#445-450

Attaching does use a `RefCell`; https://docs.rs/opentelemetry/latest/src/opentelemetry/context.rs.html#11 and RefCell [panics rather than deadlocks](https://doc.rust-lang.org/std/cell/index.html#refcellt) on an attempted illegal borrow.

crates/python/qcs_sdk/tracing_subscriber/__init__.pyi
erichulburd: Ahh, did not know about this utility. I will fix it upstream and port it over here.

erichulburd: Finally got those passing after a good tussle. They actually did help me identify a bug though.

crates/python/src/py_sync.rs
MarquessV: Testing this is non-obvious, so I've made a note to document it better. In short, we just need to make sure functions wrapped by this code (1) don't deadlock, and (2) don't monopolize the GIL and check for signals.

The way I do this is by testing `qcs_sdk.compiler.quilc.compile_program` (and `compile_program_async`). If things compile as expected, then great! It's not deadlocking. I then repeat the same test, but without the `quilc` server running. The call should hang while waiting to connect to the `quilc` server and you should be able to cancel it with `Ctrl-C`. If you can, then it's not monopolizing the GIL and Python is able to do signal handling on the main thread. 

I went ahead and ran this test and these changes ran just fine! I'll resolve this.

crates/python/pyproject.toml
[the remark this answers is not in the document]
MarquessV: That looks right to me! Since this is the first extra, when the releases as an RC, it would be good to validate that a fresh project doesn't run into any missing dependency errors, that way we're confident the full release won't break anyone.

MarquessV:
Oh - when testing I also noticed that `poetry.lock` needs to be updated.

MarquessV (approved):

MarquessV:
@erichulburd The checks are passing on this PR now. Would you like me to rebase, merge, and release an RC?

PR #393, opened 2023-11-15
feat: Add memory_values to QPUResultData

### LINKED ISSUE #392: QPUResultData should include the `memory_values` returned from a job execution result.
A `JobExecutionResult` contains a `memory_values` field that contains the final state of each memory region at the end of execution. We should include this field in `QPUResultData`.

MarquessV:
closes #392


crates/lib/src/qpu/result_data.rs
```diff
@@ -21,12 +22,25 @@ pub enum ReadoutValues {
...
+}
+
 /// This struct encapsulates data returned from the QPU after executing a job.
 #[allow(clippy::module_name_repetitions)]
 #[derive(Debug, Clone, PartialEq)]
 pub struct QpuResultData {
     pub(crate) mappings: HashMap<String, String>,
     pub(crate) readout_values: HashMap<String, ReadoutValues>,
+    /// The final contents of each memory region, keyed on region name.
+    pub(crate) memory_values: HashMap<String, Option<MemoryValues>>,
```
kalzoo, line 43: Given that `HashMap::get` already returns an `Option`, I don't follow the need for the value to also be an option. Can't we just omit the key if the value is `None`?

Another note on this down below in the `try_fold`.

crates/python/src/qpu/api.rs
```diff
@@ -187,34 +191,83 @@ pub struct ExecutionResults {
...
+                        }
+                        Some(data_value::Value::Real(value)) => Some(PyMemoryValues::from_real(
+                            py,
+                            value
+                                .data
+                                .iter()
+                                .map(|v| v.to_object(py).extract(py))
+                                .collect::<PyResult<Vec<_>>>()?,
+                        )?),
+                        None => None,
```
kalzoo, line 259: Why not just skip this case? It's mostly an artifact of protobuf making everything optional. If the values are missing, is there benefit to not just treating it as if that key wasn't in the map at all?

crates/lib/src/execution_data.rs
```diff
@@ -352,19 +352,31 @@ mod describe_register_map {
...
     }
 
+    fn dummy_memory_values(v: Vec<i64>) -> ControllerMemoryValue {
+        ControllerMemoryValue {
+            value: Some(controller::data_value::Value::Integer(IntegerDataValue {
+                data: v,
+            })),
+        }
+    }
+
```
kalzoo, line 379: Should there be test cases for the other variants (Binary + Real)?

kalzoo (changes requested):

crates/lib/src/qpu/result_data.rs
MarquessV: Yeah, forgot to get ahead of this one with a comment, so thank you for calling it out. I wasn't sure if there was something meaningful about a key appearing with a `None` value versus the key not appearing at all. Since it sounds like there isn't, I'll remove the `Option`.

crates/lib/src/execution_data.rs
MarquessV: Updated the test to include all variants.

crates/python/src/qpu/api.rs
```diff
@@ -187,34 +191,83 @@ pub struct ExecutionResults {
     /// QPU execution duration.
     #[pyo3(get)]
     pub execution_duration_microseconds: Option<u64>,
+    #[pyo3(get)]
+    pub memory: HashMap<String, Option<PyMemoryValues>>,
```
Shadow53, line 195: Why is this `Option<PyMemoryValues>`? Shouldn't a `None` correspond to an entry not existing in the map?

crates/python/src/qpu/api.rs
Shadow53: I had the same thought just now, and this thread was marked resolved without any clear indication why.

My suggestion would be a `.filter_map()` returning `Option<PyResult<PyMemoryValues>>` followed by a `.collect()`, instead of the `try_fold`

Shadow53 (commented):
Looks good, just confused about using `HashMap<String, Option<_>>`.

crates/python/src/qpu/api.rs
MarquessV: Oops, should've been cleaned up with the others, fixed.

crates/python/src/qpu/api.rs
MarquessV: `try_fold` isn't used because of `Option`, but because the conversion to Python types from rust types is fallible and it lets us handle those errors in the iterator.

kalzoo (approved):
looking good!

crates/python/src/qpu/api.rs
kalzoo: Ah, I hadn't seen this comment since the diff was outdated, but note that a `map` iterator can be fallible if you `collect::<Result<_,_>>()?`

MarquessV: Yes but you lose the syntactical sugar of `?` and `try_fold` will end early if it hits an error.

PR #407, opened 2023-12-05
feat!: expose BackendV2 options through Python

### LINKED ISSUE #406: Expose TranslationOptions through Python

Shadow53:
Resolves #406.

This refactors the SDK `TranslationOptions` to add methods to set gRPC options without needing to deal with the `prost` types.

### DIFF
--- a/crates/lib/src/qpu/translation.rs
+++ b/crates/lib/src/qpu/translation.rs
@@ -130,33 +130,25 @@ pub enum TranslationBackendMismatch {
 #[allow(clippy::module_name_repetitions)]
 #[derive(Clone, Debug, Default)]
 pub struct TranslationOptions {
-    inner: Option<ApiTranslationOptions>,
+    inner: ApiTranslationOptions,
 }
 
 impl TranslationOptions {
     /// Get the backend used for translation
     #[must_use]
     pub fn backend(&self) -> Option<&TranslationBackend> {
-        self.inner
-            .as_ref()
-            .and_then(|options| options.translation_backend.as_ref())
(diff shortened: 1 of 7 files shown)


crates/lib/src/qpu/translation.rs
```diff
@@ -128,16 +140,109 @@ impl TranslationOptions {
...
+            Err(TranslationBackendMismatch::V2)
+        }
+    }
+
+    /// If `false`, default calibrations will not be prepended to the translated program.
+    ///
+    /// # Errors
+    ///
+    /// This will return an error if the translation backend is set to something other than V2.
+    pub fn v2_prepend_default_calibrations(
```
MarquessV, line 198: I wonder if we should just drop the `v2` in the name here since `v2` is eventually going to be the only option.

Shadow53: Fair point. My thought was having the v2 in the name makes it more clear which backend must be set to avoid an error when calling the method, since it is only valid for v2.

We could also discuss whether the error makes sense -- I think it does, since the alternatives of silently replacing the backend version or doing nothing on mismatch seems like bad UX.

MarquessV: I think returning an error is a good choice for sure since it avoids any surprises. I think it makes dropping `v2` a valid option, since in the worst case, users will miss the docstring and get an error guiding them to use backend v2. In any case, not something that will keep me from letting this ship, so I'll leave it to you to decide.

MarquessV (approved):
Assuming CI goes 🟢, 🚀

PR #412, opened 2023-12-12
doc(python): Publish documentation on GitHub pages

MarquessV:
Use [pdoc](https://pdoc.dev/) and GitHub pages to build and deploy a [live site](https://rigetti.github.io/qcs-sdk-rust/qcs_sdk.html) with our `qcs-sdk-python` type hints as documentation. As part of this, I updated a some bits of the documentation with extra info or changes to make the formatting look better in the rendered web page.


crates/python/README.md
```diff
@@ -8,6 +8,10 @@ to compile and run Quil programs on Rigetti quantum processors. Internally, it i
 While this package can be used directly, [pyQuil](https://pypi.org/project/pyquil/) offers more functionality and a 
 higher-level interface for building and executing Quil programs. This package is still in early development and breaking changes should be expected between minor versions.
 
+# Documentation
+
+Documentation for the current release of `qcs_sdk` is published [here](https://rigetti.github.io/qcs-sdk-rust/qcs_sdk.html). Every version of `quil` ships [with type stubs](https://github.com/rigetti/qcs-sdk-rust/tree/main/crates/python/qcs_sdk) that provide type hints and documentation to Python tooling and editors that support the [Language Server Protocol](https://microsoft.github.io/language-server-protocol/) or similar.
```
kalzoo, line 13: does it make sense to mention LSP here? that is, LSP doesn't directly support python type hints. So it's just the python tooling that cares about the hints, whether or not it uses LSP

crates/python/qcs_sdk/__init__.pyi
```diff
@@ -258,14 +266,13 @@ class ResultData:
...
 
-    Usage
-    -----
+    ## Usage
 
-    Your usage of ``ResultData`` will depend on the types of programs you are running and where.
-    The `to_register_map()` method will attempt to build ``RegisterMap`` out of the data, where each
-    register name is mapped to a 2-dimensional rectangular ``RegisterMatrix`` where each row
+    Your usage of `ResultData` will depend on the types of programs you are running and where.
+    The `ResultData.to_register_map()` method will attempt to build `RegisterMa` out of the data, where each
```
kalzoo, line 272:
```suggestion
    The `ResultData.to_register_map()` method will attempt to build `RegisterMap` out of the data, where each
```

crates/python/qcs_sdk/__init__.pyi
```diff
@@ -201,16 +209,16 @@ class RegisterMatrix:
     Values in a 2-dimensional ``ndarray`` representing the final shot value in each memory reference across all shots.
     Each variant corresponds to the possible data types a register can contain.
 
-    Variants:
-        ``integer``: Corresponds to the Quil `BIT`, `OCTET`, or `INTEGER` types.
-        ``real``: Corresponds to the Quil `REAL` type.
-        ``complex``: Registers containing complex numbers.
+    ## Variants:
+    - ``integer``: Corresponds to the Quil `BIT`, `OCTET`, or `INTEGER` types.
```
kalzoo, line 213: the ticks weren't changed, but I'm confused about how many are needed in this new scheme. have both single and double on this line. Then, the same line in the module below is all double ticks

crates/python/qcs_sdk/__init__.pyi
```diff
@@ -306,28 +316,27 @@ class ResultData:
         ...
     def to_register_map(self) -> RegisterMap:
         """
-        Convert ``ResultData`` from its inner representation as ``QVMResultData`` or
-        ``QPUResultData`` into a ``RegisterMap``. The ``RegisterMatrix`` for each register will be
+        Convert `ResultData` from its inner representation as `QVMResultData` or
```
kalzoo, line 319: If we're going from double to single backtick, will this still render correctly in IDEs?

kalzoo (approved):
Small tweaks and some questions about backticks, but 🚢 when ready

crates/python/qcs_sdk/__init__.pyi
MarquessV: LSP supports GitHub flavored markdown which also renders ``inline code`` between double backticks. I installed VSCode to double check.

image

crates/python/README.md
MarquessV: My thinking was that "Python tooling and editors" implies both LSP servers and clients. So support could mean supporting as a client or implementing as a server. I tried to keep it broad because a lot of modern editors use LSP in a transparent way (e.g. VSCode) so many users don't even know they are using it, and other editors use a custom solution (e.g. JetBrains). I mostly wanted to express we that (1) we have type stubs and (2) a convenient way to work with the library is to figure out how to get them in your editor.

crates/python/qcs_sdk/__init__.pyi
MarquessV: With `pdoc`, paths wrapped in single backticks will attempt to link to that object in the rendered documentation. Double backticks is regular inline code. I've used double backticks here (and in one other place with the same problem) for consistency.

crates/python/README.md
MarquessV: Per-offline discussion I trimmed down this statement to generally reference that Python tooling and editors can use the type hints.

PR #415, opened 2024-01-05
feat: provide ExecutionOptions configurability as ApiExecutionOptions

### LINKED ISSUE #414: Make ExecutionOptions configurable
Update [qcs-api-client-grpc](https://crates.io/crates/qcs-api-client-grpc) to >= [`0.7.9`](https://github.com/rigetti/qcs-api-client-rust/blob/main/qcs-api-client-grpc/CHANGELOG.md#079-2024-01-05) to use `ExecutionOptions` and allow those options to be provided when executing programs.

jselig-rigetti:
Closes #414 

Allows the new proto `ExecutionOptions` to be configured, embedded within the existing `ExecutionOptions`.

### DIFF
--- a/crates/lib/src/qpu/api.rs
+++ b/crates/lib/src/qpu/api.rs
@@ -197,7 +197,7 @@ pub type QpuConnectionOptionsBuilder = ExecutionOptionsBuilder;
 ///
 /// Use [`Default`] to get a reasonable set of defaults, or start with [`ExecutionOptionsBuilder`]
 /// to build a custom set of options.
-#[derive(Builder, Clone, Debug, Default, PartialEq, Eq)]
+#[derive(Builder, Clone, Debug, Default, PartialEq)]
 pub struct ExecutionOptions {
     #[doc = "The [`ConnectionStrategy`] to use to establish a connection to the QPU."]
     #[builder(default)]
@@ -207,10 +207,13 @@ pub struct ExecutionOptions {
     timeout: Option<Duration>,
     #[doc = "Options avaialable when executing a job on a QPU, particular to the execution service's API."]
     #[builder(default = "None")]
-    api_options: Option<ApiExecutionOptions>,
+    api_options: Option<InnerApiExecutionOptions>,
 }
 
+impl Eq for ExecutionOptions {}
+
 /// Options avaialable when executing a job on a QPU, particular to the execution service's API.
+/// This is a conventent alias for [`InnerApiExecutionOptions`] which provides a builder.
 ///
 /// Use [`Default`] to get a reasonable set of defaults, or start with [`ApiExecutionOptionsBuilder`]
 /// to build a custom set of options.
@@ -240,6 +243,12 @@ impl ApiExecutionOptions {
 impl From<ApiExecutionOptions> for InnerApiExecutionOptions {
     fn from(options: ApiExecutionOptions) -> Self {
         options.inner
+    }
+}
+
+impl From<InnerApiExecutionOptions> for ApiExecutionOptions {
+    fn from(inner: InnerApiExecutionOptions) -> Self {
+        Self { inner }
     }
 }
 
@@ -274,7 +283,7 @@ impl ExecutionOptions {
 
     /// Get the [`ApiExecutionOptions`].
     #[must_use]
-    pub fn api_options(&self) -> Option<&ApiExecutionOptions> {
+    pub fn api_options(&self) -> Option<&InnerApiExecutionOptions> {
         self.api_options.as_ref()
     }
 }
--- a/crates/lib/src/qpu/api.rs
+++ b/crates/lib/src/qpu/api.rs
@@ -17,7 +17,7 @@ use qcs_api_client_grpc::{
     services::controller::{
         controller_client::ControllerClient, execute_controller_job_request,
         get_controller_job_results_request, ExecuteControllerJobRequest,
-        GetControllerJobResultsRequest,
+        ExecutionOptions as InnerApiExecutionOptions, GetControllerJobResultsRequest,
     },
 };
 pub use qcs_api_client_openapi::apis::Error as OpenApiError;
@@ -104,6 +104,9 @@ pub async fn submit(
         execution_configurations: vec![params_into_job_execution_configuration(patch_values)],
         job: Some(execute_controller_job_request::Job::Encrypted(program)),
         target: execution_options.get_job_target(quantum_processor_id),
+        options: execution_options
+            .api_options()
+            .map(|options| options.to_owned().into()),
     };
 
     let mut controller_client = execution_options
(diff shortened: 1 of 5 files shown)


crates/lib/src/qpu/api.rs
```diff
@@ -202,6 +205,59 @@ pub struct ExecutionOptions {
     #[doc = "The timeout to use for the request, defaults to 30 seconds. If set to `None`, then there is no timeout."]
     #[builder(default = "Some(Duration::from_secs(30))")]
     timeout: Option<Duration>,
+    #[doc = "If `true` and the user is authorized, jobs will bypass change protection against managed settings."]
```
MarquessV, line 208: I think this docstring needs updating

crates/lib/src/qpu/api.rs
```diff
@@ -202,6 +205,59 @@ pub struct ExecutionOptions {
...
+#[allow(clippy::module_name_repetitions)]
+pub struct ApiExecutionOptions {
+    /// the inner proto representation
+    inner: InnerApiExecutionOptions,
+}
+
+impl PartialEq for ApiExecutionOptions {
+    /// implement Eq
+    fn eq(&self, other: &Self) -> bool {
+        self.inner.bypass_settings_protection == other.inner.bypass_settings_protection
```
MarquessV, line 227: This is needed because `InnerApiExecutionOptions` doesn't impl `PartialEq`? I worry about this manual implementation falling out of sync with the available options. It seems like it would be easy to update the API client without updating this function. Any way we can avoid that?

crates/python/qcs_sdk/qpu/api.pyi
```diff
@@ -216,9 +219,44 @@ class ExecutionOptionsBuilder:
...
+    def api_options(self) -> bool:
+        raise AttributeError("api_options is not readable")
+    @api_options.setter
+    def api_options(self, api_options: ApiExecutionOptions):
+        """Execution options particular to the API call at the point of execution."""
     def build(self) -> ExecutionOptions:
         """Build the ``ExecutionOptions`` using the options set in this builder."""
 
+@final
+class ApiExecutionOptions:
```
MarquessV, line 232: This class should be named `APIExeuctionOptions` for Python reasons

crates/lib/src/qpu/api.rs
```diff
@@ -202,6 +205,59 @@ pub struct ExecutionOptions {
...
+    }
+}
+
+impl ApiExecutionOptionsBuilder {
+    /// Set the `bypass_settings_protection` value.
+    pub fn bypass_settings_protection(&mut self, bypass_settings_protection: bool) -> &mut Self {
+        self.inner
+            .get_or_insert(InnerApiExecutionOptions::default())
+            .bypass_settings_protection = bypass_settings_protection;
+        self
```
MarquessV, line 259: From a user perspective, is the difference between `ExecutionOptions` and `APIExecutionOptions` clear? For example, would a user understand why "ConnectionStrategy" is an `ExecutionOption` but not an `APIExecutionOption`? I think I see why we've nested the builders, but I wonder if specifying each option in `ExecutionOptions` directly would be better? It looks like in either case we'll have to maintain a set of functions for each option, so maybe we should opt for an approach that offers a simpler and less verbose API?

For clarity, I'm proposing getting rid of ApiExeuctionOptions and instead:

```rs
pub struct ExecutionOptions {
    // .. current options
    bypass_settings_protection: bool
}

impl From<ExecutionOptions> for InnerExecutionApiOptions {
    fn from(options: ExecutionOptions) -> Self {
         InnerExecutionApiOptions { bypass_settings_protection: ExecutionOptions.bypass_settings_protection }
    }
}
```

Does that make sense? Maybe I'm missing other important advantages for nesting the options?

crates/lib/src/qpu/api.rs
jselig-rigetti: It doesn't implement `Eq`, correct. I can update the implementation to use deconstruction so that the compiler will yell if more fields are added

```rs
    fn eq(&self, other: &Self) -> bool {
        let InnerApiExecutionOptions {
            bypass_settings_protection,
        } = self.inner;
        bypass_settings_protection == other.inner.bypass_settings_protection
    }
```

crates/lib/src/qpu/api.rs
jselig-rigetti: I agree it is confusing, and I had them combined on the first go around as you suggested, but to allow these structs to evolve independently and to allow any `impl From<T> for ApiExecutionOptions`, this seems to be more flexible. 

I'm flexible on this though, so happy to make the change considering the ergonomics.

BatmanAoD: I'm inclined to prefer a single type as well, but what's the connection to `impl From<T>`?

jselig-rigetti: Alternate implementation with a combined type: #416 - I do like that one a bit more

kalzoo: Like with TranslationOptions, I'd like to allow users to use the auto-generated code directly, because there will certainly be a delay (and potentially errors) in adding newly added gRPC execution options to any manually maintained code here.

We need to allow users to:

* provide options which are used on the _client_ (like timeout) and those used by the service (like settings protection)
* provide the service options directly (via gRPC API client) or (optionally) via more ergonomic builder.

For the Python bindings, we've accepted for the time being that there's no direct access to the gRPC fields.

crates/lib/src/qpu/api.rs
```diff
@@ -202,6 +205,62 @@ pub struct ExecutionOptions {
...
+
+impl PartialEq for ApiExecutionOptions {
+    /// implement Eq
+    fn eq(&self, other: &Self) -> bool {
+        let InnerApiExecutionOptions {
+            bypass_settings_protection,
+        } = self.inner;
+        bypass_settings_protection == other.inner.bypass_settings_protection
+    }
+}
```
Shadow53, line 232: `InnerApiExecutionOptions` implements `PartialEq` but not `Eq`, so you should be able to derive `PartialEq` instead of this manual implementation.

Should we consider removing the `Eq` (derived) impl on the types containing this one instead? `==` will work as long as `PartialEq` is implemented.

jselig-rigetti: Fixed this with https://github.com/rigetti/qcs-sdk-rust/pull/415/commits/7a3f4fd0723218329d0dd1e1fc709b6ef739a3cb - left in the following and things are working

```rs
impl Eq for ApiExecutionOptions {}
```

crates/lib/src/qpu/api.rs
MarquessV: Gotcha, I don't want to spin wheels on this, so happy to proceed with the current implementation, but I'll suggest one more alternative. My main gripe is with the nested builders, so what if we embed `ApiExecutionOptions` into `ExecutionOptions` directly and use custom setters to set options in the embedded API options, whilst still allowing it to be set directly if the user desires:

```rs
pub struct ExecutionOptions {
    // ... existing options
    
    #[builder(setter(custom))]
    bypass_settings_protection: bool
    #[builder(default)]
    api_options: ApiExecutionOptions
}

impl ExecutionOptionsBuilder {
    fn bypass_settings_protection(&mut self, value: bool) {
        self.api_options.bypass_settings_protection = value;
    }
}
```

MarquessV (approved):

crates/lib/src/qpu/api.rs
jselig-rigetti: https://github.com/rigetti/qcs-sdk-rust/pull/418 makes the argument generic, which allows direct use of  the gRPC type, but makes usage a little more awkward overall

crates/lib/src/qpu/api.rs
jselig-rigetti: Fixed with the same issue as https://github.com/rigetti/qcs-sdk-rust/pull/415#pullrequestreview-1809872395

crates/lib/src/qpu/api.rs
jselig-rigetti: I'm going to leave this as-is for now with the nested builders, I think it would be a more convenient interface but also blurs the line between which struct is which. We're not necessarily blocked from making the change in another MR though.

PR #417, opened 2024-01-08
ci: Run checks on push to main, or on pull request

MarquessV:
I believe this behavior is a bit more efficient and reflects what we want better. I believe it fixes an issue where PRs from forked repos won't run the checks.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,5 +1,9 @@
 name: Run Checks
-on: [push]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
   checks:


crates/lib/src/qpu/execution.rs
```diff
@@ -64,7 +64,7 @@ pub(crate) enum Error {
     #[error("Program when getting substitutions for program: {0}")]
     Substitution(String),
     #[error("Problem making a request to the QPU: {0}")]
-    QpuApiError(#[from] super::api::QpuApiError),
+    QpuApi(#[from] super::api::QpuApiError),
```
MarquessV, line 67: These variant names were failing clippy's [enum_variant_name](https://rust-lang.github.io/rust-clippy/master/index.html#/enum_variant) lint. Since the enum is crate private, I went ahead and renamed them.

crates/lib/src/lib.rs
```diff
@@ -30,7 +30,6 @@
     path_statements,
     patterns_in_fns_without_body,
     pointer_structural_match,
-    private_in_public,
```
MarquessV, line 33: This lint has been removed

PR #418, opened 2024-01-08
#414: use generic `ApiExecutionOptions` parameter

jselig-rigetti:
Competes with #416


jselig-rigetti:
IMO, the generic overcomplicates things. The proto struct should be the direct argument, and the `ApiExecutionOptions` should be converted before execution.

PR #420, opened 2024-01-10
chore: update to newest QCS-API types

BatmanAoD:


jselig-rigetti:
This will be done as part of #418

PR #430, opened 2024-01-26
feat: ExecutionData is now serializable.

### LINKED ISSUE #429: Make ExecutionData pickleable
It can be convenient to serialize/pickle `ExecutionData` to provide some means of persisting job results.

MarquessV:
closes #429


crates/lib/src/qpu/result_data.rs
```diff
@@ -23,7 +24,7 @@ pub enum ReadoutValues {
 }
 
 /// A row of data containing the contents of each memory region at the end of a job.
-#[derive(Debug, Clone, EnumAsInner, PartialEq)]
+#[derive(Debug, Clone, EnumAsInner, PartialEq, Deserialize, Serialize)]
 pub enum MemoryValues {
```
kalzoo, line 28: for all serialized enums, I might consider `#[serde(tag = "type", content = "content")]` which is easier to deal with in both JS and `jq`. That's stylistic though, happy to hear other opinions.

kalzoo (approved):
pickling as JSON - looks good.

One comment about serialization keys (if adopted, make sure to get all the enums) but otherwise approved.

crates/lib/src/qpu/result_data.rs
MarquessV: Good point to tag everything consistently. I do think adjacent tags are nice, but we would have to break serialization of RegisterData, as that already uses `#[serde(untagged)]`.

kalzoo: Per our chat today, I didn't realize that my request implied a breaking change, I meant to _avoid_ future breaking changes. So I'd suggest one of the following:

* Making these changes in this particular PR (since nothing relies on this yet), and accept inconsistency for the time being
* Being "consistently suboptimal" in enum tagging and following the existing convention

In both cases, this should be registered on the v1.0 wish list for later adoption.

PR #431, opened 2024-01-26
fix(lib)!: clean up Error type variants, prefer error objects to strings

### LINKED ISSUE #427: Add special error for load shedding
The SDK does not differentiate between gRPC status codes received e.g. during translation, and converts all of them [into string messages](https://github.com/rigetti/qcs-sdk-rust/blob/adf36d00deafb8e5b29cc7b0dbea68acfc93ff0d/crates/lib/src/executable.rs#L703). This makes it difficult for consumers to detect when the failure is a transient one that should succeed after a couple tries, such as when the translation backend is under load and temporarily rejects new requests.

Shadow53:
Resolves #427. See https://github.com/rigetti/qcs-sdk-rust/issues/427#issuecomment-1912571823 for why I did this instead of implementing retry logic.

There are a few main changes here, which all of which are API breaking:

1. The `QcsClient` error variant on `qpu::execution::Error` was only used for translation errors, so it was renamed to better communicate that.
2. The `Translation` variant on `executable::Error` was unused, so I changed the inner type to `GrpcClientError` from `String` and changed conversion so the above error variant becomes this one.
3. I did a brief audit of the other variants on `executable::Error` using my editor's `Find All References` feature and deleted anything that had no references.
  - For variants decorated with `#[from]`, I removed that decorator to see if anything broke due to implicit conversions no longer working. Only the variants where this did not introduce errors were deleted.

I'm open to restoring the unused variants and making them used, as appropriate. I figured that it would be good to clean things up while already making breaking changes.


crates/lib/src/executable.rs
```diff
@@ -643,29 +621,18 @@ pub enum Error {
     Compilation(String),
     /// There was a problem when translating the Quil program.
     #[error("There was a problem translating the Quil program: {0}")]
-    Translation(String),
+    Translation(GrpcClientError),
```
jselig-rigetti, line 624: Many of the other errors look to be handled within `QpuApiError` now, which also has it's own variant with the same type `QpuApiError::GrpcClientError(#[from] GrpcClientError)`

PR #435, opened 2024-02-06
fix: bump quil-rs versions

notmgsk:
quil-rs [has a fix](https://github.com/rigetti/quil-rs/pull/338) for `DEFCIRCUIT` serialization


MarquessV (approved):
Weekly cargo deny issue to fix, but looks good.

PR #436, opened 2024-02-12
chore!: Use new translation service backend for calibration programs

### LINKED ISSUE #437: Fetch calibration program using translation services gRPC endpoint
The translation module uses gRPC for all requests _except_ for fetching the calibration program. The latest release of the QCS gRPC client includes a request for getting the calibration program, we should use it.

MarquessV:
BREAKING CHANGE: The calibration program is now from `get_quilt_calibrations` as a string. The error variants have also been slightly modified.

closes #437


crates/lib/tests/mocked_qpu.rs
```diff
@@ -273,6 +275,15 @@ mod translation {
                 },
             ))
         }
+
+        async fn get_quantum_processor_quil_calibration_program(
+            &self,
+            _request: Request<GetQuantumProcessorQuilCalibrationProgramRequest>,
+        ) -> Result<Response<QuantumProcessorQuilCalibrationProgram>, Status> {
+            Ok(Response::new(QuantumProcessorQuilCalibrationProgram {
+                quil_calibration_program: "H 0\nCNOT 0 1\n".to_string(),
```
notmgsk, line 284: Not a big deal, but this might confuse someone running the tests if they see a `H 0` come back when they're expecting `DEFCAL`s etc.

notmgsk (approved):

crates/python/qcs_sdk/qpu/translation.pyi
```diff
@@ -55,26 +32,26 @@ def get_quilt_calibrations(
     quantum_processor_id: str,
     client: Optional[QCSClient] = None,
     timeout: Optional[float] = None,
-) -> QuiltCalibrations:
+) -> str:
```
Shadow53, line 35: To avoid the breaking change, couldn't we re-parse the program?

Shadow53: I guess the better question is whether we want to avoid the breaking change or not.

Shadow53 (approved):

crates/python/qcs_sdk/qpu/translation.pyi
MarquessV: `QuiltCalibrations` was derived from an OpenAPI type (and contained the program string, not a parsed Program), so I think a break is non-negotiable, since it would be weird to construct a to-be-deprecated OpenAPI struct.

crates/lib/tests/mocked_qpu.rs
MarquessV: Fair point, I put a DEFCAL in there instead.

PR #443, opened 2024-02-16
fix: Requests to cancel job no longer error due to missing target.

### LINKED ISSUE #442: Requests to cancel jobs fail due to missing targets
The proto spec for the cancel jobs request was missing a required `target` field. Without this, the request always fails. A recent update to the gRPC client adds this missing field, we need to pull that update in and add the field.

MarquessV:
closes #442


crates/lib/src/qpu/api.rs
```diff
@@ -131,9 +131,8 @@ pub async fn submit(
 /// This action is *not* atomic, and will attempt to cancel every job even when some jobs cannot be
 /// cancelled. A job can be cancelled only if it has not yet started executing.
 ///
-/// Success response indicates only that the request was received. Cancellation is not guaranteed,
-/// as it is based on job state at the time of cancellation, and is completed on a best effort
-/// basis.
+/// Cancellation is not guaranteed, as it is based on job state at the time of cancellation, and is
+/// completed on a best effort basis.
```
Shadow53, line 135: Should be explicit about whether an error is returned if a job cannot be cancelled. As a user, I would also want to know what would happen if I passed in two jobs, where one is already executing but the other can be cancelled.

Shadow53: FWIW the idiomatic thing to do is have a
```rust
/// # Errors
```
section in the comment

Shadow53: May also replace some of this comment with a "See `cancel_job`"

Shadow53 (approved):
Some notes on doc comments, but the code looks good

crates/lib/src/qpu/api.rs
MarquessV: Added a note on what errors to expect when job cancellation fails.

PR #449, opened 2024-03-05
feat: Experimental gRPC web support

MarquessV:


.github/workflows/create-linux-wheel.yml
```diff
@@ -0,0 +1,44 @@
+name: Build Linux Wheel
```
MarquessV, line 1: This workflow is a temporary measure to build a wheel for testing and should be removed before merge.

Cargo.toml
```diff
@@ -4,9 +4,9 @@ resolver = "2"
 
 [workspace.dependencies]
 qcs-api = "0.2.1"
-qcs-api-client-common = "0.7.10"
-qcs-api-client-grpc = "0.7.13"
-qcs-api-client-openapi = "0.8.11"
+qcs-api-client-common = { git = "https://github.com/rigetti/qcs-api-client-rust", branch = "grpc-web" }
+qcs-api-client-grpc = { git = "https://github.com/rigetti/qcs-api-client-rust", branch = "grpc-web" }
+qcs-api-client-openapi = { git = "https://github.com/rigetti/qcs-api-client-rust", branch = "grpc-web" }
```
MarquessV, line 9: These should be based on an actual release before merging.

crates/python/Cargo.toml
```diff
@@ -18,7 +18,7 @@ crate-type = ["cdylib", "rlib"]
 
 [dependencies]
 async-trait = "0.1.73"
-qcs = { path = "../lib", features = ["tracing"] }
+qcs = { path = "../lib", features = ["tracing", "grpc-web"] }
```
MarquessV, line 21: Right now, this makes the `grpc-web` client the only client the Python library can use. Ideally it should be opt-in, but that would come at the sacrifice of the ergonomics for the Rust crate.

crates/lib/src/client.rs
```diff
@@ -93,6 +102,8 @@ impl Qcs {
         let channel = get_channel(uri)?;
         let service =
             wrap_channel_with_retry(wrap_channel_with(channel, self.get_config().clone()));
+        #[cfg(feature = "grpc-web")]
+        let service = wrap_channel_with_grpc_web(service);
```
MarquessV, line 106: As implemented here, if a Rust user enables the "grpc-web" feature, then their clients will all be built as `grpc-web` clients. This works transparently without any additional configuration from the user. This might be a reasonable approach if we look at the Rust crate in a vacuum, but it's flawed for `qcs-sdk-python`.

In order to expose a feature like `grpc-web` to Python, we have to build the Python package with that feature enabled for everyone. This is because most users depend on a pre-built wheel being available for their platform, which means we have to build the package up front, with a static set of features. However, this would mean that all qcs-sdk-python users would be calling our services via a grpc-web proxy, which isn't desirable. 

That leads us to modifying the feature such that it enables access to the `grpc-web` client, but still requires opt-in via code. We could expose a `use_grpc_web` on the `Qcs` client, but that doesn't quite work. Each `wrap_channel_with_*` method wraps the passed in type in another type. For example, `wrap_channel_with_grpc_web(T)` returns a `GrpcWebType<T>`. This means we can't conditionally build the client based on a `use_grpc_web` flag, because if it wasn't set, the function would return `T`, but if it was set, the function would need to return `GrpcWebType<T>`, which is a trick Rust doesn't allow.

It may be possible to work around this with something like traits or generics, but I wasn't able to fenagle a solution. `tonic` has an [issue tracking implementation of a generic Client trait](https://github.com/hyperium/tonic/issues/1490) but until that exists, I don't think we can effectively solve this problem, but I would appreciate second opinions.

Here are some potential solutions I've been kicking around, curious to hear other ideas too:

* If `grpc-web` is enabled, add a `use_grpc_web` option to the `Qcs` client and an additional method that will wrap the client with the grpc-web service. Users of the `Qcs` client would be required to check the value of the option and use the new method to wrap the client if it's enabled. This is obviously not ideal and error prone. However, all the exposed methods would be `pub(crate)`, so it's still an implementation detail we have to get right, there isn't a risk to the user to misuse the client.
* Similar to the above, but make `use_grpc_web` a request level option. That is, functions like `get_quilt_calibrations` would take a `use_grpc_web` parameter. This makes the responsibility of branching based on the flag more clear, but clutters the function signatures. I think the UX isn't as good either. If I'm a user using `grpc-web`, it's more than likely because I need it for all requests, so having to set it on every request is cumbersome.

Shadow53: Is it possible to use an opaque wrapper around `Arc<Mutex<dyn GrpcService<BoxBody>>>` for the Python side? Looking at e.g. the `TranslationClient` docs, the inner client to `TranslationClient::new` is a generic `T` implementing `GrpcService<BoxBody>`. There are associated types as well, so it would really look like

```rust
// For some concrete types T, E, and F
Arc<Mutex<dyn GrpcService<BoxBody, ResponseBody=T, Error=E, Future=F>>>
```

Shadow53: So the overall structure of this feature would look like:

- The api client has a feature that enables the ability to use (but does not force) grpc-web
- The SDK (Rust) offers the same
- The client-creation methods (such as above) have a parameter to select whether to use `grpc-web` (maybe we introduce a `ClientOptions` or similar)
- The type alias `GrpcConnection` becomes `Box<dyn GrpcService<BoxBody, ResponseBody=T, Error=E, Future=F>>` for some concrete types `T`, `E`, and `F`. This should work because `GrpcService` is implemented for [all types implementing `Service`](https://docs.rs/tonic/0.9.2/tonic/client/trait.GrpcService.html#impl-GrpcService%3CReqBody%3E-for-T) with certain generics, and `Service` [is implemented](https://docs.rs/tower-service/0.3.2/tower_service/trait.Service.html#impl-Service%3CRequest%3E-for-Box%3CS%3E) for `Box<S> where S: Service` 

I believe this should work, as all of our client wrappers use the same `Pin<Box<dyn Future>>` for `F` and `E` should be the same. I'm unsure if `T` will be the same, but I think it should be.

BatmanAoD: Third option: can we just publish `qcs-sdk-python-webgrpc` as a separate PyPI package? I realize that ~doubles our Maturin build time.

MarquessV: > Third option: can we just publish `qcs-sdk-python-webgrpc` as a separate PyPI package? I realize that ~doubles our Maturin build time.

Yes, I believe we could, the two concerns that come to mind are:

1. How do we specify the dependency in downstream packages like pyQuil appropriately. E.g. it needs qcs-sdk-python or qcs-sdk-python-webgrpc. I'm not sure if there is an easy way to express that.
2. That approach doesn't scale well with additional features. If we wanted to support a second feature, we'd need 4 packages total. A third feature, 6, etc.

BatmanAoD: I guess I'm thinking of http/1.1 support less as a "feature" that makes our services better and more as a hack needed by an important client. So there'd be a separate PyPI package for `pyquil-webgrpc`, too, and we wouldn't have a matrix of feature support because this is a one-off.

That might not be the best way to look at this, though, especially since web-grpc does have the nice side-effect of giving us a way to use the API from a browser.

.github/workflows/release-python-grpc-web.yml
```diff
@@ -0,0 +1,171 @@
...
+
+  linux:
+    runs-on: ubuntu-latest
+    needs: [is-python-release, should-publish-wheels]
+    env:
+      CXXFLAGS: "-std=c++11"
+    strategy:
+      matrix:
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        target: [x86_64, aarch64]
```
MarquessV, line 80: Will make an issue to follow-up with support for PowerPC. Builds for it currently fail due to an older version of `ring` that some of our dependencies use. The latest version of ring _does_ support PowerPC though, so we should be good if we can get our dependencies to all use ring >= 0.17

.github/workflows/release-python-grpc-web.yml
MarquessV, line 1: A successful run of this workflow can be seen [here](https://github.com/rigetti/qcs-sdk-rust/actions/runs/8285060264/job/22671968423?pr=449) (sans publishing)

.github/workflows/release-python-grpc-web.yml
```diff
@@ -0,0 +1,163 @@
+name: Release Python with grpc-web clients
+
+on:
+  release:
+    types: [published]
+  push:
+    branches:
+      - 'grpc-web-rc'
```
jselig-rigetti, line 8: TODO: remove on merge

.github/workflows/release-python-grpc-web.yml
```diff
@@ -0,0 +1,156 @@
...
+    runs-on: ubuntu-latest
+    # `needs` forces this job to wait until all specified jobs
+    # are finished to run. Typically, those jobs would all have
+    # to be successful, but when combined with `if: always()`,
+    # this job is allowed to run after all the needed jobs
+    # finish, regardless of their outcome. In this case, we
+    # still make sure that at least a source distribution
+    # can be published.
+    needs: [ linux, sdist ]
+    # needs: [ macos, linux, sdist ]
```
jselig-rigetti, line 146: Does this need to depend on macos again? The latest macos build is taking over 1 hour 😞

.github/workflows/release-python-grpc-web.yml
```diff
@@ -0,0 +1,156 @@
...
+    if: always() && needs.sdist.result == 'success'
+    permissions:
+      id-token: write
+    steps:
+      - uses: actions/download-artifact@v3
+      # - name: Publish to PyPI
+      #   uses: messense/maturin-action@v1
+      #   with:
+      #     command: upload 
+      #     args: --skip-existing wheels-grpc-web/*
```
jselig-rigetti, line 156: TODO: reenable on merge

crates/lib/src/client.rs
jselig-rigetti: @BatmanAoD using a separate `pyquil-grpc-web` ultimately became the plan 👍

.github/workflows/release-python-grpc-web.yml
jselig-rigetti: Ultimately, we had to fork https://github.com/rigetti/hyper-proxy to get that dependency working. We patch it before building the python crate.

.github/action-common-python-release/action.yml
```diff
@@ -0,0 +1,72 @@
...
+      target: ${{ inputs.rust-target }}
+      profile: minimal 
+      default: true 
+  - name: Patch package metadata for grpc-web
+    shell: bash
+    if: matrix.package-name == 'qcs-sdk-python-grpc-web'
+    run: |
+      pip install toml
+      python crates/python/scripts/patch_grpc_web.py
+      cargo update
```
Shadow53, line 45: `cargo update` might update more than we need for this step, e.g. anything with "SemVer-compatible" updates. We can restrict it to specific packages using the `-p` flag.

```suggestion
      cargo update -p hyper-proxy -p qcs-sdk-python-grpc-web
```

Note: this will probably require also patching the workspace `Cargo.toml`

.github/action-common-python-release/action.yml
```diff
@@ -0,0 +1,72 @@
...
+
+  # must be build or sdist
+  - if: inputs.maturin-command == 'build'
+    name: Maturin - Build
+    uses: messense/maturin-action@v1
+    with:
+      command: ${{ inputs.maturin-command }}
+      target: ${{ inputs.maturin-target }}
+      args: -i 3.8 3.9 3.10 3.11 3.12 --release --manifest-path crates/python/Cargo.toml --out dist ${{ inputs.package-name == 'qcs-sdk-python-grpc-web' && '-F grpc-web' || '' }}
```
Shadow53, line 54: ???

```suggestion
      args: -i ${{ inputs.python-version }} --release --manifest-path crates/python/Cargo.toml --out dist ${{ inputs.package-name == 'qcs-sdk-python-grpc-web' && '-F grpc-web' || '' }}
```

.github/workflows/release-python.yml
```diff
@@ -30,37 +30,18 @@ jobs:
       CXXFLAGS: "-std=c++11 -stdlib=libc++"
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        package-name: [qcs-sdk-python, qcs-sdk-python-grpc-web]
```
Shadow53, line 33: I think you still want `python-version`, right?

crates/lib/src/client.rs
```diff
@@ -20,11 +22,19 @@ pub use qcs_api_client_common::configuration::LoadError;
...
 /// initialized by this library. This ensures that the exact Tonic version used for such clients
 /// matches what this library uses.
 pub type GrpcConnection = RetryService<RefreshService<Channel, ClientConfiguration>>;
+#[cfg(feature = "grpc-web")]
+/// A type alias for the underlying gRPC connection used by all gRPC clients within this library.
+/// It is public so that users can create gRPC clients with different APIs using a "raw" connection
+/// initialized by this library. This ensures that the exact Tonic version used for such clients
+/// matches what this library uses.
+pub type GrpcConnection =
+    GrpcWebWrapperLayerService<RetryService<RefreshService<Channel, ClientConfiguration>>>;
```
Shadow53, line 37: Stylistic thing, but I prefer `#[cfg()]` between docs and the annotated thing. Makes it easier to tell what is being annotated.

.github/action-common-python-release/action.yml
jselig-rigetti: The `-i` argument uses one command to build all the variants, instead of creating a separate matrix entry for each version. Seems a little bit less resource intensive this way, unless you think the previous way is better?

.github/workflows/release-python.yml
jselig-rigetti: Not necessary with https://github.com/rigetti/qcs-sdk-rust/pull/449#discussion_r1548734359

crates/lib/src/client.rs
jselig-rigetti: Done

.github/action-common-python-release/action.yml
jselig-rigetti: Couldn't use `-p` and `SPEC` at the same time, so changed to `cargo update hyper-proxy`

.github/action-common-python-release/action.yml
jselig-rigetti: These are the contents in the `qcs-sdk-python-grpc-web` artifacts, juts to prove that all python versions are being built

image

I'm going to resolve this comment, but please feel free to unresolve.

.github/action-common-python-release/action.yml
```diff
@@ -0,0 +1,76 @@
...
+  - if: inputs.maturin-command == 'sdist'
+    name: Maturin - Source Distribution
+    uses: messense/maturin-action@v1
+    with:
+      manylinux: '2_28'
+      command: ${{ inputs.maturin-command }}
+      target: ${{ inputs.maturin-target }}
+      args: --manifest-path crates/python/Cargo.toml --out dist
+
+  # host will not produce an installable wheel in this case
```
MarquessV, line 66: Just for clarity, the issue is that the wheel isn't installable on the host because it's for a different architecture, not that the wheel isn't installable.

MarquessV (commented):
I can't approve because I opened the MR originally, but this is looking good. Was there a test run of the workflow that validated everything built correctly?

.github/action-common-python-release/action.yml
jselig-rigetti: Correct. I was able to verify that an installation of the wheel from https://github.com/rigetti/qcs-sdk-rust/actions/runs/8548321434 produced a valid and usable python dep.

jselig-rigetti:
> I can't approve because I opened the MR originally, but this is looking good. Was there a test run of the workflow that validated everything built correctly?

Yes, I validated the wheels produced from https://github.com/rigetti/qcs-sdk-rust/actions/runs/8548321434

.github/action-common-python-release/action.yml
Shadow53:
```suggestion
  # host will not be able to install wheel in this case
```

.github/action-common-python-release/action.yml
jselig-rigetti: Good point - removed `input.python-version`

PR #452, opened 2024-04-03
open up tonic version range to include 0.11

kalzoo:
While `tonic` remains pre-1.0 and this is a sweep of major versions, they are all compatible with this package and can help resolve the confusing type errors which arise when consumers rely on distinct major versions of `tonic`.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -9,6 +9,9 @@ jobs:
   checks:
     name: Run Checks
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


Shadow53 (approved):
Looks fine to me, but CI checks failed, so I'm not seeing confirmation that this will compile.

Also, can we make sure this works with tonic 0.9, 0.10, and 0.11 in CI using a matrix or something? Don't want us to get burned by users using a version of tonic that isn't actually compatible with this library.

jselig-rigetti:
https://github.com/rigetti/qcs-sdk-rust/pull/449 requires particularly on `0.11.0`, not the permissive range here, so recommend to close this MR.

PR #463, opened 2024-04-19
feat!: Make Python bindings public

MarquessV:
In an effort to make the bindings and other code from the SDK more re-usable, I'm adjusting visibility and refactoring some of the code in both crates.


crates/lib/src/qpu/api.rs
```diff
@@ -435,11 +437,21 @@ pub enum ConnectionStrategy {
     EndpointId(String),
 }
 
-/// Methods that help select and configure a controller service client given a set of
-/// [`ExecutionOptions`] and QPU ID.
-impl ExecutionOptions {
+/// An ExecutionTarget provides methods for provided the appropriate connection to the execution
```
BatmanAoD, line 440:
```suggestion
/// An ExecutionTarget provides methods to establish the appropriate connection to the execution
```

crates/lib/src/qpu/api.rs
```diff
@@ -466,8 +479,9 @@ impl ExecutionOptions {
         }
     }
 
+    /// Get the [`cancel_controller_jobs_request::Target`] for the given quantum processor ID.
```
BatmanAoD, line 482: Side question: it wouldn't be possible to force `prost` to only generate one `Target` type, would it?

crates/lib/src/qpu/api.rs
```diff
@@ -690,7 +720,10 @@ pub enum QpuApiError {
         /// The message associated with the failed job.
         message: String,
     },
-
+    /// Error that can occur when the gRPC status code could not be decoded.
```
BatmanAoD, line 723:
```suggestion
    /// Error that can occur when the gRPC status code cannot be decoded.
```

crates/lib/src/qpu/api.rs
```diff
@@ -690,7 +720,10 @@ pub enum QpuApiError {
...
+    #[error("The status code could not be decoded: {0}")]
+    StatusCodeDecode(String), // TODO: This error is in prost. Should we really use that as a dep
```
BatmanAoD, line 725: Stale comment, or is `prost` imported somewhere I'm not seeing?

crates/python/Cargo.toml
```diff
@@ -40,6 +40,7 @@ once_cell = "1.18.0"
 opentelemetry = { version = "0.23.0" }
 opentelemetry_sdk = { version = "0.23.0" }
 tracing = { version = "0.1.37" }
+prost = "0.12.6"
```
BatmanAoD, line 43: Possibly not necessary any more

BatmanAoD (approved):

crates/lib/src/qpu/api.rs
MarquessV: Stale comment, removed

crates/python/Cargo.toml
MarquessV: Good thing to check. However, we do still need prost to access the Message trait.

crates/lib/src/qpu/api.rs
MarquessV: I think so? Each request type embeds its own `Target`, which results in a different model being exported for each. I'd assume you could lift that specification up and re-use across each request type.

.github/workflows/checks.yml
```diff
@@ -20,22 +20,19 @@ jobs:
...
-
       - uses: Swatinem/rust-cache@v2
       - name: Install cargo-make
         uses: actions-rs/cargo@v1
         with:
           command: install
           args: --debug cargo-make
-      - name: Install poetry
-        uses: snok/install-poetry@v1
       # may protect against linking failure: no space left on device
```
BatmanAoD, line 34: This comment should stay with the "Delete huge..." task, presumably

crates/python/pyproject.toml
```diff
@@ -20,45 +20,33 @@ classifiers = [
   "Programming Language :: Python :: 3.11",
   "Operating System :: OS Independent",
 ]
-dependencies = ["quil>=0.11.2"]
-
-# PEP 621 specifies the [project] table as the source for project metadata. However, Poetry only supports [tool.poetry]
-# We can remove this table once this issue is resolved: https://github.com/python-poetry/poetry/issues/3332
```
BatmanAoD, line 26: Are you sure we can get rid of this? Treeline is stuck on an older version of Poetry for... reasons 😞

BatmanAoD: Per offline discussion: this should be fine

BatmanAoD (approved):

BatmanAoD (approved):

PR #468, opened 2024-05-09
feat: increase gRPC max message size from 4MB to 50MB

### LINKED ISSUE #467: Increase gRPC max message size limit for translation
The [default max message size](https://docs.rs/tonic/latest/tonic/#max-message-size) is 4MB. To account for the possibility of larger translation requests/responses, we should increase that limit.

notmgsk:
Increases the gRPC max message size from 4MB (the `tonic` default) to 50MB. This is motivated by some users having hit the 4MB limit.

This MR also fixes a number of new clippy lints that were causing CI to fail.

Closes #467.

### DIFF
--- a/crates/lib/src/client.rs
+++ b/crates/lib/src/client.rs
@@ -22,6 +22,9 @@ pub use qcs_api_client_common::configuration::LoadError;
 pub use qcs_api_client_grpc::channel::Error as GrpcError;
 pub use qcs_api_client_openapi::apis::Error as OpenApiError;
 
+const DEFAULT_MAX_MESSAGE_ENCODING_SIZE: usize = 50 * 1024 * 1024;
+const DEFAULT_MAX_MESSAGE_DECODING_SIZE: usize = 50 * 1024 * 1024;
+
 /// A type alias for the underlying gRPC connection used by all gRPC clients within this library.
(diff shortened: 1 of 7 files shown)


crates/lib/src/client.rs
```diff
@@ -22,6 +22,9 @@ pub use qcs_api_client_common::configuration::LoadError;
 pub use qcs_api_client_grpc::channel::Error as GrpcError;
 pub use qcs_api_client_openapi::apis::Error as OpenApiError;
 
+const DEFAULT_MAX_MESSAGE_ENCODING_SIZE: usize = 50 * 1024 * 1024;
+const DEFAULT_MAX_MESSAGE_DECODING_SIZE: usize = 50 * 1024 * 1024;
```
MarquessV, line 26: We have a similar constant in [the qpu api](https://github.com/rigetti/qcs-sdk-rust/blob/main/crates/lib/src/qpu/api.rs#L42). Should we try to consolidate around one?

MarquessV (approved):
One thought, but happy to see this merged with whatever you decide!

crates/lib/src/client.rs
notmgsk: I did see that, but opted for a module-specific one simply because the conserv-related constant is set to 250MB which seemed unnecessarily high for translation messages.

PR #486, opened 2024-07-29
feat(python): ConnectionStrategy supports the pickle module

MarquessV:
This is to support rigetti/pyQuil#1719

### DIFF
--- a/crates/python/src/qpu/api.rs
+++ b/crates/python/src/qpu/api.rs
@@ -612,12 +612,6 @@ impl PyConnectionStrategy {
         }
     }
 
-    // // Implementing this method is required for the pickle module to recognize the class as
-    // // pickleable, but __reduce__ is self sufficient, so we can just return an empty dictionary here.
-    // fn __getstate__<'py>(&'py self, py: Python<'py>) -> &'py PyDict {
-    //     PyDict::new(py)
-    // }
-    //
     fn __reduce__(&self, py: Python<'_>) -> PyResult<PyObject> {
         Ok(match self.as_inner() {
             ConnectionStrategy::Gateway => PyTuple::new(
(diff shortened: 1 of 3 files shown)


crates/python/src/qpu/api.rs
```diff
@@ -608,4 +611,41 @@ impl PyConnectionStrategy {
             _ => py.NotImplemented(),
         }
     }
+
+    // // Implementing this method is required for the pickle module to recognize the class as
+    // // pickleable, but __reduce__ is self sufficient, so we can just return an empty dictionary here.
+    // fn __getstate__<'py>(&'py self, py: Python<'py>) -> &'py PyDict {
+    //     PyDict::new(py)
+    // }
+    //
```
antalsz, line 620: Can this be deleted?

crates/python/tests/qpu/test_api.py
```diff
@@ -44,3 +47,15 @@ def test_submit_retrieve(
 
     job_id = submit(program, memory, quantum_processor_id)
     results = retrieve_results(job_id)
+
+class TestPickle():
+    @pytest.mark.parametrize("strategy", [ConnectionStrategy.gateway(), ConnectionStrategy.direct_access(), ConnectionStrategy.endpoint_id("endpoint_id")])
+    def test_connection_strategy(self, strategy: ConnectionStrategy):
+        pickled = pickle.dumps(strategy)
+        unpickled = pickle.loads(pickled)
```
antalsz, line 55: Unless I'm missing something about pytest, this test is missing an `assert`

antalsz (changes requested):
Looks like there's a missing assert in one of the tests, but the implementations look good.  Also the branch name is *incredible*.

crates/python/src/qpu/api.rs
MarquessV: Sure can

crates/python/tests/qpu/test_api.py
MarquessV: Oops, good catch

PR #502, opened 2024-10-03
feat!: upgrade to http=1

BatmanAoD:
Upgrades networking & QCS libraries.

`warp` [hasn't upgraded to `hyper=1` yet](https://github.com/seanmonstar/warp/pull/1090), but it's only a dev-dependency (and is explicitly denied in our `deny.toml`); to confirm that `http=0.2` is no longer a dependency for the actual build, run:

```
❯ cargo tree -e no-dev -i http@0.2
warning: nothing to print.
```

Note: I think this works on its own, but I'd prefer to merge this first and update one minor version further: https://gitlab.com/rigetti/qcs/clients/qcs-api-client-rust/-/merge_requests/133


BatmanAoD:
The pipeline behavior once this branch gets merged to `main`, using the validator, looks right to me: https://gitlab.com/rigetti/share/service-model/-/ci/editor?branch_name=update-tonic-and-prost&tab=2

MarquessV (approved):
🚀

PR #506, opened 2024-10-11
Fix rust pyo3 publish

BatmanAoD:
Also adds the release dry-run task to PRs.

...I don't think this is quite correct yet, though. See #507.


BatmanAoD:
No longer necessary, thanks to #571

PR #509, opened 2024-10-23
feat!: Shot count is now represented as an unsigned 32-bit, up from 16-bit, allowing for higher shot counts

### LINKED ISSUE #332: `NonZeroU16` is unecessarily restrictive for QVM trials
The `NonZeroU16` type we use to specify QVM trials has a max of 65,535. However, the QVM seems to use a standard [Common Lisp integer](https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node17.html) which imposes no upper limit on an integer. We should use a larger `NonZeroUXX` type so it's possible to specify more trials.

MarquessV:
closes #332


crates/python/src/executable.rs
```diff
@@ -108,7 +108,7 @@ impl PyExecutable {
         quil: String,
         registers: Vec<String>,
         parameters: Vec<PyParameter>,
-        #[pyo3(from_py_with = "crate::from_py::optional_non_zero_u16")] shots: Option<NonZeroU16>,
+        #[pyo3(from_py_with = "crate::from_py::optional_non_zero_u16")] shots: Option<NonZeroU32>,
```
Shadow53, line 111: Linting failure on `crate::from_py::optional_non_zero_u16`. Should probably be this?

```suggestion
        #[pyo3(from_py_with = "crate::from_py::optional_non_zero_u32")] shots: Option<NonZeroU32>,
```

crates/python/src/qvm/api.rs
```diff
@@ -123,7 +123,7 @@ impl PyMultishotRequest {
 
     #[setter]
     pub fn set_trials(&mut self, trials: u16) -> PyResult<()> {
-        // `NonZeroU16` doesn't implement `PyClass`, so `pyo3` doesn't allow it to be used
+        // `NonZeroU32` doesn't implement `PyClass`, so `pyo3` doesn't allow it to be used
```
Shadow53, line 126: Looks like `u16` is being used in this area (`fn trials` and `fn set_trials`) when it should be `u32`

(Commenting here because GH does not let me comment on non-diff lines)

Shadow53: This also happens below in functions with the same names.

Shadow53 (changes requested):
Looks like a few `u16`s were missed, at least according to linting results.

Issue #510, opened 2024-11-13
THIRDPARTY file for dependency licenses


BatmanAoD:
We need to publish license-data for third-party Rust code as part of our PyPI artifacts.

PR #511, opened 2024-11-13
fix: #510: third-party licenses

BatmanAoD:
In this PR:

- [x] Generated and reviewed licensing data for third-party packages
- [x] Added license data for Rust itself
- [x] Added `NOTICE` files for Apache-2.0 projects, where present
- [x] Resolved open questions with Luke
- [x] Bundled the license texts with our binary artifacts that we distribute via PyPI
- [x] Resolved CI issues for `cargo deny` and Python-publishing

TBD, not necessarily in this PR:
* rendering this as Markdown or similar
* automatically updating the `THIRDPARTY.yml` file in each MR


.gitignore
```diff
@@ -37,3 +37,6 @@ crates/python/tests/__output__/
 
 # unversioned developer notes
 scratch/
+
+# un-consolidated updates to THIRDPARTY license data
+/THIRDPARTY-update.yaml
```
BatmanAoD, line 42: This is to facilitate re-running `cargo bundle-licenses` and comparing the output with the existing output, even though there's not yet any tooling to assist with doing so.

manual_changes_thirdparty.txt
BatmanAoD, line 1: This file is just notes from my review. I think we should keep this until we find a good way to automate the procedure.

kalzoo:
> rendering this as Markdown or similar

what's the interest in doing this? Do we expect anyone to care about the readability?

> bundling the license texts with our binary artifacts that we distribute via PyPI

This should be as simple as listing the files in the pyproject.toml

> automatically updating the THIRDPARTY.yml file in each MR

Seems like this might be a release CI step rather than in the MR, right? Things always get a little messy when the MR pipeline commits files back to its branch.

Seems like the way to go would be:

* Have a manually-written file for any exceptions that the bundling tool can't handle
* Auto-generate the primary thirdparty file in CI (and don't commit to source control, since that'll never get updated)
* (TBD) some way to detect new gaps between the two

Then those artifacts are always up to date for publication and no one has to edit them or wonder if they are supposed to edit them.

BatmanAoD:
> Do we expect anyone to care about the readability?

Well, Luke hasn't actually commented on the draft version I sent him back when I first discovered `bundle-licenses`, and the example third-party files I've seen do seem [more readable](https://docs.nvidia.com/cuopt/user-guide/third-party-licenses.html). But the YAML is readable enough except for cases where it renders the Apache license as one long in-line string with `\n` escapes. (For some reason it does this for some packages but not others.)

> Seems like this might be a release CI step rather than in the MR, right? 

I think we always want to be able to review changes made by our tooling, and do so before merging to `main`, or at least before doing a release.

> (TBD) some way to detect new gaps between the two

I mentioned this in slack as well, but `bundle-licenses` _is_ designed to do that; the [bug I linked](https://github.com/sstadick/cargo-bundle-licenses/issues/40) causes a false-positive, though.

crates/python/scripts/patch_grpc_web.py
```diff
@@ -29,11 +28,4 @@ def write(f: TextIOWrapper, data):
...
     data = toml.load(f)
     data["package"]["name"] = "qcs-sdk-python-grpc-web"
-    write(f, data)
-
-# Patch the `hyper-proxy` dependency
-
-with open(join(workspace_path, "Cargo.toml"), "r+") as f:
-    data = toml.load(f)
-    data["patch"] = {"crates-io": {"hyper-proxy": {"git": "https://github.com/rigetti/hyper-proxy"}}}
-    write(f, data)
```
BatmanAoD, line 39: We no longer depend on `hyper-proxy`

BatmanAoD:
> This should be as simple as listing the files in the pyproject.toml

I've added it for `sdist` and `whl` files. I'm struggling to test that CI produces the right thing (and I needed to fix an unrelated issue with the `grpc-web` bundling), but locally I can see that it does.

PR #513, opened 2024-11-20
Update qcs api client

erichulburd:
* Update the QCS API client to the latest.
* Expose Q-CTRL compiling to client.


MarquessV (commented):
There is a conflict with `deny.toml`, but after that is resolved LGTM

PR #515, opened 2024-11-25
Expose hardware prng sequence

erichulburd:
This PR supports two new features:

* A `CALL choose_random_real_sub_regions` instruction.
* A `RandomizedMeasurement` struct / class that will add per-shot randomized measurements to a program, as well as support the client in ascertaining which random unitary was played on each qubit per shot.

The documentation should stand on its own here. One note, I've put all of this functionality into `qpu/experimental` (both in Rust and Python). It is experimental both in the sense that it is an experimental feature in the QCS SDK that should not be considered stable as well as in the sense that it supports features relevant to QIS experimentalists.

Review Guidance:

1. Start with `crates/lib/src/qpu/experimental/random.rs`.
2. `crates/lib/src/qpu/experimental/randomized_measurements.rs` measurements builds off the primitives defined in the file above.
3. Stubs for these features are defined in `crates/python/qcs_sdk/qpu/experimental.pyi`. Their implementation is defined in `crates/python/src/qpu/experimental.pyi`.
4. There are corresponding tests for randomized measurements in `crates/lib/src/qpu/experimental/randomized_measurements.rs` and `crates/python/tests/qpu/test_experimental.py`.


CODEOWNERS
```diff
@@ -0,0 +1 @@
+crates/lib/src/qpu/experimental @erichulburd
```
erichulburd, line 1: I don't have push permissions to this repository, which is probably related to the error reported by Github here.

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+
+/// Hardware multiplication currently uses the lower 16 bits of
+/// the PRNG value.
+const MAX_UNSIGNED_MULTIPLIER: u64 = 0x0000_0000_FFFF;
+
+/// A valid seed value that may be used to initialize the PRNG. Such
+/// values are in the range `[1, MAX_SEQUENCER_VALUE]` and are losslessly
+/// convertible to `f64`.
+#[derive(Debug, Clone, Copy)]
+pub struct PrngSeedValue {
```
erichulburd, line 147: Is this type too specific to the task at hand - should we instead support a more general `SequencerValue`?

I lean "no" here. The seed value is a sequencer value with an additional validation (cannot be equal to zero). A sequencer value is too generic in nature here and does not have a known use case other than this `PrngSeedValue`. As I see it, should we want to support a generic sequencer value in the future, we would implement `TryFrom<SequencerValue> for PrngSeedValue` and `PrngSeedValue` would remain in the public interface defined here.

crates/lib/tests/prng_test_cases.json
```diff
@@ -0,0 +1,339 @@
+{
```
erichulburd, line 1: These are values that have been validated as final memory read off Rigetti QPUs.

crates/python/qcs_sdk/qpu/experimental.pyi
```diff
@@ -0,0 +1,165 @@
...
+
+    The following example declares a source with 12 real values, representing 4 sub-regions
+    each of size 3. The destination memory region is declared with 3 real values, representing
+    a single sub-region. The seed memory region is declared with a single integer value.
+
+    The `Call` instruction will pseudo-randomly choose a sub-region of the source memory region
+    and copy it to the destination memory region according to the seed value. The seed value is
+    mutated to the next element in the PRNG sequence.
+
+    >>> program = Program()
```
erichulburd, line 47: I added this doctest as an example because we cannot support a more polished interface (due to the fact that we cannot include `quil-rs` types in the Python interface). Unfortunately, pytest won't actually _run_ this doctest (it doesn't natively support stub files).

One alternative I considered was implementing some high level functions for using `ChooseRandomRealSubRegions` natively in Python by importing `quil-py` types. I stopped short there because there isn't really precedent, however, I think it's probably the best tact.

crates/python/qcs_sdk/qpu/experimental.pyi
```diff
@@ -0,0 +1,165 @@
...
+    @staticmethod
+    def from_zxzxz(inner: NDArray[np.float64]) -> "UnitarySet":
+        """
+        Initialize a `UnitarySet` as a ZXZXZ decomposition. The input array must have shape (n, 3),
+        where n is the number of unitaries in the set.
+        """
+        ...
+
+@final
+class RandomizedMeasurement:
```
erichulburd, line 133: This class is a `qcs-sdk` substitute for `quil-py`'s `Measurement`. Again, we can't include `quil-rs` types in the interface without providing a `quil-py` module within `qcs-sdk`.

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+        Ok(quil_rs::instruction::ExternSignature::new(None, parameters))
+    }
+}
+
+/// Hardware values are 48 bits long.
+const MAX_SEQUENCER_VALUE: u64 = 0xFFFF_FFFF_FFFF;
+
+/// Hardware multiplication currently uses the lower 16 bits of
+/// the PRNG value.
+const MAX_UNSIGNED_MULTIPLIER: u64 = 0x0000_0000_FFFF;
```
Shadow53, line 141: More style nits, but can we move consts to the top of the file? E.g., `MAX_SEQUENCER_VALUE` is used by `Error` but it isn't clear where to look for the definition.

Shadow53: While I'm nitpicking (sorry), can we add the highest two bytes to these numbers? I initially read `MAX_SEQUENCER_VALUE` as equivalent to `u64::MAX` because it's all Fs.

```suggestion

/// Hardware values are 48 bits long.
const MAX_SEQUENCER_VALUE: u64 = 0x0000_FFFF_FFFF_FFFF;

/// Hardware multiplication currently uses the lower 16 bits of
/// the PRNG value.
const MAX_UNSIGNED_MULTIPLIER: u64 = 0x0000_0000_0000_FFFF;
```

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+
+/// A valid seed value that may be used to initialize the PRNG. Such
+/// values are in the range `[1, MAX_SEQUENCER_VALUE]` and are losslessly
```
Shadow53, line 144: I believe the `Error` above uses an open maximum (`)` instead of `]`)

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+/// convertible to `f64`.
+#[derive(Debug, Clone, Copy)]
+pub struct PrngSeedValue {
+    as_u64: u64,
+    pub(super) as_f64: f64,
```
Shadow53, line 149: For extra data safety, I'd keep both of these private and then use accessor methods. Otherwise, we could potentially add code that modifies `as_f64` without modifying `as_u64`

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+            })
+        } else {
+            Err(Error::InvalidSeed(value))
+        }
+    }
+}
+
+fn lfsr_next(seed: u64, taps: &[u32]) -> u64 {
+    let feedback_value = taps.iter().fold(0, |acc, tap| {
+        let base = 2u64.pow(*tap - 1);
```
Shadow53, line 173: What if `tap` is 0?

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+        let bit = u64::from((seed & base) != 0);
+        acc.bitxor(bit)
+    });
+    (seed << 1) & MAX_SEQUENCER_VALUE | feedback_value
```
Shadow53, line 177: Maybe not a problem, but I don't know the precedence of `&` vs `|`. I presume this works as expected, but parens would be nice for people like me so they don't have to look it up.

```suggestion
    ((seed << 1) & MAX_SEQUENCER_VALUE) | feedback_value
```

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+/// This represents the LFSR currently implemented on Rigetti control systems. Specifically,
+/// it implements a 48-bit LFSR with taps at indices 48, 47, 21, and 20.
```
Shadow53, line 181: If the taps ever change, I presume we'll make a new `lfsr_v2_next`?

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+#[must_use]
+pub fn lfsr_v1_next(seed: PrngSeedValue) -> u64 {
+    lfsr_next(seed.as_u64, &[48, 47, 21, 20])
```
Shadow53, line 184: I have seen these numbers multiple times, so I think they should be a `const`. (See previous note about `const`s at the top of the file)

```
const V1_TAPS: [u32; 4] = [48, 47, 21, 20];
```

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+    let range = start_index..(start_index + series_length);
+    let mut collection = vec![];
+    for i in 0..(start_index + series_length) {
+        lfsr = lfsr_next(lfsr, &[48, 47, 21, 20]);
+        if range.contains(&i) {
+            collection.push(lfsr);
+        }
+    }
+    collection
+}
```
Shadow53, line 199: Looking forward a bit:

It looks like we expect the possibility of further LFSR revisions, since we have a `v1` naming scheme here. Would it be worth refactoring this code to make it easier to add a `v2`? Thinking something like

```rust
trait LfsrVersion {
    const TAPS: &[u32];
}

fn lfsr_next<LFSR: LfsrVersion>(seed: PrngSeedValue) -> u64 {
    // Same as implemented here, but replace `taps` with `LFSR::TAPS`.
}

fn generate_lfsr_sequence<LFSR: LfsrVersion>(seed: u64, start_index: u32, series_length: u32) -> Vec<u64> {
    // Same as implemented here, but call `lfsr_next::<LFSR>(lfsr)` instead.
}
```

(Alternatively, use an `enum` with an associated `taps(&self)` method, but the idea is the same)

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+
+fn prng_value_to_sub_region_index(value: u64, sub_region_count: u8) -> u8 {
+    ((value & MAX_UNSIGNED_MULTIPLIER) % u64::from(sub_region_count))
```
Shadow53, line 202: Should it be an error if `value > MAX_UNSIGNED_MULTIPLIER`?

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+        .to_u8()
+        .expect("modulo u8 should always produce a valid value")
+}
+
+/// Given a seed, start index, series length, and sub-region count, this function
+/// will generate the sequence of pseudo-randomly chosen indices Rigetti control
+/// systems.
```
Shadow53, line 209: I think something may have been lost here. The latter part doesn't make sense

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+///
+/// For instance, if the following Quil program is run for 100 shots:
+///
+/// ```quil
+/// # presumed sub-region size is 3.
+/// DECLARE destination REAL[6] # prng invocations per shot = (6 / sub_region_size)  = 2
+/// DELCARE source REAL[12]     # implicit sub-region count = (12 / sub_region_size) = 4
```
Shadow53, line 216: Typo

```suggestion
/// DECLARE source REAL[12]     # implicit sub-region count = (12 / sub_region_size) = 4
```

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+/// DECLARE seed INTEGER[1]
+/// DECLARE ro BIT[1]
+///
+/// DELAY 0 1e-6
+///
+// PRAGMA EXTERN choose_random_real_sub_regions "(destination : mut REAL[], source : REAL[], sub_region_size : INTEGER, seed : mut INTEGER)"
+// CALL choose_random_real_sub_regions destination source 3 seed
```
Shadow53, line 223:
```suggestion
/// PRAGMA EXTERN choose_random_real_sub_regions "(destination : mut REAL[], source : REAL[], sub_region_size : INTEGER, seed : mut INTEGER)"
/// CALL choose_random_real_sub_regions destination source 3 seed
```

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+/// ```
+///
+/// with a seed of 639523, you could backout the randomly chosen sub-regions with the following:
+///
+/// ```rust
+/// let seed = 639523;
```
Shadow53, line 229: Add a `use` statement for `choose_random_real_sub_region_indices`. If this code should not fail while running, it'll be a successful doctest. Otherwise, replace `rust` with `no_run` so it'll do compilation without running

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+        .map(|&value| prng_value_to_sub_region_index(value, sub_region_count))
+        .collect()
+}
+
+#[cfg(test)]
+mod tests {
+    use std::{collections::HashMap, fs::File};
+
+    fn prng_sequences() -> HashMap<u32, Vec<(u64, u64)>> {
+        serde_json::de::from_reader(File::open("tests/prng_test_cases.json").unwrap()).unwrap()
```
Shadow53, line 255: Others can weigh in, but I believe we generally have a convention of test data files living outside of `src` and then being referenced as e.g.

```rust
concat!(env!("CARGO_MANIFEST_DIR"), "/data/tests/prng_test_cases.json")
```
```

**Review by Shadow53 — COMMENTED** [review]
Partial review, will do more later.

File: crates/lib/src/qpu/experimental/randomized_measurements.rs [hunk 20]
```diff
@@ -0,0 +1,628 @@
...
+    ///
+    /// The unitaries are stored in a 2D array where each row represents
+    /// a single unitary expressed as the three RZ angles.
+    Zxzxz(Array2<f64>),
+}
+
+impl UnitarySet {
+    /// Attempt to create a new instance of [`UnitarySet`] from a 2D array
+    /// of unitaries. The array must have three columns.
+    pub fn try_new_zxzxz(unitaries: Array2<f64>) -> Result<UnitarySet> {
```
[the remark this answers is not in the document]
Shadow53: One way to avoid it would be to make `UnitarySet` a private type (and rename to something like `UnitarySetInner`), and make a `pub struct UnitarySet(UnitarySetInner);` as an opaque wrapper around it.

crates/python/qcs_sdk/_tracing_subscriber/common/__init__.pyi
```diff
@@ -12,14 +12,15 @@
...
 @final
 class InstrumentationLibrary:
     """
     Information about a library or crate providing instrumentation.
-
+    
     An instrumentation library should be named to follow any naming conventions
     of the instrumented library (e.g. 'middleware' for a web framework).
-
+
```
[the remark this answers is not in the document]
Shadow53: That feeling is stronger after seeing trailing spaces in the other files. Commenting here instead of marking each occurrence.

crates/lib/src/qpu/experimental/random.rs
erichulburd: Is it `use super:: choose_random_real_sub_region_indices`?

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
+//! This module supports low-level primitives for randomization on Rigetti's QPUs.
+use std::{convert::TryFrom, ops::BitXor};
+
+use num::{complex::Complex64, ToPrimitive};
+use quil_rs::instruction::{Call, CallError, ExternError, UnresolvedCallArgument};
+
+use crate::qpu::externed_call::ExternedCall;
+
+/// An error that may occur when simulating control system extern
```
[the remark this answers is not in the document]
erichulburd: Reworded

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// An [`ExternedCall`] that may be used to select one or more random
+/// sub-regions from a source array of real values to a destination array.
+#[derive(Debug, Clone)]
+pub struct ChooseRandomRealSubRegions {
+    destination: String,
+    source: String,
+    sub_region_size: Complex64,
+    seed: String,
```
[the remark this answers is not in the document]
erichulburd: I've added validation and more specific types to the constructor. Also struct names are clarified.

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -0,0 +1,272 @@
...
+    type Error = Error;
+
+    const NAME: &str = "choose_random_real_sub_regions";
+
+    #[allow(clippy::doc_markdown)]
+    /// Build the signature for the `PRAGMA EXTERN choose_random_real_sub_regions`
+    /// instruction. The signature is:
+    ///
+    /// rust ignore
+    ///     "(destination : mut REAL[], source : REAL[], sub_region_size : INTEGER, seed : mut INTEGER)"
```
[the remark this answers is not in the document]
erichulburd: This trips up either tests or clippy without `rust ignore`. I've changed it to ```text instead.

crates/python/qcs_sdk/_tracing_subscriber/common/__init__.pyi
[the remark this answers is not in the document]
erichulburd: This is actually pretty annoying.

Basically, every time a Python build runs, the stubs are re-written, but not formatted. You have to run `cargo make format-tracing-subscriber` to format them.

I think the solution here would be to remove the stub files from version control and make sure to format prior to publication.

crates/lib/src/qpu/experimental/randomized_measurements.rs
```diff
@@ -0,0 +1,628 @@
...
+
+        let expected_program = Program::from_str(BASE_QUIL_PROGRAM_WITH_MEASUREMENTS)
+            .expect("must be valid Quil program");
+
+        assert_eq!(randomized_program, expected_program);
+    }
+
+    #[fixture]
+    fn seeds() -> Vec<u64> {
+        vec![463_692_700, 733_101_278, 925_742_198]
```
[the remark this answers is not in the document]
erichulburd: Left a docstring.

crates/lib/src/qpu/experimental/randomized_measurements.rs
```diff
@@ -0,0 +1,628 @@
...
+            unitary_set,
+            unitary_count_as_u8,
+            qubit_randomizations,
+        })
+    }
+}
+
+impl RandomizedMeasurements {
+    /// Append the randomized measurements to a Quil program. The provided program
+    /// should not contain any preexisting measurements.
```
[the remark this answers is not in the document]
erichulburd: I'll make this an error.

crates/lib/src/qpu/experimental/randomized_measurements.rs
```diff
@@ -0,0 +1,628 @@
...
+        // Include the program body that was passed in.
+        program.add_instructions(target_program.into_body_instructions());
+
+        // Play the random unitaries on each qubit.
+        program.add_instructions(
+            self.unitary_set
+                .to_instructions(self.qubit_randomizations.as_slice()),
+        );
+
+        program.add_instruction(Instruction::Fence(Fence { qubits: Vec::new() }));
```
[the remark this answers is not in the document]
erichulburd: Because we want to play the 1Q rotations and measurements as two sequential and isolated cycles regardless as to what else is going on in the program.

crates/lib/src/qpu/experimental/randomized_measurements.rs
```diff
@@ -0,0 +1,628 @@
...
+                    qubit_randomization.destination_reference(0),
+                ),
+                rx_pi_on_2(qubit_randomization.measurement.qubit.clone()),
+                rz(
+                    qubit_randomization.measurement.qubit.clone(),
+                    qubit_randomization.destination_reference(1),
+                ),
+            ]);
+        }
+        instructions.push(Instruction::Fence(Fence { qubits: Vec::new() }));
```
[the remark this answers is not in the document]
erichulburd: See https://github.com/rigetti/qcs-sdk-rust/pull/515#discussion_r1881150871

crates/lib/src/qpu/experimental/randomized_measurements.rs
```diff
@@ -0,0 +1,628 @@
...
+    }
+
+    fn to_instructions(&self, qubit_randomizations: &[QubitRandomization]) -> Vec<Instruction> {
+        match self {
+            Self::Zxzxz(_) => Self::to_zxzxz_instructions(qubit_randomizations),
+        }
+    }
+
+    fn to_zxzxz_instructions(qubit_randomizations: &[QubitRandomization]) -> Vec<Instruction> {
+        let mut instructions = vec![Instruction::Fence(Fence { qubits: Vec::new() })];
```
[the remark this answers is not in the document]
erichulburd: https://github.com/rigetti/qcs-sdk-rust/pull/515#discussion_r1881150871

crates/lib/src/qpu/experimental/randomized_measurements.rs
```diff
@@ -0,0 +1,628 @@
...
+        for qubit_randomization in qubit_randomizations {
+            instructions.extend(vec![
+                rz(
+                    qubit_randomization.measurement.qubit.clone(),
+                    qubit_randomization.destination_reference(0),
```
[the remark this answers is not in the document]
erichulburd: I don't see any panic possibilities and because we validate the destination declaration now, this is guaranteed to produce a valid memory reference.

crates/python/qcs_sdk/_tracing_subscriber/common/__init__.pyi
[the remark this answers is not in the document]
erichulburd: Alternatively, we could have a separate rust script for building the stubfiles, rather than using them in the build script.

crates/lib/src/qpu/experimental/randomized_measurements.rs
```diff
@@ -0,0 +1,628 @@
...
+    }
+
+    fn destination_reference(&self, index: u64) -> MemoryReference {
+        MemoryReference::new(self.destination_declaration.name.clone(), index)
+    }
+}
+
+const RANDOMIZED_MEASUREMENT_SOURCE: &str = "randomized_measurement_source";
+const RANDOMIZED_MEASUREMENT_DESTINATION: &str = "randomized_measurement_destination";
+const RANDOMIZED_MEASUREMENT_SEED: &str = "randomized_measurement_seed";
```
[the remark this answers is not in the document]
erichulburd: I'm missing some context - did I miss a comment?

crates/lib/src/qpu/experimental/random.rs
Shadow53: It would be `qcs` instead of `crate`, but yes. Basically, write the import like it's being used in a separate crate.

crates/lib/src/qpu/experimental/randomized_measurements.rs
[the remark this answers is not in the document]
Shadow53: Possibly. In short, it's easier to find `const` definitions if they're at the top of the file instead of mixed with struct definitions

crates/lib/src/qpu/experimental/random.rs
erichulburd: I'll change this to use 0 based indexing.

crates/lib/src/qpu/experimental/random.rs
erichulburd: Yup. Pretty unlikely to change, but certainly possible.

crates/lib/src/qpu/experimental/random.rs
erichulburd: Nope - the value will be masked and the extra bits are discarded. For our purposes here, that's fine because the function is private and this will only be used to convert a PRNG value to an index just as we do on the control system.... For general multiplication it would be an issue, but that's well out of scope here.

crates/lib/src/qpu/experimental/random.rs
erichulburd: I've implemented this locally, but I actually think the simple functions are better in this case; including the trait in the signature requires that the trait is public, and there's no real reason to support public implementations at this point.

Shadow53: I'm fine leaving this as-is, especially if we aren't expecting a v2 any time soon.

For future reference, we wouldn't have to worry about public implementations if we made `LfsrVersion` a _sealed_ trait (publicly visible, not publicly implementable)

crates/python/src/qpu/experimental.rs
```diff
@@ -110,20 +110,23 @@ rigetti_pyo3::impl_repr!(PyUnitarySet);
 impl PyUnitarySet {
     #[staticmethod]
     fn from_zxzxz(inner: &PyArray2<f64>) -> PyUnitarySet {
-        Self(UnitarySet::Zxzxz(inner.to_owned_array()))
+        Self(UnitarySet::try_new_zxzxz(inner.to_owned_array()).unwrap())
```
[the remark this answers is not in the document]
erichulburd: I'll return the error; good catch.

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -205,28 +279,30 @@ fn prng_value_to_sub_region_index(value: u64, sub_region_count: u8) -> u8 {
...
+/// CALL choose_random_real_sub_regions destination source 3 seed
 /// ```
 ///
-/// with a seed of 639523, you could backout the randomly chosen sub-regions with the following:
+/// with a seed of 639,523, you could backout the randomly chosen sub-regions with the following:
 ///
 /// ```rust
-/// let seed = 639523;
+/// use qcs::qpu::experimental::random::choose_random_real_sub_regions_indices;
+///
```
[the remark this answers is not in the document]
erichulburd: misspelling on my part; I'll keep this open until CI passes.

crates/lib/src/qpu/experimental/randomized_measurements.rs
```diff
@@ -213,11 +213,14 @@ impl RandomizedMeasurements {
...
-            .map(|randomization| randomization.measurement.qubit.clone())
+            .map(|randomization| &randomization.measurement.qubit)
             .collect::<HashSet<_>>();
         if !qubits_with_redundant_measurements.is_empty() {
             return Err(Error::ProgramContainsPreexistingMeasurements(
-                qubits_with_redundant_measurements,
+                qubits_with_redundant_measurements
+                    .into_iter()
+                    .cloned()
+                    .collect(),
```
Shadow53, line 223: Oh, I missed that this was here 🤦🏻. Cloning in the initial iterator (line 216) makes sense now.

kalzoo (changes requested):
Looking good, about ready to approve. I'm as uneasy as before about merging this experimental utility into the SDK itself, but it's the best option we have at the moment because of python binding & distribution. Your take here is neat, compartmentalized, and tested.

Things I'd like to see before merge:

* A module doc as a general explainer about what's going on here. Doesn't have to specify a higher-level application but should give some context - as it is, I think almost any user might miss the point here.
* A trustworthy explainer link for LFSR for context
* A Rust e2e example under `examples` since it would require live QPU access

Thinking ahead: I'm also wondering if "program builder" or "program operator" is going to be a common pattern going forward and what that might look like. Doesn't block this PR, but I do wonder if a trait might help conceptually group tools like this in the mind of a user and help them understand how they're used:

```rust
trait ProgramOperator {
  fn apply(&self, program: Program) -> Result<Program, Self::Error> {
     ...
  }
}
```

crates/lib/src/qpu/externed_call.rs
```diff
@@ -0,0 +1,43 @@
...
+//! Trait implementations support semantically meaningful interfaces
+//! that may be converted into Quil [`quil_rs::instruction::Call`]
+//! instructions as well as their associated `PRAGMA EXTERN`
+//! instructions.
+use std::convert::TryInto;
+
+use quil_rs::quil::ToQuilError;
+
+/// A trait for supporting `PRAGMA EXTERN` and [`quil_rs::instruction::Call`] instructions.
+pub trait ExternedCall: Sized + TryInto<quil_rs::instruction::Call> {
```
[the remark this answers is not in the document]
erichulburd: Probably, but I don't think it's worth it to implement that abstraction in quil-rs yet because nothing there would use the functionality. I've removed the trait altogether and opened https://github.com/rigetti/quil-rs/issues/426.

crates/lib/src/qpu/experimental/randomized_measurements.rs
```diff
@@ -0,0 +1,685 @@
...
+                )
+            })
+            .collect()
+    }
+}
+
+/// A set of unitaries, each of which may be expressed as a set of Quil
+/// instructions.
+#[derive(Debug, Clone)]
+pub struct UnitarySet(UnitarySetInner);
```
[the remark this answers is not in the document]
erichulburd: I meant the actual `UnitarySet` instantiations are up to the user, however, we need to support the `impl` methods on `UnitarySet` in order to be able to construct the program properly.

Under this consideration, it's probably best to refactor this as a trait, which can be [implemented in Python](https://pyo3.rs/v0.23.3/trait-bounds.html)!

crates/python/qcs_sdk/qpu/experimental.pyi
erichulburd: I've used pdoc to [include](https://pdoc.dev/docs/pdoc.html#quickstart) test files within the documentation. This has the following advantages:

1. Example source code is tested.
2. Examples are only implemented once (rather than repeated b/t docs and tests).
3. Is compatible with pdoc.
4. Is compatible with PyO3 modules (I've not been able to get doctests within stubfiles for PyO3 modules to run at all).

Disadvantage:

More testing boilerplate than is ideal.

crates/lib/src/qpu/experimental/randomized_measurements.rs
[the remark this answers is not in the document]
erichulburd: This worked out pretty well (though wasn't trivial). I went ahead and monkey patched a Python ABC class into the Rust module to make the interface clear.

erichulburd:
> Looking good, about ready to approve. I'm as uneasy as before about merging this experimental utility into the SDK itself, but it's the best option we have at the moment because of python binding & distribution. Your take here is neat, compartmentalized, and tested.
> 
> Things I'd like to see before merge:
> 
> * A module doc as a general explainer about what's going on here. Doesn't have to specify a higher-level application but should give some context - as it is, I think almost any user might miss the point here.
> * A trustworthy explainer link for LFSR for context
> * A Rust e2e example under `examples` since it would require live QPU access
> 
> Thinking ahead: I'm also wondering if "program builder" or "program operator" is going to be a common pattern going forward and what that might look like. Doesn't block this PR, but I do wonder if a trait might help conceptually group tools like this in the mind of a user and help them understand how they're used:
> 
> ```rust
> trait ProgramOperator {
>   fn apply(&self, program: Program) -> Result<Program, Self::Error> {
>      ...
>   }
> }
> ```

* I've added a bit more flesh to the Rust and Python module documentation.
* The e2e integration test passed for me locally.
* I think Wikipedia is a good entrypoint for the LFSR algorithm and have linked it.

CODEOWNERS
BatmanAoD: @Shadow53 did you add Eric since this comment was made? It looks like he should already have `admin` permission.

erichulburd:
> I think Wikipedia is a good entrypoint for the LFSR algorithm and have linked it.

It's actually kind of difficult to find a concise and effective summary of LFSR that is _not_ Wikipedia. If we want to avoid linking Wikipedia,: https://pylfsr.github.io/ may be our best best.

* https://www.mathworks.com/matlabcentral/fileexchange/66871-lfsr-linear-feedback-shift-register concisely describes the algorithm graphically.
* https://www.wolframscience.com/nks/notes-7-5--random-number-generators/ provides some historical context and discussion in context of cellular automata.

crates/lib/src/qpu/experimental/randomized_measurements.rs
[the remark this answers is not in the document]
erichulburd: As per my comment below, I'm moving this to a separate PR.

erichulburd:
I've decided to separate out the random module from the randomized measurements. The latter is more feature complete, but the research thrust here does not strictly require it. If we want to extend support for that feature we can merge https://github.com/rigetti/qcs-sdk-rust/pull/536 .

CODEOWNERS
erichulburd: Looks like I've been added.

PR #516, opened 2024-12-13
Expose api execution options timeout ii

### LINKED ISSUE #490: Support the job timeout API execution option
API execution options has a `timeout` field that represents the timeout for a job to execute. The SDK should support this in its own execution options. This will improve ergonomics for Rust users, and make it possible for Python users to set the field.

erichulburd:
Closes #490

### DIFF
--- a/crates/lib/src/qpu/api.rs
+++ b/crates/lib/src/qpu/api.rs
@@ -356,6 +356,26 @@ impl Default for ExecutionOptions {
 
 impl Eq for ExecutionOptions {}
 
+fn try_std_duration_from_pbjson_types_duration(
+    duration: ::pbjson_types::Duration,
+) -> Result<Duration, QpuApiError> {
+    let seconds = u64::try_from(duration.seconds)
+        .map_err(|error| QpuApiError::StdFromProtoDurationConversion { error, duration })?;
+    let nanos = u32::try_from(duration.nanos)
+        .map_err(|error| QpuApiError::StdFromProtoDurationConversion { error, duration })?;
+    Ok(Duration::new(seconds, nanos))
+}
+
+fn try_pbjson_types_duration_from_std_duration(
+    duration: Duration,
+) -> Result<::pbjson_types::Duration, QpuApiError> {
+    let seconds = i64::try_from(duration.as_secs())
+        .map_err(|error| QpuApiError::ProtoFromStdDurationConversion { duration, error })?;
+    let nanos = i32::try_from(duration.subsec_nanos())
+        .map_err(|error| QpuApiError::ProtoFromStdDurationConversion { duration, error })?;
+    Ok(::pbjson_types::Duration { seconds, nanos })
+}
+
 /// Options available when executing a job on a QPU, particular to the execution service's API.
 /// This is a conventent alias for [`InnerApiExecutionOptions`] which provides a builder.
 ///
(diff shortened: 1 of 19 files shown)


crates/lib/src/qpu/api.rs
```diff
@@ -404,6 +421,17 @@ impl ApiExecutionOptionsBuilder {
             .bypass_settings_protection = bypass_settings_protection;
         self
     }
+
+    /// Set the `timeout` value. See [`ApiExecutionOptions::timeout`] for more information.
+    pub fn timeout(
+        &mut self,
+        timeout: Option<::pbjson_types::Duration>,
+    ) -> Result<&mut Self, QpuApiError> {
```
BatmanAoD, line 429: Looks like this is infallible; why return a `Result`?

crates/python/Cargo.toml
```diff
@@ -10,7 +10,7 @@ categories = ["api-bindings", "compilers", "science", "emulators"]
 readme = "./README.md"
 
 [features]
-libquil = ["qcs/libquil"]
+# libquil = ["qcs/libquil"]
```
BatmanAoD, line 13: ?

crates/python/src/qpu/api.rs
```diff
@@ -421,7 +422,7 @@ impl PyExecutionOptions {
     fn api_options(&self) -> Option<PyApiExecutionOptions> {
         self.as_inner()
             .api_options()
-            .map(|x| PyApiExecutionOptions(x.clone().into()))
+            .map(|x| PyApiExecutionOptions((*x).into()))
```
BatmanAoD, line 425: Was this suggested by `clippy`? It feels less explicit to me.

crates/python/src/qpu/api.rs
```diff
@@ -539,6 +547,32 @@ py_wrap_type! {
...
+#[pyclass(name = "QpuApiDuration")]
+#[derive(Debug, Clone)]
+pub struct PyQpuApiDuration {
+    inner: QpuApiDuration,
+}
+
+#[pymethods]
+impl PyQpuApiDuration {
+    #[new]
+    fn new(seconds: i64, nanos: i32) -> Self {
```
BatmanAoD, line 559: Is there no "standard" duration type that PyO3 can use?

BatmanAoD (commented):
Some quibbles, but this looks generally correct.

crates/python/src/qpu/api.rs
erichulburd: They don't support `std::time::Duration`. They support `chrono::TimeDelta` but I don't think that's worth adding a new dependency.

crates/python/src/qpu/api.rs
erichulburd: Yes, but `x` implements Copy now, so this is different.

crates/python/Cargo.toml
erichulburd: Ahh right, I did this to run tests locally on my MBP with getting libquil errors. Would be nice to support `cargo make dev-flow` without libquil.

crates/python/src/qpu/api.rs
BatmanAoD: Sad day.

crates/lib/src/qpu/api.rs
erichulburd: Leftover from a previous implementation where I had `std::time::Duration` in the signature, which was a fallible conversion. I reverted that since pyo3 doesn't support Duration.

crates/python/Cargo.toml
BatmanAoD: Since the dependency is optional, I assume it's intentional that `dev-flow` turns it on. I don't know for sure, though.

erichulburd: Oh - yup looks like `dev-test-flow` is a cargo make predefined task which tests with `all-features`. So locally I need to `cargo make pytest-flow && cargo make stubtest`.

erichulburd:
If you merge, we can squash to the single first commit "feat: expose api execution options timeout"

BatmanAoD (approved):

Issue #507, opened 2024-10-11
Correct versioning for publishing multiple packages


BatmanAoD:
I think we need something like this: https://gitlab.com/rigetti/qcs/clients/qcs-api-client-rust/-/blob/main/knope.toml?ref_type=heads.

-----

I wrote this to explain the problem, before I remembered we must have a solution in the qcs-api repo:

I think we also need to somehow split the Knope release for each package into separate steps, and I'm not sure whether Knope actually supports that. The simplest but most annoying solution might be to just prohibit making changes to both packages in a single PR. The sequence of events for a release is:

* Knope updates both `Cargo.toml` files and the `Cargo.lock` file to have new package versions.
  * We could potentially add a script to update the `python/Cargo.toml` file's dependency on the core `qcs` package at this point, but we _can't_ update the lockfile, because the updated version won't exist on crates.io yet.
* Knope pushes git commit & tags.
* Cargo publishes the `lib` crate.
* Cargo publishes the `python` crate.

What we need is:

* Knope updates the `lib` crate's `Cargo.toml` file and the lockfile.
* Cargo publishes the `lib` crate.
* A script runs that updates the `python` crate's dependency on the `lib` crate to the newly-published version.
* Knope updates the `python` crate's `Cargo.toml` file and the lockfile.
* Knope pushes git commit & tags.

jselig-rigetti:
As part of this, we should also update `knope` in the github actions to the latest version

BatmanAoD:
...also need to re-establish publishing the `python` package to `crates.io`

PR #520, opened 2024-12-19
chore: update github actions knope version to 0.7.2

jselig-rigetti:
Recently #519 failed because of https://github.com/knope-dev/action/issues/10

This fixes that by using the compatible https://github.com/knope-dev/knope/releases/tag/v0.7.2 tag

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -25,7 +25,7 @@ jobs:
         fetch-depth: 0
         token: ${{ secrets.PAT }}
     - name: Install Knope
-      uses: knope-dev/action@v2.0.0
+      uses: knope-dev/action@v2.1.0
       with:
         version: 0.18.0 # Test before updating, breaking changes likely: https://github.com/knope-dev/action#install-latest-version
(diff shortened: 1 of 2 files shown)


BatmanAoD:
Knope is on v0.18.1; any reason for keeping this so far behind?

jselig-rigetti:
> Knope is on v0.18.1; any reason for keeping this so far behind?

@BatmanAoD After heeding the **_Test before updating, breaking changes likely_** comment, I tried a `knope prerelease --dry-run` locally and the CHANGELOG it generated seemed to think that every commit in the repo history was a breaking change.

The goal for this PR is just to fix the current broken release, but you have a good point - I added a comment to a mostly related issue you opened https://github.com/rigetti/qcs-sdk-rust/issues/507#issuecomment-2555911686

BatmanAoD:
@jselig-rigetti Bleargh... we're using Knope 0.16 in other repos, and the behavior has changed a fair bit since 0.7. You're right, though, that it can be part of #507.

BatmanAoD (approved):

PR #522, opened 2024-12-27
fix: #521: exclude isa dead qubits from quilc ISA

### LINKED ISSUE #521: Exclude dead qubits from `Compiler` ISA structure
`quilc` does not permit compiling with an ISA that has qubits with no valid operations, i.e. "dead" qubits. However, there is no guarantee that a Rigetti ISA has no dead qubits. Thus, this SDK should remove such "dead" qubits when generating the data structure to send to `quilc`.

BatmanAoD:
Resolves #521


crates/lib/src/compiler/isa/mod.rs
```diff
@@ -64,8 +64,13 @@ impl TryFrom<InstructionSetArchitecture> for Compiler {
         let qubits = qubits
             .into_iter()
             .map(|(k, v)| (k.to_string(), v))
+            .filter(|(_, q)| q.has_valid_operations())
+            .collect();
+        let edges = edges
+            .into_iter()
+            .map(|(k, v)| (k.to_string(), v))
+            .filter(|(_, e)| e.has_valid_operations())
             .collect();
```
kalzoo, line 73: To get ahead of a related error - should only be those edges for which both qubits are also included

kalzoo (changes requested):
Looking good so far - please address that comment and add test coverage for it then we're 👍

PR #526, opened 2025-01-11
feat!: use gzip compression for grpc

BatmanAoD:
Note: conserv does not support this yet, so ultimately we'll need to cache a flag somewhere indicating if compression is supported, and set it to `false` if we get a "compression not supported" response. See https://grpc.github.io/grpc/core/md_doc_compression.html

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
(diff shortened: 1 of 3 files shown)


Shadow53:
Would it be possible to use [content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation) to determine whether compression is supported? (See also [MDN: Compression](https://developer.mozilla.org/en-US/docs/Web/HTTP/Compression))

PR #532, opened 2025-01-31
fix: build wheels in better tailored environments

jselig-rigetti:
Still a work in progress - adding a "build and test all the wheels" step as part of pipeline MRs seems pretty heavy, but may be necessary.


jselig-rigetti:
Verified with test pipelines which have since been removed 👍

PR #533, opened 2025-02-10
fix: test all python releases individually, use manylinux 2_34

jselig-rigetti:
https://github.com/rigetti/qcs-sdk-rust/pull/532 wheels apparently aren't building correctly in `python 3.9` this will at least test installing them in all environments.

### DIFF
--- a/.github/action-common-python-release/action.yml
+++ b/.github/action-common-python-release/action.yml
@@ -8,6 +8,8 @@ inputs:
 
   artifact-key:
     description: "Unique upload-artifact key. Example: 'macos' or 'linux-x86_64'"
+  python-version:
+    description: "Python version used for script steps and wheel"
   python-architecture:
     description: "Python architecture used for script steps"
   rust-target:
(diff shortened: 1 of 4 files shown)


jselig-rigetti:
Testing locally with this:

```sh
# on my host machine
docker run -it --platform="linux/x86_64" --entrypoint /bin/bash -v $(pwd):/code --network="host" quay.io/pypa/manylinux_2_34_x86_64:latest

# inside container
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
. "$HOME/.cargo/env"
cargo install cargo-make

cd /code/crates/python
rm -rf .venv
python3.9 -m venv .venv
source .venv/bin/activate

cargo make install-deps

maturin release -i 3.9
```

.github/action-common-python-release/action.yml
```diff
@@ -42,17 +44,17 @@ runs:
     name: Maturin - Build
     uses: messense/maturin-action@v1
     with:
-      manylinux: '2_28'
+      manylinux: '2_34'
```
jselig-rigetti, line 47: This seemed to be the fix in https://github.com/rigetti/qcs-sdk-rust/actions/runs/13248276833

Issue #531, opened 2025-01-31
Some wheels published fail to install with "BadZipFile - Bad CRC-32 for file"


jselig-rigetti:
Found during https://github.com/rigetti/pyquil/pull/1817

This may only affect some platforms and versions, but can be reproduced by:

```sh
docker run -it --platform="linux/x86_64" python:3.9 pip install qcs-sdk-python
```

Resulting in:
```
ERROR: Exception:
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/cli/base_command.py", line 160, in exc_logging_wrapper
    status = run_func(*args)
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/cli/req_command.py", line 247, in wrapper
    return func(self, options, args)
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/commands/install.py", line 507, in run
    installed = install_given_reqs(
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/req/__init__.py", line 73, in install_given_reqs
    requirement.install(
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/req/req_install.py", line 796, in install
    install_wheel(
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/operations/install/wheel.py", line 729, in install_wheel
    _install_wheel(
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/operations/install/wheel.py", line 589, in _install_wheel
    file.save()
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/operations/install/wheel.py", line 388, in save
    shutil.copyfileobj(f, dest)
  File "/usr/local/lib/python3.9/shutil.py", line 205, in copyfileobj
    buf = fsrc_read(length)
  File "/usr/local/lib/python3.9/zipfile.py", line 926, in read
    data = self._read1(n)
  File "/usr/local/lib/python3.9/zipfile.py", line 1016, in _read1
    self._update_crc(data)
  File "/usr/local/lib/python3.9/zipfile.py", line 944, in _update_crc
    raise BadZipFile("Bad CRC-32 for file %r" % self.name)
zipfile.BadZipFile: Bad CRC-32 for file 'qcs_sdk/qcs_sdk.cpython-39-x86_64-linux-gnu.so'
```

jselig-rigetti:
Starting in the repo root directory, running either of the following
```
[...]

PR #535, opened 2025-02-11
fix: wheel building for linux, macos

jselig-rigetti:
From https://github.com/rigetti/qcs-sdk-rust/issues/531#issuecomment-2649549494


.github/workflows/check-release-python.yml
Shadow53, line 1: Are these commented-out sections intended to stay?

.github/action-common-python-release/action.yml
```diff
@@ -71,3 +71,9 @@ runs:
     with:
       name: wheels_${{ inputs.package-name }}-${{ inputs.python-version }}-${{ inputs.artifact-key }}
       path: dist/
+
+  - name: unzip test
+    shell: bash
+    run: |
+      unzip -l dist/*
+      unzip -T dist/*
```
Shadow53, line 79: Aren't we already testing wheel installation? I figure that would cover testing the zip file as well.

jselig-rigetti: I'd think so. But it's still failing 😞

.github/workflows/check-release-python.yml
jselig-rigetti: I'm not sure yet - trying things to figure out what works.

.github/action-common-python-release/action.yml
jselig-rigetti: removed, you are right.

.github/workflows/check-release-python.yml
jselig-rigetti: removed this file.

PR #536, opened 2025-02-18
Feat randomized measurements

erichulburd:
A continuation of #515  that explicitly supports randomized measurements.

This PR supports one new feature:

* A `RandomizedMeasurement` struct / class that will add per-shot randomized measurements to a program, as well as support the client in ascertaining which random unitary was played on each qubit per shot.

The documentation should stand on its own here. One note, I've put all of this functionality into `qpu/experimental` (both in Rust and Python). It is experimental both in the sense that it is an experimental feature in the QCS SDK that should not be considered stable as well as in the sense that it supports features relevant to QIS experimentalists.

Review Guidance:

1. `crates/lib/src/qpu/experimental/randomized_measurements.rs` measurements builds off the primitives defined in the file above.
2. Stubs for these features are defined in `crates/python/qcs_sdk/qpu/experimental/randomized_measurements.pyi`.
3. There are corresponding tests for randomized measurements in `crates/lib/src/qpu/experimental/randomized_measurements.rs` as well as `crates/lib/examples/randomized_measurements.rs`.


erichulburd:
We can reopen this at some point in the future if a partner requests the feature.

PR #540, opened 2025-03-10
fix(python): don't use compression when storing github actions artifacts

jselig-rigetti:
Update some actions versions to (hopefully) reduce warning spam, and set upload-artifacts compression level to 0


jselig-rigetti:
Related to https://github.com/rigetti/qcs-sdk-rust/issues/531

PR #552, opened 2025-08-05
chore: upgrade 'warp' (dev-dependency)

BatmanAoD:
I got an email notification that `warp` has upgraded to `http=1`, so this upgrade fully removes `http=v0.2.12` from our `dev` dependencies. (It was already purged from non-dev dependencies.)


BatmanAoD:
This will be done as part of https://github.com/rigetti/qcs-sdk-rust/pull/587

PR #558, opened 2025-10-15
fix: update to the latest quil-rs release; should fix Rust crate release

BatmanAoD:
Relates to #507, but doesn't fix it


BatmanAoD:
Trying here, though I'm not sure if this repo is actually set up to be able to publish from a branch: https://github.com/rigetti/qcs-sdk-rust/actions/runs/18538213339

PR #560, opened 2025-10-16
fix: put pyo3 behind feature flag for qcs crate

gsmith-rigetti:
The `pyo3` dependency for the `qcs` crate breaks one of our internal Docker builds and really isn't necessary by default, so I've put it behind a `"pyo3"` feature flag.


BatmanAoD (approved):
Approved, modulo Alexander's suggestion to change the feature name

PR #565, opened 2025-10-31
fix: update QCS API-client versions

BatmanAoD:


jselig-rigetti (commented):
Are all the errors in the "run checks" new?

jselig-rigetti (approved):

PR #588, opened 2026-05-18
feat: use QCS "core dependencies" crate

BatmanAoD:
relies on https://github.com/rigetti/rigetti-pyo3/pull/77 and https://github.com/rigetti/pyo3-opentelemetry/pull/35


.devcontainer/devcontainer-lock.json
BatmanAoD, line 1: I set up a devcontainer so that I could fix build failures locally, even with the `quilc` feature.

crates/lib/Cargo.toml
BatmanAoD, line 1: Many of the changes here are from running `cargo autoinherit`.

scripts/install-libquil
BatmanAoD, line 1: This is basically just a codeblock from the `checks` workflow, extracted into a script so that it can be called from the devcontainer setup.

.github/workflows/checks.yml
```diff
@@ -22,16 +22,9 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           version: '3.20.1'
-      - name: Install libquil dependencies
+      - name: Install quilc
```
jselig-rigetti: Was libquil previously failing, or is it not relevant to test it specifically?

scripts/install-libquil
jselig-rigetti: Should this be called `install-libquil` ? Seems specific to that and not the server package

deny.toml
```diff
@@ -177,19 +177,30 @@ allow = [
 ]
 # List of crates to deny
 deny = [
-    # Multiple versions of these crates (and possibly others?) are known to cause
-    # compilation and subtle unit test failures. We deny multiple versions to make
-    # the cause of those failures more explicit. If the deny check fails, perhaps,
-    # because a dependency depends on one of these crates that do not affect the
-    # APIs used in this workspace, then we can disable the deny check if, otherwise,
-    # there's no trivial solution.
+    { crate = "qcs-dependencies-client",      deny-multiple-versions = true },
```
jselig-rigetti:
```suggestion
    { crate = "qcs-dependencies-client", deny-multiple-versions = true },
```

jselig-rigetti (approved):
Looks good 👍

.github/workflows/checks.yml
BatmanAoD: I just misunderstood what the deleted steps were doing, since the actual libquil bit looks like it's running a quilc install script. I've restored the original name and changed the script name.

scripts/install-libquil
BatmanAoD: yep, done but not pushed, sorry

.github/workflows/checks.yml
BatmanAoD: ...but also, it definitely appears that there are combinations of features we're not testing somehow. 😬

crates/lib/tests/mocked_qpu.rs
```diff
@@ -344,12 +345,14 @@ mod qpu {
...
-    use tonic::{codec::CompressionEncoding, transport::Server, Request, Response, Status};
+    use qcs_dependencies_client::tonic::{
+        codec::CompressionEncoding, transport::Server, Request, Response, Status,
+    };
 
     #[derive(Default, Debug)]
     pub struct ControllerService {}
 
-    #[tonic::async_trait]
+    #[qcs_dependencies_client::tonic::codegen::async_trait]
```
Shadow53: `async-trait` is a direct dependency. Is there a reason for using a hidden re-export?

crates/lib/Makefile.toml
```diff
@@ -37,11 +37,15 @@ install_crate = "cargo-deny"
...
+command = "cargo"
+args = ["--all-features --all-targets"]
+
 [tasks.pre-ci-flow]
-dependencies = ["deny", "lint"]
+dependencies = ["deny", "lint", "check"]
 
 [tasks.ci-flow]
-dependencies = ["test-flow"]
+dependencies = ["pre-ci-flow", "test-flow"]
```
Shadow53: Seems like we should consider using the [default ci-flow task](https://github.com/sagiegurari/cargo-make/blob/95dcc545db8cf08af6fbec524e200e7c80b06027/src/lib/descriptor/makefiles/rust.toml#L2)

Even if we don't, these should all be put into `ci-flow`, as the `pre-TASK` and `post-TASK` patterns are for setup and teardown of things like supporting Docker containers.

.gitignore
```diff
@@ -1,6 +1,9 @@
 # Cargo / rustc artifacts
 /target
 
+# QCS infrastructure (fetched at build time)
+.qcs-infrastructure/
+
```
Shadow53: I don't see anything that creates this directory

BatmanAoD: Oh - yeah, this will probably never apply to public repos. Will remove.

crates/lib/tests/mocked_qpu.rs
BatmanAoD: I don't know why we're using the re-export here, but the only code changes I made were to update existing paths using `tonic` and other "core" dependencies. I can switch to raw `async_trait` here.

crates/lib/Makefile.toml
BatmanAoD: I haven't previously looked into the defaults since we have our own somewhat-standardized setup. That looks...more complex than I'd like. I'll get rid of `pre-ci-flow`, though.

crates/lib/scripts/check-py-api.sh
```diff
@@ -7,9 +7,20 @@
...
+tmpdir=$(mktemp -d)
+cleanup() {
+  popd
+  rm -rf "$tmpdir"
+  # The `griffe` check command creates a branch (and worktree), but does not clean them up.
+  git worktree list --porcelain \
+    | awk '/^worktree /{wt=$2} /^branch refs\/heads\/griffe-lib-v/{print wt}' \
+    | xargs -r -I{} git worktree remove --force {}
+  git show-branch --list 'griffe-lib-v*' 2>/dev/null | awk -F'[][]' '{print $2}' | xargs -r git branch -D
+}
```
BatmanAoD, line 22: This isn't necessary in CI but made local iteration on this script way easier.

Cargo.toml
```diff
@@ -25,14 +24,48 @@ tracing = { version = "0.1", features = ["log"] }
...
-pyo3-opentelemetry = "0.6.0"
-pyo3-tracing-subscriber = { version = "0.4.0", default-features = false }
-qcs-api-client-common = "0.17.0"
-qcs-api-client-grpc = "0.17.0"
-qcs-api-client-openapi = "0.18.0"
+optipy = "0.1"
+# XXX use release
+pyo3-opentelemetry = "0.7.0-rc.2"
+# XXX use release
+pyo3-tracing-subscriber = "0.5.0-rc.2"
```
BatmanAoD: Oops, forgot CI doesn't use `relint` in public repos. Will update to the release.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends cmake pkg-config sudo && \
+    /tmp/install-libquil && \
+    rm -rf /var/lib/apt/lists/* /tmp/install-libquil
+
+RUN cargo install cargo-binstall
+RUN cargo binstall --no-confirm cargo-make cargo-deny ripgrep knope
+
+# The user is assumed to own the repo, even though the container is run with a different user.
+RUN git config --global --add safe.directory /workspaces/qcs-sdk-rust
```
Shadow53, line 15: TIL

crates/lib/python/qcs_sdk/qpu/experimental/random.pyi
```diff
@@ -11,10 +11,7 @@ class ChooseRandomRealSubRegions:
     An [`ExternedCall`] that may be used to select one or more random
     sub-regions from a source array of real values to a destination array.
     """
-    NAME: builtins.str = 'choose_random_real_sub_regions'
-    r"""
-    The name of the function referenced by the `PRAGMA EXTERN` and `CALL` instructions.
-    """
```
Shadow53, line 17: Why were these removed?

crates/lib/src/qpu/experimental/random.rs
```diff
@@ -161,14 +161,21 @@ impl ChooseRandomRealSubRegions {
...
+#[cfg(feature = "python")]
 #[cfg_attr(not(feature = "stubs"), optipy::strip_pyo3(only_stubs))]
 #[cfg_attr(feature = "stubs", gen_stub_pymethods)]
-#[cfg_attr(feature = "python", pyo3::pymethods)]
+#[pyo3::pymethods]
 impl ChooseRandomRealSubRegions {
     #[classattr]
     #[pyo3(name = "NAME")]
-    /// The name of the function referenced by the `PRAGMA EXTERN` and `CALL` instructions.
-    pub const EXTERN_NAME: &str = "choose_random_real_sub_regions";
```
Shadow53, line 171: Why the change from `const` to `fn`?

crates/lib/src/client.rs
```diff
@@ -38,25 +32,13 @@ const MAX_TRANSLATION_OUTBOUND_REQUEST_SIZE: usize = 50 * 1024 * 1024;
 /// It is public so that users can create gRPC clients with different APIs using a "raw" connection
 /// initialized by this library. This ensures that the exact Tonic version used for such clients
 /// matches what this library uses.
-#[cfg(not(any(feature = "grpc-web", feature = "tracing")))]
-pub type GrpcConnection = RetryService<RefreshService<Channel, ClientConfiguration>>;
```
Shadow53, line 42: Was this alias deleted entirely?

crates/lib/README.md
```diff
@@ -14,15 +14,15 @@ This crate is documented primarily via [rustdoc] comments and examples, which ar
...
-| `python`                  | Exposes Python bindings via PyO3.               | `experimental`, `tracing-opentelemtry` |
-| `stubs`                   | Enables Python typing for stub generation.                            | `python`         |
+| Feature                   | Description                                                                  | Implied Features                       |
+| ------------------------- | ---------------------------------------------------------------------        | ----------------                       |
+| `manual-tests`            | Used to run tests that require some manual setup.                            |                                        |
+| `experimental`            | Enables experimental features specific to Rigetti QPUs.                      |                                        |
+| `libquil`                 | Links against the `libquil` dynamic library. Not supported on all platforms. |                                        |
+| `tracing`                 | Enables the `tracing` features in `qcs-api-client-*` crates.                 |                                        |
```
Shadow53: Wasn't this feature removed?

crates/lib/python/qcs_sdk/qpu/experimental/random.pyi
BatmanAoD: That seems to be a result of the pyo3-otel update. I'll take another look.

crates/lib/src/client.rs
BatmanAoD: Yes, this was only used when `tracing` is not enabled.

BatmanAoD: ...unless you mean, is there an alias under a different feature combination, in which case, yes, it's now unconditionally aliased on line 42, with tracing support and without grpc-web support.

crates/lib/README.md
BatmanAoD: Oops, yes, I thought I sync'd the README with the Cargo.toml, but I guess not.

crates/lib/src/client.rs
Shadow53: Yeah, the question was whether `GrpcConnection` was still around in some form.

crates/lib/src/qpu/experimental/random.rs
BatmanAoD: Compile failure when using `experimental` but not `python`; fixed now

crates/lib/python/qcs_sdk/qpu/experimental/random.pyi
BatmanAoD: It was actually due to the fix for the compile-failure in `random.rs`. Resolved now.

crates/lib/Cargo.toml
```diff
@@ -119,30 +106,30 @@ rigetti-pyo3 = { workspace = true, optional = true, features = ["opentelemetry",
...
-hex = "0.4.3"
-insta = "1.29.0"
-maplit = "1.0.2"
+assert2 = { workspace = true }
+erased-serde = { workspace = true }
+float-cmp = { workspace = true }
+hex = { workspace = true }
+insta = { workspace = true }
+maplit = { workspace = true }
 # keep pinned until this crate matures
```
jselig-rigetti: I guess this comment is no longer relevant in this spot 😅

scripts/install-libquil
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+if [[ $# -ne 0 ]]; then
+    echo "Usage: $0"
+    echo "Installs quilc, the Quil compiler."
+    exit 0
+fi
```
jselig-rigetti, line 7: Is this to reserve future potential positional arguments or something?

Shadow53 (approved):

scripts/install-libquil
Shadow53: Pretty common to make sure scripts error when unknown arguments are passed instead of just ignoring them

crates/lib/Cargo.toml
BatmanAoD: Oops, good catch! I used `autoinherit` for this, which of course doesn't move comments.

scripts/install-libquil
BatmanAoD: I do this in most scripts that don't do any other arg-parsing. It's mostly just so that if anyone tries `-h`, `--help`, `-?`, or anything else, they'll get a helpstring rather than jumping right into `sudo` commands. Secondarily, it's so that in case someone does try to use any command-line arguments or flags, they're not silently ignored.

PR #592, opened 2026-07-24
feat!: update dependencies (including PyO3)

### LINKED ISSUE #590: We should update `rigetti-pyo3` and related dependencies
We've updated `pyo3`, `pyo3_stub_gen`, and other related dependencies in `rigetti-pyo3`. We should update this repository to take advantage of those changes.

BatmanAoD:
Note: updating PyO3 is a build-breaking change

Resolves #590


jselig-rigetti (dismissed):

.github/workflows/checks.yml
```diff
@@ -38,10 +39,23 @@ jobs:
...
+        uses: astral-sh/setup-uv@v6
         with:
-          python-version: '3.13'
-      - uses: abatilo/actions-poetry@v4
+          enable-cache: true
+          cache-dependency-glob: crates/lib/uv.lock
+
+      # Rust builds of this crate link against libpython (the `stubs`/`python` features), so
+      # the interpreter and its `LIBDIR` have to be on the environment for every later step —
+      # not just the Python ones. `ensure-venv` exports both; publish them job-wide.
```
Shadow53: Nit

```suggestion
      # not just the Python ones. `ensure-venv` exports both; publish them pipeline-wide.
```

Or

```suggestion
      # not just the Python ones. `ensure-venv` exports both; publish them to the entire pipeline.
```

Shadow53: Alternatively, could `ensure-venv` check for the existence of `GITHUB_ENV`/`GITHUB_PATH` and write to it directly?

.github/workflows/checks.yml
```diff
@@ -38,10 +39,23 @@ jobs:
...
+      - name: Set up Python virtual environment
+        shell: bash
+        run: |
+          . scripts/ensure-venv
+          echo "VIRTUAL_ENV=${VIRTUAL_ENV}" >> "$GITHUB_ENV"
+          echo "VENV_PATH=${VENV_PATH}" >> "$GITHUB_ENV"
+          echo "LD_LIBRARY_PATH=${LD_LIBRARY_PATH}" >> "$GITHUB_ENV"
+          echo "${VIRTUAL_ENV}/bin" >> "$GITHUB_PATH"
```
Shadow53: Does this prepend or append?

.github/workflows/checks.yml
```diff
@@ -18,14 +18,15 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Install protoc
-        uses: arduino/setup-protoc@v1
+        uses: arduino/setup-protoc@v3
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-          version: '3.20.1'
+          # v3 matches against protobuf's post-3.20 release names, so '3.20.1' no longer
+          # resolves. 21.0 is its immediate successor, and matches release-python.yml.
```
Shadow53: This seems more appropriate as a PR comment rather than a source code comment. Applies to other instances of this comment as well.

.github/workflows/checks.yml
Shadow53, line 1: Common things like installing `protoc` and `uv`, and setting up the venv, may benefit from using [reusable workflows](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows)

.github/workflows/checks.yml
```diff
@@ -80,11 +95,28 @@ jobs:
...
+          echo "${VIRTUAL_ENV}/bin" >> "$GITHUB_PATH"
 
-      - uses: abatilo/actions-poetry@v4
+      # Also a dependency of `test-python-package`, but run first so that stale generated
+      # files are reported before the longer test run, and as a named step of their own.
+      - name: Check Generated Python Files
+        run: cargo make --cwd crates/lib --makefile Makefile.toml check-generated-python-files
 
       - name: Test Python Package
         run: cargo make --cwd crates/lib --makefile Makefile.toml test-python-package
```
Shadow53: Since the "Check Generated Python Files" step ran previously, you can skip it here using `CARGO_MAKE_WORKSPACE_SKIP_MEMBERS` (I think) or `--skip-tasks`

crates/lib/Makefile.toml
```diff
@@ -47,40 +59,79 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
+if [ "$(readlink -f "$VIRTUAL_ENV")" != "$(readlink -f "$VENV_PATH")" ]; then
+    echo "Error: virtual env path mismatch: VIRTUAL_ENV=${VIRTUAL_ENV:-<unset>} vs VENV_PATH=${VENV_PATH}" >&2
+    exit 1
+fi
+uv sync --active --group dev
+'''
 
-# NOTE: if using `uv`, you may need to set `PYTHONHOME` to the root of the virtual environment:
-#
-#   export PYTHONHOME=$(python3 -c "import sys; print(sys.base_prefix)")
```
Shadow53, line 56: Is this no longer relevant, then?

crates/lib/Makefile.toml
```diff
@@ -47,40 +59,79 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
 [tasks.generate-stubs]
+# `stub_gen` is a binary rather than an `extension-module` cdylib, so it links against
+# libpython and needs to find it at load time. Sourcing `ensure-venv` here (rather than only
+# setting `VIRTUAL_ENV`) is what puts the interpreter's `LIBDIR` on `LD_LIBRARY_PATH`; that
+# export does not survive from the `install-python-deps` task, since each task is its own
+# process.
```
Shadow53: IIRC, `cargo make` keeps environment variables set by previous tasks. I know it does for variables set in the TOML, e.g. `[tasks.foo.env]`. Not as sure about those set in the script itself, but I still think it does.

crates/lib/Makefile.toml
```diff
@@ -47,40 +59,79 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
+    echo "Run 'cargo make --cwd crates/lib generate-stubs' and commit the result:" >&2
+    git status --porcelain . >&2
+    git --no-pager diff . >&2
+    exit 1
+fi
+echo "Generated Python files are up to date."
+'''
 
 [tasks.package-python]
+env = { VIRTUAL_ENV = "${VENV_PATH}" }
```
Shadow53: What happens if `VENV_PATH` is not set?

crates/lib/Makefile.toml
```diff
@@ -47,40 +59,79 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
-args = ["run", "maturin", "develop"]
+dependencies = ["install-python-deps", "generate-stubs"]
+command = "uv"
+args = ["run", "--active", "maturin", "develop", "--uv"]
 
 [tasks.install-python-package-with-stubs]
-description = "Install the Python package with up-to-date stubs (not needed for testing)."
-dependencies = ["generate-stubs", "install-python-package"]
+description = "Deprecated alias of `install-python-package`, which now always generates stubs."
+dependencies = ["install-python-package"]
```
Shadow53: Does anything use this that we can't change to use `install-python-package` instead?

crates/lib/Makefile.toml
```diff
@@ -91,12 +142,13 @@ dependencies = [
...
 [tasks.stubtest]
+env = { VIRTUAL_ENV = "${VENV_PATH}" }
 description = "Check Python stubfiles."
 dependencies = ["install-python-package-with-stubs"]
-command = "poetry"
+command = "uv"
 cwd = "python"
 args = [
-    "run",
+    "run", "--active",
```
Shadow53: For consistency

```suggestion
    "run",
    "--active",
```

crates/lib/Makefile.toml
```diff
@@ -141,44 +196,53 @@ dependencies = ["test", "test-python-package"]
...
 
 [tasks.document-python-package]
+env = { VIRTUAL_ENV = "${VENV_PATH}" }
 description = "Generate Python package documentation."
 dependencies = ["install-python-package-with-stubs"]
-command = "poetry"
+command = "uv"
 args = [
-    "run",
+    "run", "--active",
```
Shadow53:
```suggestion
    "run",
    "--active",
```

scripts/ensure-venv
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+
+# Source this file to ensure that the current shell is using a Python virtual environment.
+# If not sourced, the script will still create a virtual env.
+#
+# VENDORED from `qcs-infrastructure` (scripts/cli-tools-base-image/ensure-venv). This repo is
+# public and cannot reference that project, so changes made there should be copied here.
```
Shadow53: Should we make a not in the internal project as a reminder?

.github/workflows/checks.yml
BatmanAoD: Yeah, looks like that would be the `github-ci` version of `qcs-infrastructure/gitlab-ci`; nice!

.github/workflows/checks.yml
BatmanAoD: yeah I need to find a way to get Claude to do less of this kind of in-line comment

.github/workflows/checks.yml
BatmanAoD: https://gitlab.com/rigetti/qcs/utilities/qcs-infrastructure/-/work_items/48

crates/lib/Makefile.toml
BatmanAoD: It's set in the global `[env]` block, so that's not possible.

.github/workflows/checks.yml
BatmanAoD: `GITHUB_ENV` and `GITHUB_PATH` only apply between steps in a job, not between jobs, so the original comment is correct. I guess `ensure-venv` *could* do that, though it seems a little like feature creep: nothing else in that script has any logic specific to CI. (It doesn't even use any CI vars or refer to GitLab.)

In this case, though, it seems like the script could be sourced in the same `run` block that actually does the `cargo make`.

.github/workflows/checks.yml
BatmanAoD: I don't know: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-of-adding-a-system-path

In any case, I've deleted this

.github/workflows/checks.yml
BatmanAoD: Oh, since `check-generated-python-files` is a dependency, there's honestly no reason to even run it separately.

crates/lib/src/python/mod.rs
```diff
@@ -78,17 +93,25 @@ fn init_module(m: &Bound<'_, PyModule>) -> PyResult<()> {
     }
 
     let py = m.py();
-    init_submodule("qcs_sdk", py, m)?;
+    init_submodule("qcs_sdk._qcs_sdk", py, m)?;
 
     m.add("__version__", env!("CARGO_PKG_VERSION"))?;
+    m.setattr("__doc__", PACKAGE_DOC)?;
```
BatmanAoD, line 99: TBD: do this, or just return to having a handwritten `__init__.py`?

crates/lib/src/bin/stub_gen.rs
```diff
@@ -16,16 +16,62 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    ///
+    /// `pyo3_stub_gen` writes the `from ... import ...` form only for modules under our own
+    /// package, and a plain `import a.b.c` for every other module — but it refers to types in
+    /// those modules by the last component alone, as in `program.Program`. A dotted `import`
+    /// binds only the leftmost name, so those references are undefined. This affects any type
+    /// we borrow from `quil` or `qcs_api_client_common`.
+    ///
+    /// Only imports whose fully-qualified path is never used are rewritten, which leaves
+    /// correct cases such as `import collections.abc` (used as `collections.abc.Sequence`)
+    /// alone.
```
BatmanAoD: I want to know before merging whether this actually makes sense

crates/lib/Makefile.toml
BatmanAoD: Just checked, and no, it does not

crates/lib/Makefile.toml
```diff
@@ -1,6 +1,18 @@
 [env]
 RUST_BACKTRACE = 0
 CARGO_MAKE_CARGO_BUILD_TEST_FLAGS = "--features libquil,_insecure-issuer-validation,stubs"  # Disable --all-features to avoid manual tests in CI for now
+# Python tooling follows the `uv` conventions shared across QCS repos. This project is public
+# and so cannot extend `qcs-infrastructure`'s `pytasks.toml`; `scripts/ensure-venv` is vendored
+# from that project instead, and the tasks below mirror its recipes.
```
asaites: It's probably better to leave this comment out.

scripts/ensure-venv
BatmanAoD: I'll just make a TODO comment referring to https://gitlab.com/rigetti/qcs/utilities/qcs-infrastructure/-/work_items/48

crates/lib/Makefile.toml
```diff
@@ -1,6 +1,18 @@
...
+PYTHON_VERSION = { value = "3.13", condition = { env_not_set = ["PYTHON_VERSION"] } }
+# One virtual environment for the whole workspace, so every task shares it.
+#
+# `VIRTUAL_ENV` is deliberately NOT set here: PyO3's build script reads it, so a global value
+# would break every `cargo` build that happens before the venv exists — including the pure-Rust
+# `ci-flow`, which compiles with the `stubs` feature. It is set per-task instead (which is what
+# lets `uv run --active` find the venv), on tasks that first ensure the venv exists.
+VENV_PATH = { value = "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/.venv", condition = { env_not_set = ["VENV_PATH"] } }
```
asaites: I think it's actually better _not_ to share the `.venv` among workspace members. In that case, I suggest using one of these instead:

- `${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/qcs-sdk-rust/.venv`
- `${CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY}/.venv`

That said, I don't think there's any need to set it explicitly, since it is already what `uv` will default to.

crates/lib/Makefile.toml
```diff
@@ -1,6 +1,17 @@
...
+# Python tooling follows the `uv` conventions shared across QCS repos. This project is public
+# and so cannot extend `qcs-infrastructure`'s `pytasks.toml`; `scripts/ensure-venv` is vendored
+# from that project instead, and the tasks below mirror its recipes.
+PYTHON_VERSION = { value = "3.13", condition = { env_not_set = ["PYTHON_VERSION"] } }
+# One virtual environment for the whole workspace, so every task shares it.
+#
+# `VIRTUAL_ENV` is deliberately NOT set here: PyO3's build script reads it, so a global value
+# would break every `cargo` build that happens before the venv exists — including the pure-Rust
+# `ci-flow`, which compiles with the `stubs` feature. It is set per-task instead (which is what
+# lets `uv run --active` find the venv), on tasks that first ensure the venv exists.
```
asaites: This comment is seemingly irrelevant, and doesn't seem particularly accurate. I think it should just be removed.

If the `VIRTUAL_ENV` variable isn't set, PyO3 [looks for the system Python](https://pyo3.rs/v0.29.2/building-and-distribution.html#configuring-the-python-version) instead. The real issue is that it wouldn't make sense to set unless you actually know it's the relevant environment, and even then, I _don't_ think we should set here. Regardless, if the `python` feature is enabled (the `stubs` feature implies it), then the PyO3 build script runs and looks for an interpreter from which it can gather implementation-specific configuration. If the feature is not enabled, that doesn't happen.

crates/lib/Makefile.toml
```diff
@@ -47,40 +58,79 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
+script_runner = "bash"
+script = '''
+. "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/scripts/ensure-venv"
+# Sanity-check, so that a stale venv elsewhere in the tree can't be picked up silently.
+if [ "$(readlink -f "$VIRTUAL_ENV")" != "$(readlink -f "$VENV_PATH")" ]; then
+    echo "Error: virtual env path mismatch: VIRTUAL_ENV=${VIRTUAL_ENV:-<unset>} vs VENV_PATH=${VENV_PATH}" >&2
+    exit 1
+fi
+uv sync --active --group dev
+'''
```
asaites: In some other repos we have that use `uv`, we just use `uv run`, for example:

```toml
[tasks.stubtest]
dependencies = ["install-python-package"]
description = "Validate type stubs using stubtest"
cwd = "python"
command = "uv"
args = [
    "run", "--group", "dev",
    "stubtest",
    "--allowlist", "stubtest-allowlist",
    "--ignore-disjoint-bases",
    "qcs_api_client_common",
]
```

If `uv` _is_ installed in a virtual env, it'll print a warning that the environment used by `uv run` is the project's environment (`.venv/`), but that can be safely ignored. Using `--active` potentially bypasses the local `.venv` 

We _probably_ should just let `uv` be managed by the host, as we would with other tools (so we don't really care if `uv` is a binary installed via homebrew, a module in a virtual environment, an alias to `pipx`, etc.). That said, in at least one other project, we explicitly require that we're running in a virtual environment and install `uv` into it:

```toml
[tasks.check-venv]
description = "Check if a virtual environment is activated"
script = [
    '''
    if [ -z "$VIRTUAL_ENV" ]; then
        echo "No virtual environment activated. Please activate one."
        exit 1
    else
        echo "Virtual environment is active."
    fi
    '''
]

[tasks.upgrade-pip]
dependencies = ["check-venv"]
description = "Upgrade pip to ensure it supports depency groups"
command = "pip"
args = ["install", "--upgrade", "pip"]

[tasks.install-uv]
dependencies = ["check-venv", "upgrade-pip"]
description = "Install dependencies using uv"
command = "pip"
args = ["install", "uv"]
```

Realistically, I think the virtual environment check only makes sense if we're going to install/upgrade `pip` and `uv` ourselves from the `Makefile.toml`, and I am inclined to say we shouldn't (or at least not by default); instead, we should just let `uv` do its thing and call it a day. In CI, we should ensure the right tools are installed, and for developers, we should tell them what tools they need (though I'm not opposed to a convenience task to handle that install -- which we could then call during CI a setup phase).

crates/lib/Makefile.toml
```diff
@@ -47,40 +58,79 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
 
-# NOTE: if using `uv`, you may need to set `PYTHONHOME` to the root of the virtual environment:
-#
-#   export PYTHONHOME=$(python3 -c "import sys; print(sys.base_prefix)")
 [tasks.generate-stubs]
+# `stub_gen` is a binary rather than an `extension-module` cdylib, so it links against
+# libpython and needs to find it at load time. Sourcing `ensure-venv` here (rather than only
+# setting `VIRTUAL_ENV`) is what puts the interpreter's `LIBDIR` on `LD_LIBRARY_PATH`; that
+# export does not survive from the `install-python-deps` task, since each task is its own
+# process.
```
asaites: Again, this comment is largely irrelevant. The hypothetical issue at hand is that building against one interpreter and then running linked to another is just asking for trouble.

crates/lib/Makefile.toml
```diff
@@ -47,40 +58,79 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
+dependencies = ["install-python-deps"]
 description = "Generate Python stub files."
-command = "cargo"
-args = ["run", "-p", "qcs", "--features", "stubs", "--bin", "stub_gen"]
-install_crate = false
+script_runner = "bash"
+script = '''
+. "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/scripts/ensure-venv"
+cargo run -p qcs --features stubs --bin stub_gen
+'''
```
asaites: In context of my other comments, we should just run using `uv run`:


```suggestion
command = "uv"
args = ["run", "cargo", "run", "-p", "qcs-api-client-common", "--features", "stubs", "--bin", "stub_gen"]
install_crate = false
```

crates/lib/Makefile.toml
```diff
@@ -1,6 +1,17 @@
 [env]
 RUST_BACKTRACE = 0
 CARGO_MAKE_CARGO_BUILD_TEST_FLAGS = "--features libquil,_insecure-issuer-validation,stubs"  # Disable --all-features to avoid manual tests in CI for now
+# Python tooling follows the `uv` conventions shared across QCS repos. This project is public
+# and so cannot extend `qcs-infrastructure`'s `pytasks.toml`; `scripts/ensure-venv` is vendored
+# from that project instead, and the tasks below mirror its recipes.
+PYTHON_VERSION = { value = "3.13", condition = { env_not_set = ["PYTHON_VERSION"] } }
```
asaites: I don't see a good reason to set this manually, particularly if it might be wrong.

crates/lib/Makefile.toml
```diff
@@ -47,40 +58,79 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
+'''
 
 [tasks.package-python]
+env = { VIRTUAL_ENV = "${VENV_PATH}" }
 description = "Build a release version of the Python package to the active virtual environment."
 dependencies = ["install-python-deps", "generate-stubs"]
-command = "poetry"
-args = ["run", "maturin", "build", "--release"]
+command = "uv"
+args = ["run", "--active", "maturin", "build", "--release"]
```
asaites: I don't think this will have the intended effect; in particular, I'm not confident just setting the `VIRTUAL_ENV` is impact all the things that might be trying to use Python -- the `activate` scripts within a virtual environment certainly do more, including `PATH` manipulation. Even if it does work, if we accidentally remove or break the `env` line above, this can trash the user's current virtualenv. Finally, just using `uv run` will default to creating/using the project `.venv`, so I really think that's the better solution.

crates/lib/Makefile.toml
```diff
@@ -47,40 +58,79 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
+# `generate-stubs` also writes the package's `__init__.py` files, which are what expose the
+# compiled `qcs_sdk._qcs_sdk` modules under their public `qcs_sdk` names — so it is a build
+# input now, not just a convenience for type checkers.
 description = "Build the Python package and install to the active virtual environment."
-dependencies = ["install-python-deps"]
-command = "poetry"
-args = ["run", "maturin", "develop"]
+dependencies = ["install-python-deps", "generate-stubs"]
+command = "uv"
+args = ["run", "--active", "maturin", "develop", "--uv"]
```
asaites: Given that it's already running from `uv`, do we really need this flag?

crates/lib/Makefile.toml
asaites: Or can't we just delete it?

crates/lib/Makefile.toml
```diff
@@ -141,44 +193,53 @@ dependencies = ["test", "test-python-package"]
...
 [tasks.format]
+env = { VIRTUAL_ENV = "${VENV_PATH}" }
+# Only hand-written Python. Everything under `python/qcs_sdk` is generated — reformatting it
+# makes `check-generated-python-files` fail until the next `generate-stubs` run.
+# (`tests` is the Rust integration-test directory and contains no Python.)
 dependencies = ["install-python-deps"]
-command = "poetry"
-args = ["run", "black", "python/qcs_sdk", "tests"]
+command = "uv"
+args = ["run", "--active", "black", "tests_py", "scripts"]
```
asaites: An aside to this PR, we should probably use `ruff` for this instead.

asaites (commented):
I've only reviewed the `Makefile.toml`, but I had several thoughts, so I reckoned I'd share them immediately. I'll have a look at the rest of the PR soon.

crates/lib/Makefile.toml
BatmanAoD: There's only one pyproject in this crate, but for Magneto and `qcs-api-client-rust`, I thought it made sense to share the `.venv` because, within a single project, dependencies should *usually* be compatible. Why is it better not to share?

crates/lib/Makefile.toml
BatmanAoD: The comment is talking about what happens when `VIRTUAL_ENV` is set to a *nonexistent* venv path, and it's correct that building with the `stubs` feature breaks in that case. But since this specific project can just use `uv`'s default `.venv` path, I'm deleting the comment and the variable anyway.

crates/lib/Makefile.toml
BatmanAoD: Resolving because I've removed this and `ensure-venv` from this MR, but we should still discuss it since `qcs-infrastructure` tries to set up shared venvs.

crates/lib/Makefile.toml
BatmanAoD: The issue it's describing isn't just about *which* interpreter is linked, it's that sometimes libpython isn't available *at all*. I know I've run into that previously, but it doesn't seem to be an issue here, so I've deleted `ensure-venv`.

crates/lib/Makefile.toml
BatmanAoD: This was also to support `ensure-venv`, and in generally I would expect we *do* want an explicit Python version, but I'll go ahead and delete it from here.

crates/lib/Makefile.toml
BatmanAoD: Removed from here, but again, this is something that I did in `qcs-infrastructure` that we need to discuss there

crates/lib/Makefile.toml
BatmanAoD: How would `maturin` know that it's running inside of a `uv run` without this flag?

crates/lib/Makefile.toml
BatmanAoD: yes, claude is way too conservative about "breaking changes" within parts of a project that can't be used by external projects

crates/lib/Makefile.toml
BatmanAoD: Discussed the rationale for `--active` in other projects offline; but in this project, I've removed it.

crates/lib/src/python/mod.rs
asaites: I think there are a few points in favor of this approach (generating `__init__.py` rather than writing it by hand):

- It keeps the module content co-located.
- It reduces the the burden of keeping `__all__` up-to-date.
- If desired, we can easily force the Rust and Python modules to have the same docs.
- Developers can assume/treat nearly all Python files as generated.

crates/lib/Makefile.toml
asaites: 👍 

For what it's worth, I do develop and test using multiple Python versions.

crates/lib/Makefile.toml
BatmanAoD: went ahead and made that change

crates/lib/Makefile.toml
asaites: This flag affects what how it installs the package: whether it runs `pip install -e .` or `uv pip install -e .` [There are differences](https://docs.astral.sh/uv/pip/compatibility/#compatibility-with-pip-and-pip-tools) in how `uv` handles it, but those are likely immaterial if we're all in on ~`pip`~`uv`.

The actual potential issue is that, somewhat ironically, [it will use the active `VIRTUAL_ENV` rather than the `uv`-managed one](https://docs.astral.sh/uv/pip/environments/#discovery-of-python-environments)! In fact, if the `.venv` doesn't exist and `VIRTUAL_ENV` isn't set (and it's not in a Conda environment), the command prints an error telling you to use a virtual environment.

I'm now looking more closely at some of our other projects' `Makefile.toml`s, and realize in some cases this turns out not to matter, because we have an `install-deps` task that uses `uv pip install --group dev .`. The consequence is that it expects an active virtualenv, and it'll install it there. That's _also_ what is breaking some of the other tasks that intend to use the `.venv` environment but are using the `uv pip` interface instead of `uv run`.

With all that in mind, I think we should avoid `uv pip` entirely, and instead just use `uv run`, including to run `maturin develop`, and _specifically without_ the `--uv` flag (because `maturin` _does_ have logic to recognize its running within a `uv`-managed environment, and we _don't_ want it to sidestep it via `uv pip install`).

crates/lib/Makefile.toml
BatmanAoD: I don't really know what the context was for this, and I haven't actually run into it being a problem. From googling, I suspect it was a workaround for this issue, which appears to have been fixed: https://github.com/astral-sh/uv/issues/8821

crates/lib/Makefile.toml
BatmanAoD: >those are likely immaterial if we're all in on pip.

Do you mean all in on `uv`? I think that internally we're all-in on `uv`, so for `maturin develop` it does make sense not to worry about `pip`. We can't assume that downstream users are using `uv`, but I'm guessing `maturin` doesn't ever produce "uv-only" wheels anyway, so maybe that doesn't matter.

BatmanAoD: If I understand your second link, it's basically saying that `uv pip install/sync` is like `uv install/sync --active`? That makes sense, since `pip` itself *always* respects the active virtual environment, but I don't understand your conclusion that it's therefore better *not* to use the `--uv` flag on `maturin develop`. If it's using `uv pip install` instead of `uv install` anyway, then there's no way to opt out of the need for `VIRTUAL_ENV` to be set, unless I'm misunderstanding.

asaites: > Do you mean all in on `uv`?

Yes, exactly.

> I don't understand your conclusion that it's therefore better not to use the `--uv` flag on `maturin develop`

To get right to the point: if we use `uv run --group dev maturin develop`, the `--uv` flag won't change the behavior, independent of whether or not there's an active virtual environment and/or if we pass `--active` to `uv run`.

> If I understand your second link, it's basically saying that `uv pip install`/`sync` is like `uv install/sync --active`?

The broader point I was trying to make is that this isn't exactly true, and the ways in which it isn't are probably worth preferring to avoid `uv pip`.

Specifically, note that `uv pip` changes the how `uv` finds and manages environments. It _prefers_ the current `VIRTUAL_ENV` over `.venv`, and it _fails with an error_ if it can't find one. In contrast, `uv run` ordinarily uses its managed `.venv`, _and_ it'll create it when necessary. In fact, using `uv run --active` will _also_ create/manage `.venv` if there's not an active virtual environment.

In at least that way, `uv install --active` is not equivalent to `uv pip install`; certainly it's not why the `uv pip` interface exists. As [they put it](https://docs.astral.sh/uv/pip/compatibility/):

> Informally, the intent is such that existing pip and pip-tools users can switch to uv without making meaningful changes to their packaging workflows; and, in most cases, swapping out pip install for uv pip install should "just work".
>
> However, uv is not intended to be an exact clone of pip, and the further you stray from common pip workflows, the more likely you are to encounter differences in behavior. In some cases, those differences may be known and intentional; in others, they may be the result of implementation details; and in others, they may be bugs.

That's why I said I think we should avoid the interface, which would include avoiding the `--uv` flag for `maturin` (even if it doesn't make an observable difference if we run it through `uv run`).

BatmanAoD: Okay, I've adopted `uv run --group dev` everywhere, and dropped the `--uv` from `maturin develop`.

.github/workflows/checks.yml
```diff
@@ -80,11 +80,13 @@ jobs:
...
         with:
-          python-version: '3.13'
-
-      - uses: abatilo/actions-poetry@v4
+          enable-cache: true
+          cache-dependency-glob: crates/lib/uv.lock
 
       - name: Test Python Package
-        run: cargo make --cwd crates/lib --makefile Makefile.toml test-python-package
+        shell: bash
```
asaites: Do you specifically need `shell: bash` here? I noticed in `publish-docs.yml` there's a similar `run:` invocation, but no `shell` specifier. Given all the stuff `cargo-make` handles internally, I'd be surprised if it is needed.

crates/lib/src/python/mod.rs
```diff
@@ -78,17 +93,25 @@ fn init_module(m: &Bound<'_, PyModule>) -> PyResult<()> {
...
+    // `pyo3_tracing_subscriber_build`, so `qcs_sdk._tracing_subscriber` is its import path.
     pyo3_tracing_subscriber::add_submodule("qcs_sdk", "_tracing_subscriber", py, m)?;
 
     Ok(())
 }
 
 #[cfg(feature = "stubs")]
 mod stubs {
-    use pyo3_stub_gen::{define_stub_info_gatherer, generate::Module, Result, StubInfo};
+    use pyo3_stub_gen::{
```
asaites, line 111: I apologize in advance that I'm doing this as a drive-by, but I wanted to at least share the gist with you since I won't be able to finish this review until this evening/tomorrow. I bet this can be cleaned up more, but here's a version of this module that generates valid stubs (as far as `stubtest` is concerned, with the caveat mentioned in my earlier comment):

```rust
    use pyo3_stub_gen::{
        define_stub_info_gatherer, export_verbatim, exclude_from_all,
        derive::gen_type_alias_from_python,
        generate::Module, module_doc, reexport_module_members, Result,
        StubInfo,
    };
    use std::path::Path;

    #[derive(pyo3::IntoPyObject)]
    struct Final<T>(T);

    impl<T> pyo3_stub_gen::PyStubType for Final<T>
    where
        T: pyo3_stub_gen::PyStubType,
    {
        fn type_output() -> pyo3_stub_gen::TypeInfo {
            let mut inner = T::type_output();
            inner.import.insert("typing".into());
            inner.name = format!("typing.Final[{}]", inner.name);
            inner
        }
    }

    pyo3_stub_gen::module_variable!("qcs_sdk._qcs_sdk", "__version__", Final<&str>);

    pyo3_stub_gen::module_variable!(
        "qcs_sdk._qcs_sdk.compiler.quilc",
        "DEFAULT_COMPILER_TIMEOUT",
        Final<f64>
    );

    pyo3_stub_gen::module_doc!("qcs_sdk._qcs_sdk", "{}", super::PACKAGE_DOC);

    gen_type_alias_from_python!(
        "qcs_sdk._qcs_sdk.client",
        r#"
        from qcs_api_client_common._qcs_api_client_common import configuration

        AuthServer: TypeAlias = configuration.AuthServer
        ClientCredentials: TypeAlias = configuration.ClientCredentials
        ExternallyManaged: TypeAlias = configuration.ExternallyManaged
        OAuthSession: TypeAlias = configuration.OAuthSession
        RefreshToken: TypeAlias = configuration.RefreshToken
        SecretRefreshToken: TypeAlias = configuration.SecretRefreshToken
        "#
    );

    reexport_module_members!("qcs_sdk" from "qcs_sdk._qcs_sdk");
    reexport_module_members!("qcs_sdk.client" from "qcs_sdk._qcs_sdk.client");
    reexport_module_members!("qcs_sdk.compiler" from "qcs_sdk._qcs_sdk.compiler");
    reexport_module_members!("qcs_sdk.compiler.quilc" from "qcs_sdk._qcs_sdk.compiler.quilc");
    reexport_module_members!("qcs_sdk.diagnostics" from "qcs_sdk._qcs_sdk.diagnostics");
    reexport_module_members!("qcs_sdk.qpu" from "qcs_sdk._qcs_sdk.qpu");
    reexport_module_members!("qcs_sdk.qpu.api" from "qcs_sdk._qcs_sdk.qpu.api");
    reexport_module_members!("qcs_sdk.qpu.experimental" from "qcs_sdk._qcs_sdk.qpu.experimental");
    reexport_module_members!("qcs_sdk.qpu.experimental.random" from "qcs_sdk._qcs_sdk.qpu.experimental.random");
    reexport_module_members!("qcs_sdk.qpu.isa" from "qcs_sdk._qcs_sdk.qpu.isa");
    reexport_module_members!("qcs_sdk.qpu.translation" from "qcs_sdk._qcs_sdk.qpu.translation");
    reexport_module_members!("qcs_sdk.qvm" from "qcs_sdk._qcs_sdk.qvm");
    reexport_module_members!("qcs_sdk.qvm.api" from "qcs_sdk._qcs_sdk.qvm.api");

    reexport_module_members!("qcs_sdk._qcs_sdk.client" from "qcs_api_client_common._qcs_api_client_common"; "configuration");
    reexport_module_members!("qcs_sdk._qcs_sdk.compiler.quilc" from "quil._quil"; "program");
    exclude_from_all!("qcs_sdk._qcs_sdk.client", "configuration");
    exclude_from_all!("qcs_sdk._qcs_sdk.compiler.quilc", "program");

    reexport_module_members!("qcs_sdk" from "qcs_sdk._qcs_sdk.client"; "QCSClient");

    reexport_module_members!("qcs_sdk" from "qcs_sdk"; "_tracing_subscriber");
    reexport_module_members!("qcs_sdk._qcs_sdk" from "qcs_sdk"; "_tracing_subscriber");
    export_verbatim!("qcs_sdk._qcs_sdk", "_tracing_subscriber");
    export_verbatim!("qcs_sdk._qcs_sdk", "_gather_diagnostics");
    export_verbatim!("qcs_sdk._qcs_sdk", "__version__");

    module_doc!("quil._quil.validation", "");

    define_stub_info_gatherer!(internal_stub_info);

    /// Gather stub information to generate stub files.
    pub fn stub_info() -> Result<StubInfo> {
        let manifest_dir: &Path = env!("CARGO_MANIFEST_DIR").as_ref();
        let mut stubs = StubInfo::from_pyproject_toml(manifest_dir.join("pyproject.toml"))?;

        // `pyo3_stub_gen` gathers declarations from every linked crate, so dependencies that
        // define their own Python packages — `quil-rs`, `qcs-api-client-common` — contribute
        // their modules here too. Generation rejects any non-empty module outside this
        // package's module path, and writing files for them would shadow the real, installed
        // packages, so drop them. The cost is that our stubs refer to foreign types by their
        // private paths (`quil._quil.program.Program`), since resolving the public alias needs
        // the foreign module to still be present at generation time.
        stubs
            .modules
            .retain(|name, _| name == "qcs_sdk" || name.starts_with("qcs_sdk."));

        Ok(stubs)
    }
```

crates/lib/python/stubtest-allowlist
```diff
@@ -1,46 +1,19 @@
...
-qcs_sdk.qpu.__all__
-qcs_sdk.qpu.api.__all__
-qcs_sdk.qpu.experimental.__all__
-qcs_sdk.qpu.experimental.random.__all__
-qcs_sdk.qpu.isa.__all__
-qcs_sdk.qpu.translation.__all__
-qcs_sdk.qvm.__all__
-qcs_sdk.qvm.api.__all__
+# The generated stubs don't (yet) carry an `__all__` for the private extension module.
+qcs_sdk._qcs_sdk.__all__
```
asaites: This comment isn't true, and the `stubtest` error reported when this line is removed is correct: there are things exported at runtime that aren't in the stub, namely `__version__`, `_gather_diagnostics`, and `_tracing_subscriber`.

I've done some experimentation (posted in another comment on the `src/python/mod.rs`), and found that we can coerce the generator to output stubs that `stubtest` is (mostly) happy with and that will import correctly. With those changes, we can drop this line along with line 5, _and_ we can run the `stubtest` task **without** the special `stubtest-mypy.ini` config or `--ignore-unused-allowlist` options.

asaites (commented):
I haven't quite finished my review yet, but will tonight or tomorrow, but in the meantime, I've left comments on my experimentation with `pyo3_stub_gen` to get valid output as far as `stubtest` is concerned.

crates/lib/src/python/mod.rs
BatmanAoD: Huh, I did not realize that `reexport_module_members` would work for this. It seems very reasonable.

BatmanAoD: Adopted this with minimal changes.

.github/workflows/checks.yml
BatmanAoD: looks totally unnecessary to me, removed

crates/lib/src/python/mod.rs
asaites: Thank you! I'm taking another look now.

crates/lib/Makefile.toml
```diff
@@ -46,41 +46,57 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
-# NOTE: if using `uv`, you may need to set `PYTHONHOME` to the root of the virtual environment:
-#
-#   export PYTHONHOME=$(python3 -c "import sys; print(sys.base_prefix)")
 [tasks.generate-stubs]
 description = "Generate Python stub files."
-command = "cargo"
-args = ["run", "-p", "qcs", "--features", "stubs", "--bin", "stub_gen"]
-install_crate = false
+script_runner = "bash"
+script = '''
```
asaites: Instead of invoking a shell, it might make more sense to have a private task for the `cargo run`, a separate task for formatting, and then the `generate-stubs` just ties them together:

```toml
[tasks.generate-stubs]
description = "Generate Python stub files."
dependencies = ["generate-stubs-private", "format-stubs"]

[tasks.generate-stubs-private]
private = true
command = "uv"
args = ["run", "cargo", "run", "-p", "qcs", "--features", "stubs", "--bin", "stub_gen"]

[tasks.format-stubs]
description = "Format Python stubs."
command = "uv"
args = ["run", "--group", "dev", "ruff", "format", "python/qcs_sdk"]
```

Either way, I _think_ you want the `cargo run` to happen within `uv run`'s context to ensure `pyo3_stub_gen` is using the expected interpreter. It'll _probably_ work fine without it so long as there is _some_ reasonably up-to-date Python interpreter on the path (it's used to gather some runtime representations of constants inserted into the generated stubs, so it's not likely to create a linking issue), but if we do use `uv run`, it should make issues less likely.

crates/lib/Makefile.toml
```diff
@@ -46,41 +46,57 @@ dependencies = ["deny", "lint", "check", "test-flow"]
...
+fi
+echo "Generated Python files are up to date."
+'''
 
 [tasks.package-python]
-description = "Build a release version of the Python package to the active virtual environment."
-dependencies = ["install-python-deps", "generate-stubs"]
-command = "poetry"
-args = ["run", "maturin", "build", "--release"]
+description = "Build a release version of the Python package and install in the virtual environment."
```
asaites: Both the original comment and this one are wrong, in the sense that `maturin build` just produces a wheel, but doesn't install it.


```suggestion
description = "Build a release version of the Python package."
```

asaites (approved):
Approved, with a couple non-blocking suggestions.

crates/lib/Makefile.toml
BatmanAoD: Applied this change

PR #596, opened 2026-08-31
fix: no more 'grpc-web' wheels

BatmanAoD:
Python releases cannot use the 'grpc-web' feature any more, since it no longer exists


BatmanAoD:
Successful prerelease: https://github.com/rigetti/qcs-sdk-rust/actions/runs/33430875479

Shadow53 (approved):

PR #597, opened 2026-09-09
feat!: use endpoint-ID for Gateway-bound jobs, when possible

### LINKED ISSUE #598: Gateway connection using Endpoint ID rather than QPU ID
When the endpoint ID for a job is known, that should take precedence over the QPU ID.

BatmanAoD:
Resolves #598


crates/lib/src/python/mod.rs
```diff
@@ -109,10 +109,8 @@ fn init_module(m: &Bound<'_, PyModule>) -> PyResult<()> {
 #[cfg(feature = "stubs")]
 mod stubs {
     use pyo3_stub_gen::{
-        define_stub_info_gatherer, export_verbatim, exclude_from_all,
-        derive::gen_type_alias_from_python,
-        generate::Module, module_doc, reexport_module_members, Result,
-        StubInfo,
+        define_stub_info_gatherer, derive::gen_type_alias_from_python, exclude_from_all,
+        export_verbatim, generate::Module, module_doc, reexport_module_members, Result, StubInfo,
```
BatmanAoD: I think this must be a `cargo clippy --fix` change, but when I re-run that, it deletes `generate::Module` in addition to the reordering, so I'm not sure what's going on.

crates/lib/src/qpu/execution.rs
```diff
@@ -199,14 +200,15 @@ impl<'a> Execution<'a> {
     where
         S: Into<Cow<'a, str>>,
     {
+        let endpoint_id: Cow<'a, str> = endpoint_id.into();
+        let connection_strategy = ConnectionStrategy::EndpointId(endpoint_id.clone().into_owned());
```
jselig-rigetti: This feels weird, if we just immediately convert using `into` and then `clone`, it feels like the type constraint should just be `ToString` or something, or just have the argument itself be a `String`. Just a nit.

crates/lib/python/qcs_sdk/_qcs_sdk/qpu/api/__init__.pyi
```diff
@@ -174,10 +176,25 @@ class ConnectionStrategy:
...
+            If given, used as the request's target instead of the quantum processor ID.
+            """
+        @property
+        def liveness(self) -> EndpointLiveness:
+            r"""
+            Whether the gateway should route to live hardware, simulated hardware, or either.
+            """
+        def __new__(
+            cls,
+            liveness: EndpointLiveness = EndpointLiveness.LIVE_ONLY,
```
jselig-rigetti: hmm, what do we do if someone wants a simulated endpoint but doesn't pass an endpoint id?

jselig-rigetti (approved):
In general approved, but some questions.
Curious what @asaites thinks RE python changes.

crates/lib/src/qpu/execution.rs
BatmanAoD: Looks like `to_string` is what the original uses, so I'll restore that.

crates/lib/python/qcs_sdk/_qcs_sdk/qpu/api/__init__.pyi
BatmanAoD: That's fine; it's how I intend to use this in Gateway load-test, with `ConServ-integration-QPU-9Q`.

crates/lib/src/qpu/execution.rs
```diff
@@ -180,67 +178,21 @@ impl<'a> Execution<'a> {
...
-            params,
-            Some(&self.quantum_processor_id.clone()),
-            translation_options,
-            execution_options,
-        )
-        .await
-    }
-
-    /// Run on specific QCS endpoint and wait for the results.
-    pub(crate) async fn submit_to_endpoint_id<S>(
```
BatmanAoD: This use-case is covered by `submit_to_target` with a connection-strategy using the endpoint ID.

